PR #4, opened 2017-12-07
PR for `pspec_graduation`

nkern:
this is a PR to graduate some of the basic code for `hera_pspec`, including `dataset.py`, `version.py` and `conversions.py` with tests. The `setup.py` installation and travis testing environment are also graduated, and the old `pspec.py` code is also graduated for the purpose of backwards compatibility, although future PRs may eliminate this from `master` and move it to a branch.


nkern:
I'll review the code that @philbull committed, which was the `dataset.py` stuff, and if Phil you can review my commits (`version.py`, `conversions.py` and the setup and testing environment) then we should be good to merge--if you've already gone through it then feel free to just sign off on it. 

@acliu if you also want to sign off on this that might be good too (I know you are busy so no pressure to actually review it).

PR #27, opened 2018-01-30
Refactor bootstrapping code

philbull:
Refactor the support functions for bootstrapping, and the way `PSpecData` handles groups of baselines.


philbull:
Sure, I can make it do that.

philbull:
OK, I've addressed @carinacheng's comment, fixed some bugs, and added unit tests.

philbull:
I've now (hopefully) fixed the units of `pspec()`, added tests on the normalisation based on Parseval's theorem, and added a couple of extra methods to output units and an array of delays.

hera_pspec/tests/test_bootstrap.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_collapse_over_lsts(self):
+        #collapse_over_lsts(pspec_list)
+        pass
+    
+    def test_group_baselines(self):
+        """
+        Test baseline grouping behavior.
+        """
+        # Generate example lists of baselines
+        bls1 = [(0,i) for i in range(1)]
```
acliu: The original pspecdata tutorial notebook has tuples with 3 elements. I think the code works fine with 3-element tuples (right?), but all of this should be clarified.

hera_pspec/tests/test_pspecdata.py
```diff
@@ -219,17 +273,93 @@ def test_get_G(self):
...
+                
                 # Test that all elements of G are positive up to numerical noise
                 # with the threshold set to 10 orders of magnitude down from
                 # the smallest value on the diagonal
                 for i in range(Nfreq):
                     for j in range(Nfreq):
-                        self.assertGreaterEqual(G[i,j], -min_diagonal*10**-10)
+                        self.assertGreaterEqual(G[i,j], -min_diagonal*1e-10)
             
+    def test_parseval(self):
```
acliu: I'm actually no longer sure that this is the right test. Let's talk about it in person.

hera_pspec/tests/test_bootstrap.py
```diff
@@ -0,0 +1,102 @@
...
+                                                 keep_remainder=True, 
+                                                 randomize=rand)
+                    except:
+                        continue
+                    count = np.sum([len(_g) for _g in g])
+                    self.assertEqual(count, len(bls))
+        
+        # Check that random seed works
+        g1 = boot.group_baselines(bls5, 3, randomize=True, seed=10)
+        g2 = boot.group_baselines(bls5, 3, randomize=True, seed=11)
```
acliu: `g2` doesn't seem to be used?

hera_pspec/bootstrap.py
```diff
@@ -0,0 +1,134 @@
...
+    # 2. Average over time for each bootstrap sample
+    NotImplementedError()
+
+
+def group_baselines(bls, Ngroups, keep_remainder=False, randomize=False, 
+                    seed=None):
+    """
+    Group baselines together into equal-sized sets.
+    
+    These groups can be passed into PSpecData.pspec(), where the corresponding
```
acliu: Can we have an explicit demonstration of this?

hera_pspec/pspecdata.py
```diff
@@ -608,45 +607,98 @@ def p_hat(self, M, q):
...
+        """
+        # Frequency units of UVData are always Hz => always convert to ns
+        delay_units = 'ns'
+        
+        # Work out the power spectrum units
+        if len(self.dsets) == 0:
+            raise IndexError("No datasets have been added yet; cannot "
+                             "calculate power spectrum units.")
+        else:
+            pspec_units = "(%s)^2 (ns)^-1" % self.dsets[0].vis_units
```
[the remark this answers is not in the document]
philbull: `self.dsets[0]` is a `UVData` object. The `UVData` class has a standard property called `vis_units`, so I think this should be fine.

hera_pspec/tests/test_bootstrap.py
philbull: Yes, I'll add a comment in the tests. This code just shuffles tuples or lists of tuples -- it doesn't care if they are valid 3-element baseline tuples or not. I think we don't need a validity check at this point either, as all of that will be done on passing the baselines into `pspec()`

philbull:
By agreement with Adrian, some of the requested changes (e.g. to documentation) can be made at a later date. We will merge this now to push forward the graduation of this code into `master`...

Issue #46, opened 2018-03-10
PSpecData.validate_datasets across time axis


nkern:
I was accidentally using two UVData objects with different Ntimes, which caused a broadcast_error in the pspec calculation. There should be a similar check across the time axes in `PSpecData.validate_datasets`. I have a branch off of `pspec_graduation` to add this in, and added unit-tests for the function as well.

@philbull, was there a reason why a check along time wasn't included originally?

nkern:
there are a lot of branches floating around right now. I think we should merge one into master as soon as possible so we can all start making PRs without having to merge branches w/ other branches. it seems that probably I should have branched off of `boot_refactor` for this fix, as that seems to be the most up-to-date: is that right @philbull?

nkern:
addressed by #53

PR #47, opened 2018-03-10
Time validation on datasets

nkern:
Validation of time axes of input dsets should be checked in `PSpecData.validate_datasets`. otherwise, this leads to a less-informative numpy broadcasting error further downstream..


nkern:
addresses #46

Issue #41, opened 2018-03-08
pspecdata suggestions to reduce pre-processing


jkerrigan:
Specify which axes to cross-multiply (e.g. LST and/or redundant baselines). Current stage of the hera_pspec module only cross-multiplies along LSTs.

carinacheng:
Also have as an option to cross-multiply autos too, if wanted.

nkern:
addressed by #60

PR #48, opened 2018-03-12
adds the ability to form cross-spectra between redundant baseline groups

nkern:
This PR slightly reformats `PSpecData.pspec` to allow for forming cross-spectra between redundant baseline groups. However, default behavior isn't changed. Also adds tests for `PSpecData.pspec`.

Addresses part of issue #41.


nkern:
note that this has the `time_validation` PR branch rebased in, so one should ignore changes to `pspecdata.validate_datasets`, which will automatically be removed from this PR once that branch is merged in w/ `pspec_graduation`.

PR #51, opened 2018-03-12
Graduate new PSpec code into the `master` branch

philbull:
The new `PSpecData` object and supporting infrastructure are now mature enough to be merged into `master`, so users can start trying them out. The API isn't stable yet, but the core methods are reasonably well tested and documented, so should be usable.


hera_pspec/conversions.py
```diff
@@ -60,7 +60,7 @@ class Cosmo_Conversions(object):
     # astropy load attribute
     _astropy = _astropy
 
-    def __init__(self, Om_L=0.68440, Om_b=0.04911, Om_c=0.26442, H0=100.0,
+    def __init__(self, Om_L=0.68440, Om_b=0.04911, Om_c=0.26442, H0=67.27,
```
nkern: Why was this changed to 67? I think we said we wanted to keep this default at 100.0 such that the output distance units are in h-1 Mpc by default.

acliu: In isolation, we thought this was ok. But when combined with the way we were calling the conversions module downstream, it meant that the code got very confusing because the default behaviour became one where the results didn't change whether or not you used little_h.

acliu:
I'm going to trust that Phil didn't sneakily change something after we approved his other PRs ;) In other words, I'm going to approve on the assumption that nothing changed (aside from tests etc.) after my last review of the bootstrap work. Can Phil confirm that this is the case?

hera_pspec/conversions.py
nkern: okay, so it seems like then we should update the documentation in `conversions.py` to specify that the units don't have little_h in them.

hera_pspec/pspecbeam.py
```diff
@@ -0,0 +1,331 @@
...
+                Default: 10000
+
+        Returns
+        -------
+        scalar: float
+                [\int dnu (\Omega_PP / \Omega_P^2) ( B_PP / B_P^2 ) / (X^2 Y)]^-1
+                in h^-3 Mpc^3 or Mpc^3.
+        """
+
+        integration_freqs = np.linspace(lower_freq,upper_freq,num_steps)
```
nkern: also should have `endpoint=True`

hera_pspec/conversions.py
philbull: The `scalar()` method now has a kwarg called `little_h` that lets the user choose whether to output in h^-1 units. If we don't set H_0 to something other than 100, this parameter has no effect, so I think it makes sense to change the default here.

philbull:
I did make some further changes (as discussed) after merging everything into `pspec_graduation`. They were mostly changes to tests, but I also removed and/or renamed some obsolete files.

hera_pspec/pspecbeam.py
```diff
@@ -0,0 +1,331 @@
...
```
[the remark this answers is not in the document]
nkern: scratch that, it should be `endpoint=False`

hera_pspec/pspecbeam.py
nkern: should be `endpoint=False`

hera_pspec/conversions.py
acliu: Alright, I just fixed this along the lines that we discussed, where the user can now explicitly ask for little_h to be used or not in every distance function.

hera_pspec/pspecbeam.py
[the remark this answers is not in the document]
acliu: Ok, I changed that. (It makes basically no difference, actually, because this does not correspond to the frequency channels of the measurement, but is just a grid in frequency to mimic a continuous integral).

hera_pspec/pspecbeam.py
acliu: Ok, I changed that. (It makes basically no difference, actually, because this does not correspond to the frequency channels of the measurement, but is just a grid in frequency to mimic a continuous integral).

PR #52, opened 2018-03-15
Fixed bugs in pspecbeam

nkern:
There are a few bugs in `pspecbeam` I found while fixing one particular bug. These include

1. All `np.linspace()` calls should have `endpoint=False`, which is actually my fault for the confusion from the previous PR where I said the opposite.

2. `aipy` is not loaded in the module even though there is an `aipy` dependence when using a `taper` kwarg that is non-default.

3. `num_freqs` directly clashes with `num_steps`, which causes a broadcasting error when specifying a non-default `taper`. `num_freqs` is otherwise not used, so it is eliminated from the module.

4. With the above updates, the pspec scalar convergence test does not actually converge at only 20000 num_steps for some reason, not even close, even though this test was passing before. Not sure why this is the case, but I did a benchmark and the calculation accuracy is hindered by the `interp1d` call, which defaults to linear interpolation. Simply changing this to quadratic interpolation improved the accuracy of the scalar calculation and sped it up slightly, such that it now converges to an error of 1e-5 at 50000 steps, which is the new default.

5. `compute_pspec_scalar()` is duplicated in both `PSpecBeamGauss` and `PSpecBeamUV`. I understand why this was probably done, because the beam integrals are different for the two objects, but code duplication should be avoided at all costs. A nice way we get around this in `hera_cal` is to outsource the core functionality to a more agnostic, module-level function, which is then called by each class in a way that suits them. I did this here for `compute_pspec_scalar`. This is also nice b/c you may want to use this core functionality on your own without having to instantiate one of these objects.. say you have a beam that isn't gaussian or a pyuvdata object.

6. in general there were some pep8 issues, although not technically bugs. also I added some more comments to the code where I thought necessary.

7. lastly, I updated some of the testing for the module

### DIFF
--- a/hera_pspec/pspecbeam.py
+++ b/hera_pspec/pspecbeam.py
@@ -7,9 +7,12 @@ from scipy.interpolate import interp1d
 import aipy
 
 
-def compute_pspec_scalar(cosmo, beam_freqs, omega_ratio, lower_freq, upper_freq, 
+def _compute_pspec_scalar(cosmo, beam_freqs, omega_ratio, lower_freq, upper_freq, 
                          num_steps=50000, stokes='pseudo_I', taper='none', little_h=True):
     """
+    This is not to be used by the novice user to calculate a pspec scalar. 
+    Instead, look at the PSpecBeamUV and PSpecBeamGauss classes.
+
     Computes the scalar function to convert a power spectrum estimate
     in "telescope units" to cosmological units
 
@@ -75,14 +78,16 @@ def compute_pspec_scalar(cosmo, beam_freqs, omega_ratio, lower_freq, upper_freq,
     # derived from the beam model to the same frequency grid as the power spectrum
     # estimation
     beam_model_freqs_MHz = beam_freqs / 1e6
+    num_freqs = len(beam_freqs)
     dOpp_over_Op2_fit = interp1d(beam_model_freqs_MHz, omega_ratio, kind='quadratic')
     dOpp_over_Op2 = dOpp_over_Op2_fit(integration_freqs_MHz)
 
     # Get B_pp = \int dnu taper^2 and Bp = \int dnu
     if taper == 'none':
-        dBpp_over_BpSq = np.ones_like(integration_freqs)
+        dBpp_over_BpSq = np.ones_like(num_steps)
     else:
-        dBpp_over_BpSq = aipy.dsp.gen_window(num_steps, taper)**2
+        dBpp_over_BpSq = aipy.dsp.gen_window(num_freqs, taper)**2
+        dBpp_over_BpSq = intper1d(beam_freqs, dBpp_over_BpSq)(integration_freqs, kind='nearest')
     dBpp_over_BpSq /= (integration_freqs[-1] - integration_freqs[0])**2
 
     # integrate to get scalar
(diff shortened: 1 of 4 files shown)


nkern:
figure_1
this is the scalar using `kind='linear'`, `kind='quadratic'` or `kind='cubic'` in the `interp1d` call. you can see that `quadratic` gets us to convergence faster than `linear` but at the same rate as `cubic`

nkern:
ok @acliu , I think this is ready for a review from you now.

hera_pspec/pspecbeam.py
```diff
@@ -1,24 +1,167 @@
...
+import numpy as np
+import pyuvdata
+import os
+from hera_pspec import conversions
 from scipy import integrate
 from scipy.interpolate import interp1d
+import aipy
+
+
+def _compute_pspec_scalar(cosmo, beam_freqs, omega_ratio, pspec_freqs, num_steps=20000,
```
acliu: Is there a reason the default `numsteps` parameter here does not match the defaults elsewhere in the file?

hera_pspec/pspecbeam.py
```diff
@@ -1,24 +1,167 @@
...
+    scalar = 1.0 / integrate.trapz(d_inv_scalar, x=integration_freqs)
+
+    return scalar
 
 
 class PSpecBeamBase(object):
 
     def __init__(self, cosmo=None):
-        if cosmo != None:
+        if cosmo is not None:
```
acliu: Just curious, what's the rationale here? I would've thought that testing for equality would be more robust than testing for identity.

hera_pspec/tests/test_pspecbeam.py
```diff
@@ -99,14 +102,12 @@ def test_Gaussbeam(self):
...
-        # assertAlmostEqual does absolute comparisons, so we want to put things 
-        # on a sensible scale
-        scalar /= 10**9
-        scalar_double_steps /= 10**9
-        self.assertAlmostEqual(scalar, scalar_double_steps)
+        scalar_large_Nsteps = self.gauss.compute_pspec_scalar(lower_freq, upper_freq, num_freqs, num_steps=5000)
+        self.assertAlmostEqual(scalar / scalar_large_Nsteps, 1.0, delta=1e-5)
+
+        # test taper execution
+        scalar = self.bm.compute_pspec_scalar(lower_freq, upper_freq, num_freqs, num_steps=2000, taper='blackman')
```
acliu: Can we add a numerical test for this as well? Just to make sure that when the taper is applied the number is correct. I'm happy to be the one to add this if it's easier for you; just let me know!

hera_pspec/pspecbeam.py
nkern: nope, i guess just a typo..

hera_pspec/pspecbeam.py
nkern: yeah I agree, however, this is just the PEP8 standard

hera_pspec/tests/test_pspecbeam.py
nkern: do we have an analytic number to compare to? if you do feel free to add it. otherwise I'm just going to see what that number is (given the current format of `_compute_pspec_scalar` and the input args) and put in an `almost_equal` float comparison, which I think is how we're currently doing it...

nkern: either way feel free to do it yourself if you'd like!

nkern: did it

hera_pspec/pspecbeam.py
nkern: fixed

nkern:
@acliu once this passes and if you're ok w/ everything feel free to merge it

PR #56, opened 2018-04-01
enables forming cross-redundant baseline spectra

nkern:
updates `PSpecData.pspec` to take in a list of redundant baseline group to form all combinations of cross-baseline spectra.


nkern:
@philbull the parseval test is failing and I'm not sure why. if you have any insight it would be appreciated!

PR #57, opened 2018-04-03
Container class for UVPSpec objects

philbull:
This PR implements a container class for collecting together UVPSpec power spectrum objects. It's based on HDF5, and allows power spectra to be grouped together into a simple directory-like structure. Inside the file, the power spectra are structured as `group_name/pspec_name/<data>`.


nkern:
add a Jy2mK conversion step in `pspec_red` and add relevant flag in config file. looks like this:
```
data.data_array *= beam.Jy_to_mK(uvd.freq_array.flatten())[None, None, :, None]
```

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+    def _store_pspec(self, pspec_group, ps):
+        """
+        Store a UVPSpec object as group of datasets within the HDF5 file.
+        
+        Parameters
+        ----------
+        pspec : HDF5 group
+            HDF5 group to store power spectrum data in.
+        
+        ps : UVPSpec
```
nkern: a little pedantic, but to be consistent with other parts of the repo, can we use the variable name `uvp` (instead of `ps`) to refer to an instance of a UVPSpec object?

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+            raise IOError("HDF5 file was opened read-only; cannot write to file.")
+        
+        # Get data and attributes from UVPSpec object (stored in dicts)
+        assert isinstance(ps, UVPSpec)
+        
+        # Write UVPSpec to group
+        ps.write_to_group(pspec_group, run_check=True)
+        
+    
+    def _load_pspec(self, grp):
```
nkern: change this variable to `pspec_group` to match convention in `_store_pspec`?

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Load a new UVPSpec object from a HDF5 group.
+        
+        Parameters
+        ----------
+        grp : HDF5 group
+            Group containing datasets that contain power spectrum and 
+            supporting information, in a standard format expected by UVPSpec.
```
nkern: Needs a `Returns` section

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        # Check that group is tagged as containing UVPSpec (pspec_type attribute)
+        if 'pspec_type' in grp.attrs.keys():
+            if grp.attrs['pspec_type'] != UVPSpec.__name__:
+                raise TypeError("HDF5 group is not tagged as a UVPSpec object.")
+        else:
+            raise TypeError("HDF5 group is not tagged as a UVPSpec object.")
+        
+        # Create new UVPSpec object and fill with data from this group
+        pspec = UVPSpec()
```
nkern: see previous comment on `uvp` nomenclature for instances of `UVPSpec`

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+        ----------
+        group : str, optional
+            Which group the power spectrum belongs to.
+        
+        psname : str, optional
+            The name of the power spectrum to return.
+        
+        Returns
+        -------
+        pspec : UVPSpec or list of UVPSpec
```
nkern: what is actually returned is not called `pspec`

hera_pspec/container.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Force HDF5 file to flush to disk.
+        """
+        self.data.flush()
+    
+    def __del__(self):
+        """
+        Make sure that HDF5 file is closed on destruct.
+        """
+        try:
```
nkern: why is there a `try: except:` here? why would this every fail? If this is needed, can we be specific as to what Error is will except?

hera_pspec/uvpspec.py
```diff
@@ -554,91 +555,205 @@ def key_to_indices(self, key, *args):
...
+        
+        pols : list of str or int
+            List of polarization strings or integers to keep. See 
+            pyuvdata.utils.polstr2num for acceptable options.
+        
+        only_pairs_in_bls : bool, optional
+            If True, keep only baseline-pairs whose first _and_ second baseline
+            are found in the 'bls' list. Default: False.
         """
         # clear object
```
nkern: you can erase this line b/c the `_clear()` method is called in `read_from_group`

hera_pspec/uvpspec.py
```diff
@@ -729,28 +830,53 @@ def _clear(self):
         for p in self._all_params:
             if hasattr(self, p):
                 delattr(self, p)
-
+    
+    def __str__(self):
```
nkern: this is a function that is currently uncovered by travis. we just need a very simple test for this, like not even anything with asserts, just something that runs it and makes sure it finishes with no errors..

scripts/pspec_red.py
```diff
@@ -0,0 +1,137 @@
...
+import pyuvdata as uv
+import os, sys, glob, time
+
+# Default settings for pspec calculation
+pspec_defaults = {
+    'overwrite':                    False,
+    'little_h':                     True,
+    'avg_group':                    False,
+    'exclude_auto_bls':             False,
+    'exclude_conjugated_blpairs':   False,
```
nkern: will become `exclude_permutations`

scripts/pspec_red.py
```diff
@@ -0,0 +1,137 @@
...
+# Use the flags included in the UVData files
+wgts = [None for f in files]
+
+# Convert data files from Jy to mK if requested
+if 'convert_jy_to_mk' in data_cfg.keys():
+    if data_cfg['convert_jy_to_mk']:
+        for i in range(len(dsets)):
+            freqs = dsets[i].freq_array.flatten()
+            dsets[i].data_array *= beam.Jy_to_mK(freqs)[None, None, :, None]
+
```
nkern: you should also add
`dsets[i].units = 'mK'`

nkern:
Lots of great stuff in here @philbull! Thanks for prettying-up the doc strings on *a lot* of the various functions in the repo. I gave some minor feedback on nomenclature, missing info in doc-strings, and a few lines to add to various parts of the PR.

The only major feedback I have is that I think we should increase the test coverage such that coveralls is passing. Otherwise, what's the point of having it as a check, right?

hera_pspec/tests/test_pspecbeam.py
```diff
@@ -5,20 +5,6 @@
 from hera_pspec import pspecbeam
 from hera_pspec.data import DATA_PATH
 
-
```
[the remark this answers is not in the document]
philbull: It didn't seem to be doing anything. OK, I'll replace it.

nkern (approved):
looks great! thanks for getting the test coverage up!

PR #59, opened 2018-04-03
Adds UVPSpec File Format

nkern:
This adds the UVPSpec file format when carrying power spectra in-memory and when writing to disk.


nkern:
we should wait on this until the redundant baseline branch is merged in...

PR #61, opened 2018-04-05
Uvpspec file format

nkern:
Adds the `UVPSpec` file format and interfaces with `PSpecData`


hera_pspec/pspecdata.py
```diff
@@ -926,7 +1215,27 @@ def rephase_to_dset(self, dset_index=0, inplace=True):
 
         if inplace is False:
             return dsets
-            
+
+
+def get_delays(freqs):
```
philbull: Should this go in the `utils` file perhaps?

hera_pspec/pspecdata.py
```diff
@@ -727,42 +765,88 @@ def scalar(self, stokes='I', taper='none', little_h=True, num_steps=2000):
                 numerical integral
                 Default: 10000
 
+        spw_select : tuple, Example: (200, 300)
```
philbull: Maybe `spw_range` instead of `spw_select`?

hera_pspec/pspecdata.py
```diff
@@ -727,42 +765,88 @@ def scalar(self, stokes='I', taper='none', little_h=True, num_steps=2000):
...
+                                                            taper=taper, little_h=little_h, num_steps=num_steps)
+        else:
+            scalar = beam.compute_pspec_scalar(freqs[0], freqs[-1], len(freqs), stokes=stokes,
+                                                            taper=taper, little_h=little_h, num_steps=num_steps)
 
         return scalar
 
-    def pspec(self, bls, beam=None, input_data_weight='identity', norm='I', 
-              taper='none', little_h=True, verbose=True):
+    def pspec(self, bls1, bls2, dset1_ind, dset2_ind, input_data_weight='identity', norm='I', taper='none',
```
philbull: The `dset1_ind` keyword is explicit, but also a bit clunky. Can we get away with just `dset1` and `dset2`?

philbull: Ah, no. How about `dsets=(dset1, dset2)`?

hera_pspec/pspecdata.py
nkern: yeah, the `get_delays` should probably go in `utils`, b/c it has as possibly broader use-case beyond just `pspecdata`..

hera_pspec/pspecdata.py
nkern: good idea

hera_pspec/pspecdata.py
nkern: yup

hera_pspec/uvpspec.py
```diff
@@ -0,0 +1,681 @@
...
+    ant2 = int(np.floor(blpair / 1e6 - ant1*1e3))
+    ant3 = int(np.floor(blpair / 1e3 - ant1*1e6 - ant2*1e3))
+    ant4 = int(np.floor(blpair - ant1*1e9 - ant2*1e6 - ant3*1e3))
+
+    # form antnums tuple
+    antnums = ((ant1, ant2), (ant3, ant4))
+
+    return antnums
+
+def _antnums_to_blpair(antnums):
```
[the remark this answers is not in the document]
nkern: these I think should stay here for two reasons:

1) they are only used by uvpspec, not by any other module and

2) they are really only used internally to the module (hence the `_`), and aren't really exposed to the user..

hera_pspec/uvpspec.py
```diff
@@ -308,25 +338,46 @@ def key_to_indices(self, key):
...
+
+        Parameters
+        ----------
+        spws : list of spectral window integers to select
+
+        bls : list of i6 baseline integers or baseline tuples, Ex. (2, 3) 
+            Select all baseline-pairs whose first _or_ second baseline are in bls list.
+            This changes if and_bls == True.
+
+        and_bls : bool, if True, keep only baseline-pairs whose first _and_ second baseline
```
philbull: This is a bit of an awkward name.

hera_pspec/uvpspec.py
```diff
@@ -241,28 +253,46 @@ def spw_to_indices(self, spw):
...
         """
-        return np.arange(self.Nspwdlys)[self.spw_array == spw]
+        # convert to list if int
+        if isinstance(spw, (np.int, int)):
+            spw = [spw]
+
+        # assert exists in data
+        assert np.array(map(lambda s: s in self.spw_array, spw)).all(), "spws {} not all found in data".format(spw)
+
+        return np.arange(self.Nspwdlys)[reduce(operator.add, map(lambda s: self.spw_array == s, spw))]
```
philbull: `Nspwdlys` is a slightly confusing name.

PR #64, opened 2018-04-08
Fix extra taper fnc get q

nkern:
there were two calls to a tapering function in `pspecdata`, one in `get_Q` which is applied to the `Q` FT matrix, and another in `q_hat` which is applied directly to the data. Eliminating the one in `get_Q` makes OQE match the results Aaron P and I cooked up a few weeks ago, meaning that with these pushes I think OQE is properly normalized! 

Comparison with w/o tapering function (before fix)
oqe_comp_notaper

Comparison with tapering function (before fix)
oqe_comp_taper

Comparison with tapering function (after fix, title should read `fix_extra_taper_fnc_get_Q`, not `master`)
oqe_comp_taper_noq


nkern:
@philbull there was a test that was failing for `test_q_hat`, so I changed it slightly. not sure if this my changes covers what you were looking to test, but this passes. let me know your thoughts, thanks

PR #66, opened 2018-04-10
added convert_to_deltasq method to uvpspec

nkern:
Added the ability to convert a P(k) into a Delta^2. This mandated some changes to `uvpspec`, such as optionally carrying along cosmological parameters, and also made minor changes to `pspecdata` to accommodate.


hera_pspec/conversions.py
```diff
@@ -286,6 +292,51 @@ def X2Y(self, z, little_h=True):
...
+        little_h : boolean, optional
+            Whether to have cosmological length units be h^-1 Mpc or Mpc
+            Default: h^-1 Mpc
+
+        Return
+        ------
+        bl2kpara : float, conversion factor in units [h Mpc-1 / meters]
+        """
+        # Parsons 2012, Pober 2014, Kohn 2018
+        bl2kpara = 2*np.pi / (self.dRperp_dtheta(z, little_h=little_h) * (units.c / self.z2f(z)))
```
philbull: This says `kpara`, but aren't we doing `kperp`?

hera_pspec/conversions.py
```diff
@@ -286,6 +292,51 @@ def X2Y(self, z, little_h=True):
...
+
+        return bl2kpara
+
+    def tau_to_kpara(self, z, little_h=True):
+        """
+        Conversion factor from delay [seconds] to k_parallel mode [h Mpc-1] at a specified redshift.
```
philbull: For clarity, it would be good to say something like "multiply a delay [sec] by this factor to convert it to a k_parallel mode [h^-1 Mpc]", just so it's clear exactly what this factor is for.

hera_pspec/pspecdata.py
```diff
@@ -1085,11 +1094,14 @@ def pspec(self, bls1, bls2, dsets, input_data_weight='identity', norm='I', taper
                     pol_data.extend(pv.T)
 
                     # insert integration info
-                    wgts1 = (~dset1.get_flags(bl1)).astype(np.float)
-                    nsamp1 = np.sum(dset1.get_nsamples(bl1) * wgts1, axis=1) / np.sum(wgts1, axis=1).clip(1, np.inf)
-                    wgts2 = (~dset2.get_flags(bl2)).astype(np.float)
-                    nsamp2 = np.sum(dset2.get_nsamples(bl2) * wgts2, axis=1) / np.sum(wgts2, axis=1).clip(1, np.inf)
-                    pol_ints.extend(np.mean([nsamp1, nsamp2], axis=0))
+                    wgts1 = self.w(key1).T
+                    nsamp1 = np.sum(dset1.get_nsamples(bl1)[:, self.spw_range[0]:self.spw_range[1]] * wgts1, axis=1) / np.sum(wgts1, axis=1).clip(1, np.inf)
```
philbull: Please add a comment to make it clear what's being done here (it seems to be ignoring NaNs?).

hera_pspec/pspecdata.py
```diff
@@ -708,31 +708,37 @@ def p_hat(self, M, q):
...
         """
-        # Frequency units of UVData are always Hz => always convert to ns
-        delay_units = 'ns'
-        
         # Work out the power spectrum units
         if len(self.dsets) == 0:
             raise IndexError("No datasets have been added yet; cannot "
                              "calculate power spectrum units.")
+        if self.primary_beam is None:
+            pspec_units = "improper normalization"
```
philbull: Is there something a little more informative that we could write here? e.g. `Beam or normalization not specified`?

hera_pspec/uvpspec.py
```diff
@@ -160,6 +164,103 @@ def get_integrations(self, key, *args):
...
+
+        return np.array(blp_avg_sep)
+
+    def get_kvecs(self, spw, little_h=True):
+        """
+        Get cosmological wavevectors for power spectra given an adopted cosmology.
+
+        Parameters
+        ----------
+        spw : int, choice of specral window
```
philbull: Typo

hera_pspec/uvpspec.py
```diff
@@ -160,6 +164,103 @@ def get_integrations(self, key, *args):
...
+        k_perp = blpair_seps * self.cosmo.bl_to_kperp(avg_z, little_h=little_h)
+
+        # get kparas
+        k_para = self.get_dlys(spw) * self.cosmo.tau_to_kpara(avg_z, little_h=little_h)
+
+        return k_perp, k_para
+
+    def convert_to_deltasq(self, little_h=True):
+        """
+        Convert P(k) -> Delta^2(k) units.
```
philbull: Can we be explicit about what the input and output units are, please? I think this will be useful for debugging later.

hera_pspec/conversions.py
nkern: good catch! thanks

hera_pspec/conversions.py
nkern: good point

hera_pspec/pspecdata.py
nkern: sure!

hera_pspec/pspecdata.py
nkern: sure. the `clip()` is just ensuring that there is no zerodivision error for times that are entirely flagged across frequency

PR #69, opened 2018-04-19
updated pspecdata.pspec to be more condusive for bootstrapping

nkern:
`PSpecData.pspec` method no longer tries to form cross spectra between baseline groups in `bls1` and `bls2`, and exports that functionality to the `pspecdata.construct_blpairs` function.

now, `bls1` and `bls2` are lists of baseline or baseline-groups that get zipped together directly to form a `blpair` list.

future PR will introduce helper functions for creating a bootstrapping pipeline around `PSpecData.pspec`.


hera_pspec/pspecdata.py
```diff
@@ -813,55 +813,34 @@ def scalar(self, stokes='pseudo_I', taper='none', little_h=True, num_steps=2000,
         return scalar
 
     def pspec(self, bls1, bls2, dsets, input_data_weight='identity', norm='I', taper='none', 
-              little_h=True, avg_group=False, exclude_auto_bls=False, exclude_conjugated_blpairs=False,
-              spw_ranges=None, verbose=True, history=''):
+              little_h=True, spw_ranges=None, verbose=True, history=''):
```
philbull: I think it's very nice how much simpler this is!

PR #70, opened 2018-04-20
Add functions to convert freqs and redshifts to `spw_range` tuples

philbull:
Helper functions to make it easier to select spectral window ranges.


hera_pspec/utils.py
```diff
@@ -69,3 +69,126 @@ def get_delays(freqs):
...
+        frequency range, and bounds_error is False, '(None, None)' is returned. 
+    """
+    # Check for a single tuple input
+    is_tuple = False
+    if isinstance(z_range, tuple):
+        is_tuple = True
+        z_range = [z_range,]
+    
+    # Convert redshifts to frequencies (in Hz)
+    NU21CM = 1420405751.7667 # 21cm rest freq. in Hz
```
nkern: there is the variable in `conversions.units.f21` for this. although its just a constant, its nice to be self-consistent across the repo as the `conversions` module was designed to be used for this purpose

hera_pspec/tests/test_uvpspec.py
```diff
@@ -10,6 +10,84 @@
 from collections import OrderedDict as odict
 
 
+def build_example_uvpspec():
```
nkern: i like this, thanks!

hera_pspec/utils.py
```diff
@@ -69,3 +69,126 @@ def get_delays(freqs):
...
+    freq_range = []
+    for zrange in z_range:
+        zmin, zmax = zrange
+        freq_range.append( (NU21CM/(1.+zmax), NU21CM/(1.+zmin)) )
```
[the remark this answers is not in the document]
philbull: I turned it into a "static method", which lets you call the function without instantiating the class it belong to.

hera_pspec/utils.py
```diff
@@ -69,3 +69,126 @@ def get_delays(freqs):
...
+        i.e. it includes all channels in freq_range[0] <= freq < freq_range[1]. 
+        Frequencies are in Hz.
+    
+    bounds_error : bool, optional
+        Whether to raise an error if a specified lower/upper frequency is 
+        outside the frequency range available in 'data'. Default: True.
+
+    Returns
+    -------
+    spw_range : tuple or list of tuples
```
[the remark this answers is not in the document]
philbull: This is `spw_range_from_freqs`, so it returns spw indices (needed by various methods), given a set of frequencies (which users are likely to carry in their heads), i.e. freq_range [Hz] -> spw_range [indices]. Does that make sense?

nkern:
So i tried running these functions in a notebook and i'm not sure its running as expected..

philbull:
OK, this was pretty buggy -- sorry! I've fixed it now, but am still worried that the names of the functions aren't obvious.

hera_pspec/utils.py
```diff
@@ -70,6 +72,135 @@ def get_delays(freqs):
     delay = np.fft.fftshift(np.fft.fftfreq(freqs.size, d=np.median(np.diff(freqs))))
     return delay
 
+
+def spw_range_from_freqs(data, freq_range, bounds_error=True):
+    """
+    Return the first and last frequency array indices for a spectral window,
```
nkern: I think the name of the function is fine, but this line is a little confusing. The `first and last frequency array indices` sounds like I'm getting a frequency index, when really I'm getting a spectral window. For example, I can't do `freq_array[output]` with the output of this function. Perhaps 'Return the spectral window of the first and the last frequency specified in `freq_range`'

hera_pspec/utils.py
```diff
@@ -70,6 +72,135 @@ def get_delays(freqs):
...
+        spw_range.append(spw)
+    
+    # Unpack from list if only a single tuple was specified originally
+    if is_tuple: return spw_range[0]
+    return spw_range
+
+
+def spw_range_from_redshifts(data, z_range, bounds_error=True):
+    """
+    Return the first and last frequency array indices for a spectral window,
```
nkern: same comment as above

hera_pspec/utils.py
nkern: Or perhaps "Return the spectral window that each frequency specified in `freq_range` falls in"

philbull: OK, as long as the user understands "spectral window" to mean "a tuple that specifies the start and end indices of the frequency array", I think that's fine.

philbull:
Good stuff, thanks.

PR #74, opened 2018-04-21
Add ability to name datasets in PSpecData

philbull:
Adds a new `labels` argument so datasets can be referred to using strings instead of indices if desired.

Also adds `__str__` function to PSpecData to allow pretty printing.

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -1357,15 +1357,18 @@ class PSpecData(object):
         uvp.taper = taper
         uvp.norm = norm
         uvp.git_hash = version.git_hash
+        
         if self.primary_beam is not None:
             uvp.cosmo_params = str(self.primary_beam.cosmo.get_params())
         if self.primary_beam is not None and hasattr(self.primary_beam, 'filename'): 
             uvp.beamfile = self.primary_beam.filename
-        if hasattr(dset1.extra_keywords, 'filename'): uvp.filename1 = dset1.extra_keywords['filename']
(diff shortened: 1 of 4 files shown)


nkern:
Can we interface this with the `UVPSpec.tag` variable? I can make this change and push to this branch if you are ok w/ that @philbull

hera_pspec/pspecdata.py
```diff
@@ -1214,15 +1358,18 @@ def pspec(self, bls1, bls2, dsets, input_data_weight='identity', norm='I',
...
-        if hasattr(dset1.extra_keywords, 'filename'): uvp.filename1 = dset1.extra_keywords['filename']
-        if hasattr(dset2.extra_keywords, 'filename'): uvp.filename2 = dset2.extra_keywords['filename']
-        if hasattr(dset1.extra_keywords, 'tag'): uvp.tag1 = dset1.extra_keywords['tag']
-        if hasattr(dset2.extra_keywords, 'tag'): uvp.tag2 = dset2.extra_keywords['tag']
-
+        if hasattr(dset1.extra_keywords, 'filename'):
+            uvp.filename1 = dset1.extra_keywords['filename']
+        if hasattr(dset2.extra_keywords, 'filename'): 
+            uvp.filename2 = dset2.extra_keywords['filename']
+        uvp.label1 = self.labels[self.dset_idx(dsets[0])]
```
nkern: this will break if `labels` were not passed to `PSpecData` upon instantiation. can we fix this and then add a test to check this fix?

PR #80, opened 2018-04-26
Updates to how cosmology is handled within `hera_pspec`

nkern:
This updates how the passing of cosmology info is done between the various objects (i.e. all of them) in `hera_pspec`. This is done to be much clearer to the user, and be explicit about when and how cosmologies are adopted how they can be overwritten. 

In particular, if a cosmology is adopted to create a `UVPSpec` object, if it is overwritten, then the power spectrum scalar is re-calculated and the power spectra are appropriately re-normalized.


hera_pspec/noise.py
```diff
@@ -6,6 +6,63 @@
...
+    P_N = scalar * (Tsys * 1e3)^2 / (t_int * Ncoherent) / sqrt(Nincoherent)
+
+    where scalar is a nomalization given by the cosmological model and beam response, i.e. X2Y * Omega_eff
+    Tsys is the system temp in Kelvin, t_int is the integration time of the underlying data [sec], 
+    Ncoherent is the number of coherent averages before forming power spectra, and Nincoherent is the 
+    number of incoherent averages after squaring.
+
+    Parameters
+    ----------
+    Tsys : float, System temperature in Kelvin
```
philbull: `scalar` parameter isn't documented

hera_pspec/pspecdata.py
```diff
@@ -872,16 +872,21 @@ def units(self, little_h=True):
...
             raise IndexError("No datasets have been added yet; cannot "
                              "calculate power spectrum units.")
+
+        # get visibility units
+        vis_units = self.dsets[0].vis_units
+
+        # set pspec norm units
         if self.primary_beam is None:
-            pspec_units = "({})^2 Hz [beam normalization not specified]".format(self.dsets[0].vis_units)
+            norm_units = "Hz str [beam normalization not specified]".format(self.dsets[0].vis_units)
```
philbull: This has no format placeholder, e.g. {}.

PR #82, opened 2018-04-27
Add new plotting module

philbull:
Provides a general purpose 1D delay spectrum plotting function for `UVPSpec` objects. Features include:
- Plot in delay or cosmological units
- Average over baselines or times (or neither)
- Folding the x axis

Tests and an example Jupyter notebook are also included.


hera_pspec/plot.py
```diff
@@ -0,0 +1,171 @@
...
+def ax():
+    fig, (ax1, ax2) = plt.subplots(1, 2, sharex=True, sharey=True)
+    do_my_stuff_with_first_axis(ax=ax1)
+    do_my_stuff_with_second_axis(ax=ax2)
+"""
+
+def delay_spectrum(uvp, blpairs, spw, pol, average_blpairs=False, 
+                   average_times=False, fold=False, plot_noise=False, 
+                   delay=True, deltasq=False, little_h=True, 
+                   legend=False, ax=None):
```
nkern: can we add a kwarg `copy=True` which will copy the uvp object. certain operations (averaging and folding) edit the object in place, but we should be able to plot a `uvp` object without permanently changing its state. So something like
```
if copy:
    uvp = copy.deepcopy(uvp)
```
as the first line of the function

hera_pspec/plot.py
```diff
@@ -0,0 +1,171 @@
...
+    
+    delay : bool, optional
+        Whether to plot the power spectrum in delay units (ns) or cosmological 
+        units (h/Mpc). Default: True.
+    
+    deltasq : bool, optional
+        If True, plot dimensionless power spectra, Delta^2. This is ignored if 
+        delay=True. Default: False.
+    
+    little_h : bool, optional
```
nkern: This should not be a kwarg, but should be inferred from the `UVPSpec` object. the power spectrum are formed with `little_h` equal to true or false, and the plotting should not alter that state. we can either 1) make a new `UVPSpec.little_h` attribute, or you can infer it from a `h` in or not in `UVPSpec.units`...

hera_pspec/plot.py
nkern: nvm, i see that you have a `deepcopy` further down! thanks

hera_pspec/plot.py
```diff
@@ -0,0 +1,171 @@
...
+    for i, blpgrp in enumerate(blpairs_in):
+        if not isinstance(blpgrp, list):
+            blpairs.append([blpairs_in[i],])
+        else:
+            blpairs.append(blpairs_in[i])
+    
+    # Average over blpairs or times if requested
+    blpairs_in = copy.deepcopy(blpairs) # Save input blpair list
+    if average_blpairs:
+        uvp_plt = uvp.average_spectra(blpair_groups=blpairs,
```
nkern: I think this could be as simple as 
```
uvp = uvp.average_spectra(..., inplace=False)
```
and then further down you could additionally do a time average as
```
uvp = uvp.average_spectra(time_avg=True, inplace=False)
```
do you need to form `blpair_list`? I think it does that for you in the function if its empty...

Also it seems like in this structure one cannot *both* average over baselines and times? It would be nice to have
```
if average_times:
   uvp = uvp.average_spectra(..., inplace=False)
if average_blpairs:
    uvp = uvp.average_spectra(..., inplace=False)
```
so that we could do both. Also, its better to have time averaging precede blpair averaging.

philbull: I simplified it a little bit, thanks. It can already do both types of averaging at the same time -- if `average_blpairs` is True, it passes `time_avg=average_times` to `average_spectra()`.

hera_pspec/plot.py
```diff
@@ -0,0 +1,171 @@
...
+            
+            # If blpairs were averaged, only the first blpair in the group 
+            # exists any more (so skip the rest)
+            if average_blpairs: break
+    
+    # Set log scale
+    ax.set_yscale('log')
+    
+    # Plot noise spectra, if requested
+    # FIXME: This takes Tsys as an argument. So should probably be a separate
```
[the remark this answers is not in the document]
philbull: OK, I'll defer this for now. I've removed the comment from the code.

PR #84, opened 2018-04-27
Fixed window function math for tapering function

acliu:
This fixes the window functions to account for tapering. Currently it assumes a frequency-independent primary beam, but fixing this will require some changes to pyuvdata that haven't been implemented yet.


nkern:
the `get_G` call in `PSpecData.pspec` doesn't get the `taper` kwarg propagated..

bhazelton:
Out of curiosity, what do you need from `pyuvdata` that doesn't currently exist?

acliu:
Right now we have something that returns Omega_pp(\nu_i) ~ \int A^2 (\nu_i) dOmega. Turns out we'll need Omega_pp(\nu_i, \nu_j) ~ \int A(\nu_i) A(\nu_j) dOmega as well.

bhazelton:
What does the i & j index? different antennas, frequencies, or something else?

acliu:
Different frequencies.

acliu:
Ok, just pushed. Hopefully this time it's a little less buggy.

I mentioned to @nkern today that in working through this, I realized that the pspec scalar normalization  infrastructure is not currently able to accommodate non-trivial tapers + non-trivial weighting matrices simultaneously. Details in the notes that I've attached, which will be posted to the HERA memo series.

There *is* a hacky way to do this, involving assigning half (i.e., the square root) of the weighting matrix to Q and having another square root of the weight act on the data. But this is beginning to feel like epicycles upon epicycles, so if there is a need to push further along this direction I'm ripping out the old stuff and implementing Q "properly" (i.e., I'll subclass and make a version where Q is truly equal to dC/dp).

[main.pdf](https://github.com/HERA-Team/hera_pspec/files/1969547/main.pdf)

hera_pspec/pspecdata.py
```diff
@@ -681,6 +685,75 @@ def get_G(self, key1, key2):
                 G[i,j] += np.einsum('ab,ba', iR1Q[i], iR2Q[j])
 
         return G / 2.
+
+    def get_H(self, key1, key2, taper='none'):
```
philbull: I don't see `H_ab` defined anywhere in the note?

hera_pspec/pspecdata.py
```diff
@@ -681,6 +685,75 @@ def get_G(self, key1, key2):
...
+            <q_a> = \sum_b H_{ab} p_b
+
+        This is given by
+
+            H_ab = (1/2) Tr[R_1 Q_a^alt R_2 Q_b]
+
+        (See HERA memo #44). As currently implemented, this approximates the
+        primary beam as frequency independent. Under this approximation, the
+        our H_ab is defined using the equation above *except* we have
+        Q^{yet_another} rather than Q_b, where
```
philbull: `yet_another` is not a great name for a variable...

hera_pspec/pspecdata.py
```diff
@@ -750,54 +826,61 @@ def get_MW(self, G, mode='I'):
...
         # Check that mode is supported
         modes = ['G^-1', 'G^-1/2', 'I', 'L^-1']
         assert(mode in modes)
 
         # Build M matrix according to specified mode
         if mode == 'G^-1':
-            M = np.linalg.pinv(G, rcond=1e-12)
-            #U,S,V = np.linalg.svd(F)
-            #M = np.einsum('ij,j,jk', V.T, 1./S, U.T)
+            raise NotImplementedError
```
philbull: Please say what's not implemented, e.g. `NoteImplementedError("G^-1 mode not currently supported.")`

philbull:
(OK, I now see that you do raise warnings about this.)

hera_pspec/pspecdata.py
acliu: Actually, it's there. Equation 33. Unfortunate page break location makes it hard to see.

hera_pspec/pspecdata.py
acliu: Changed it to `Q^{tapered}`

hera_pspec/pspecdata.py
acliu: Done!

PR #85, opened 2018-04-27
changed "pseudo_stokes" to just "stokes" across the repo

nkern:
In accordance w/ recent `pyuvdata` PR, this PR updates all references to `pseudo_stokes` pol to just `stokes` pol (i.e. `pseudo_I` -> `I`).

### DIFF
--- a/hera_pspec/noise.py
+++ b/hera_pspec/noise.py
@@ -67,7 +67,7 @@ class Sensitivity(object):
 
         self.beam = beam
 
-    def calc_scalar(self, freqs, stokes, num_steps=5000, little_h=True):
+    def calc_scalar(self, freqs, pol, num_steps=5000, little_h=True):
         """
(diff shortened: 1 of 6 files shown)


bhazelton:
@nkern I'm actually planning to go through pyuvdata and remove all reference to Stokes because they are inaccurate and deceptive -- there is no such thing as a Stokes visibility. The only space that you can actually be in Stokes is in image space. I've been discussing this with @jaguirre and we're just trying to decide on the nomenclature we want to use.

nkern:
@bhazelton ok sounds good. thanks Bryna. I think that puts us in agreement b/c the second commit on this PR is actually changing everything from `stokes` --> `pol`

bhazelton:
ok, great!! I'll ask you for comments once I get the pyuvdata PR put together so we can try to keep things in sync.

Issue #72, opened 2018-04-20
make sure weights are called w/ identity weighting


nkern:
if `input_data_weight = 'identity'`, the weights `self.w` in `pspecdata` are not called, even if flags exist. make sure we add this in, with a master override button in case we want to assume `self.w` is always one (even if it actually isn't)

nkern:
addresed by #124

Issue #55, opened 2018-03-16
handling flagged data


nkern:
`hera_pspec.PSpecData.pspec` should throw out power-spectra when one or more of the baselines involved have completely flagged data. perhaps we can also assign a metric to each power spectrum that is the average of all the data wgts that went into forming the power spectrum, so we can make cuts on "reliability" of power spectra that have flagged data in them...

nkern:
along a similar line, right now in `PSpecData.pspec` when we choose `identity` weighting it constructs an identity matrix for the R matrix, **regardless of the weight vector**. Should we enable the I matrix to be binary (1 or 0) depending on whether the input weights are 1 or not 1?

nkern:
addressed by coarse binning (#122), RFI (#124) and decorrelation (#137)  PRs

PR #94, opened 2018-05-07
pspecdata.I now accounts for flags in data by taking time avg

nkern:
previously, `pspecdata.I()` was not accounting for data flags. this PR accounts for data flags by taking the time average of the data weights and multiplying that into the diagonal of identity.

addresses #72 and partially addresses #55

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -296,7 +296,6 @@ class PSpecData(object):
         else:
             raise TypeError("dset must be either an int or string")
     
-    
     def blkey(self, dset, bl=None, pol=None):
         """
         Return a key specifying a particular dataset, baseline, and 
@@ -337,7 +336,6 @@ class PSpecData(object):
         if pol is not None: key += (pol,)
         return key
         
-    
     def x(self, key):
         """
         Get data for a given dataset and baseline, as specified in a standard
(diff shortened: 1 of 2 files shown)


acliu:
Ok, I just checked some of the math, to see if having the non-trivial weights messes up anything with our normalization. It turns out that it does, but it can be accommodated within our current architecture. I can do this on a separate branch, or we can do it on this one if you're not in a hurry to get this done. I'll include the revised mathematical expressions in the notes. (It makes sense that there's a difference---consider a situation where the first half of the band is all flagged. Then you've effectively halved the bandwidth, and the normalization needs to know about that. For more complicated flagging/weights it's not as simple as just changing the bandwidth, but the expression can still be written down).

@nkern @philbull Are the weights here purely binary weights to indicate flagged vs. not flagged? Or are we intending to use this for more general weights as well? My preference is that `C=I` really is a pure identity matrix. We can put in a warning if the user decides to use `C=I` if there's flagged data. Essentially, my thinking is that using non-uniform weights is an example of `C \neq I`. We should be explicit about this. After all, if there's a dataset that happens to have no channels completely flagged after LST binning (as was the case in Zaki's paper, for instance), I can imagine situations where we want to compare a pure `C=I` treatment with one where we pretend certain channels are flagged to see what happens.

If the weights are intended to only be binary, I could be receptive to having them incorporated into `C=I`. But then maybe "weights" isn't the best name. Essentially, the philosophical argument is whether flagged channels should be considered normal channels that just happen to have infinite noise variance, or if flags are special and should be treated as such.

Issue #93, opened 2018-05-06
deprecate `PSpecData.C()`


nkern:
It does the same thing as `C_empirical`, so we should deprecate this to avoid confusion

philbull:
I'm not sure we want to get rid of `set_C`, as the original idea was to give users a way to set the covariance manually, e.g. if they have some better method than using the empirical estimate, like a covariance based on simulations or something. The empirical covariance is just calculated by `C` as a default.
I guess the user can still set `iC` manually though?
As long as we retain the ability to manually set the (inverse) covariance, it doesn't matter too much I think.

PR #95, opened 2018-05-07
deprecated transiting_code, legacy_pspec.py and C_empirical()

nkern:
This deprecates the `transiting_code` folder that has been hanging around during the graduation of OQE. This is now completed and therefore is no longer necessary to keep around.

This also deprecates the `legacy_pspec.py` module that was put in place in case other users wanted to interface with legacy techniques through `hera_pspec`. This was never really used (according to my knowledge), so we are deprecating it to clean things up.

Lastly, we found that within `pspecdata`, the `C_empirical()` and `C()` methods were doing the same thing of calling the `utils.cov()` function to calculate empirical covariances. We are deprecating `C()` to avoid confusion and ambiguity as to what function is called to calculate an empirical covariance.

Addresses #93


hera_pspec/pspecdata.py
```diff
@@ -431,10 +402,9 @@ def set_C(self, cov):
         self.clear_cov_cache(cov.keys())
         for key in cov: self._C[key] = cov[key]
 
-    def C_empirical(self, key):
+    def C_model(self, key, mtype='empirical'):
```
philbull: Maybe `model=` instead of `mtype=`?
I'm a bit worried about consistency here by the way. What happens if the user calls this with one model type, and then tries to use another model type later on? Is the cache dropped and recomputed in this case? Or should we store one cache per model type?

philbull: It's also unclear to me how we will support more sophisticated covariance matrix methods in future. Should we be getting the user to use `set_C` manually? Or should we attach some wrapper object, similar to `PSpecBeam`, that handles all of the covmat stuff?

nkern: Yeah that's a good point about the cache. As it stands there is only one slot for a covariance matrix, regardless of its type, so the most recently "set" covmat is the one it will pick up on. Perhaps when we re-haul inverse covariance weighting we should take a closer look at how we implement the storing of various covariance matrix types.

PR #96, opened 2018-05-07
Added noise simulation test file and unittest

nkern:
1) Created a HERA noise simulation test file in `data/`
2) Added a `testing.py` file for repo-wide access to helpful functions used in testing, that the average user might also want access to.
3) Put a `noise_sim()` function in `testing.py` to run the noise test file through OQE, and added a unit-test for `noise.py` to ensure accurate thermal errorbars to within 5%
3) eliminated stokes pol correction to `uvpspec.generate_noise_spectra`, as this averaging step should be reflected in the data itself, which would automatically propagate this through to the thermal noise estimate produced by `uvpspec.generate_noise_spectra`


nkern:
@philbull this PR is so old and there is some weird overlap, so I think I'm going to close this, branch of master and add the edits I want and open a new PR...

nkern:
@philbull Also I think we decided to not merge this b/c it wasn't an __ensemble average__ and felt a little uncomfortable with the somewhat arbitrary 5% level. I say we hold off on this for a little while until I get the ensemble averaging statistics more to our liking..

Issue #115, opened 2018-05-18
integration time of dset1 and dset2 in pspecdata should be inversely averaged


nkern:
near line 1320 of `pspecdata`, when forming the `pol_ints` it should probably be 
```
1/np.mean([1/nsamp1.clip(1e-10, np.inf), 1/nsamp2.clip(1e-10, np.inf)])
```

Issue #111, opened 2018-05-16
Make it easier to get the blpairs in a pspec object


jsdillon:
I'd suggest either adding a class object that's a list of all the blpairs or add a function like .get_blpairs().

Right now, I have to do something like `blpairs = [uvp.blpair_to_antnums(blp) for blp in uvp.blpair_array]` which is a little annoying.

Issue #106, opened 2018-05-15
Clarify spw_ranges documentation


jsdillon:
https://github.com/HERA-Team/hera_pspec/blob/729520a69ff2ceb46f9f2ad1d30497090a745d32/hera_pspec/pspecdata.py#L1088

From reading the documentation, it's not clear if spectral ranges are [start,stop), which is how numpy does things, or [start,stop] (i.e. inclusive or exclusive of the stop index).

Issue #105, opened 2018-05-15
uninformative error when trying to form a single-baseline power spectrum


jsdillon:
As a first guess during the warm up exercise, without reading the documentation too closely, I tried 
`psd.pspec((24, 25),(37, 38),[0,1])`

This gives `TypeError: object of type 'int' has no len()`, which is kind of uninformative… perhaps you can add an assert that the things that are supposed to be lists actually are lists?

PR #119, opened 2018-05-23
minor upgrades to uvpspec and pspecdata

nkern:
addressing various 5/2018 DR comments, including uvp.get_blpairs func, doc string clarify and more informative assert in pspecdata, inversely averaging nsamp1 and nsamp2 in pspecdata.pspec(), and making Jy_to_mK a method on pspecdata.

Addresses issues #104, #105, #106, #111, and #115

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -4,7 +4,7 @@ import pyuvdata
 import copy, operator, itertools
 from collections import OrderedDict as odict
 import hera_cal as hc
-from hera_pspec import uvpspec, utils, version
+from hera_pspec import uvpspec, utils, version, pspecbeam
 from hera_pspec import uvpspec_utils as uvputils
 from pyuvdata import utils as uvutils
 
@@ -1144,9 +1144,9 @@ class PSpecData(object):
             A list of spectral window channel ranges to select within the total 
             bandwidth of the datasets, each of which forms an independent power 
             spectrum estimate. Example: [(220, 320), (650, 775)].
-            Each tuple should contain a start and stop channel used to index 
-            the `freq_array` of each dataset. The default (None) is to use the 
-            entire band provided in each dataset.
+            Each tuple should contain a start (inclusive) and stop (exclusive) 
+            channel used to index the `freq_array` of each dataset. The default 
+            (None) is to use the entire band provided in each dataset.
         
         verbose : bool, optional
             If True, print progress, warnings and debugging info to stdout.
@@ -1215,6 +1215,8 @@ class PSpecData(object):
         dset2 = self.dsets[self.dset_idx(dsets[1])]
 
         # assert form of bls1 and bls2
+        assert isinstance(bls1, list), "bls1 and bls2 must be fed as a list of antpair tuples"
+        assert isinstance(bls2, list), "bls1 and bls2 must be fed as a list of antpair tuples"
         assert len(bls1) == len(bls2), "length of bls1 must equal length of bls2"
         for i in range(len(bls1)):
             if isinstance(bls1[i], tuple):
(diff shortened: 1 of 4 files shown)


hera_pspec/pspecdata.py
```diff
@@ -1399,8 +1401,8 @@ def pspec(self, bls1, bls2, dsets, pols, input_data_weight='identity', norm='I',
                     nsamp1 = np.sum(dset1.get_nsamples(bl1 + (p[0],))[:, self.spw_range[0]:self.spw_range[1]] * wgts1, axis=1) / np.sum(wgts1, axis=1).clip(1, np.inf)
                     nsamp2 = np.sum(dset2.get_nsamples(bl2 + (p[1],))[:, self.spw_range[0]:self.spw_range[1]] * wgts2, axis=1) / np.sum(wgts2, axis=1).clip(1, np.inf)
 
-                    # take average of nsamp1 and nsamp2 and multiply by integration time [seconds] to get total integration
-                    pol_ints.extend(np.mean([nsamp1, nsamp2], axis=0) * dset1.integration_time)
+                    # take inverse average of nsamp1 and nsamp2 and multiply by integration time [seconds] to get total integration
```
philbull: Why are we taking the inverse average now? It would be good to add a comment to document why we need to do this.

hera_pspec/pspecdata.py
```diff
@@ -1505,12 +1507,11 @@ def pspec(self, bls1, bls2, dsets, pols, input_data_weight='identity', norm='I',
     def rephase_to_dset(self, dset_index=0, inplace=True):
         """
         Rephase visibility data in self.dsets to the LST grid of dset[dset_index] 
-        using hera_cal.utils.lst_rephase. 
+        using hera_cal.utils.lst_rephase. Each integration in all other dsets are
```
philbull: "are" => "is"

hera_pspec/pspecdata.py
```diff
@@ -1505,12 +1507,11 @@ def pspec(self, bls1, bls2, dsets, pols, input_data_weight='identity', norm='I',
...
-        using hera_cal.utils.lst_rephase. 
+        using hera_cal.utils.lst_rephase. Each integration in all other dsets are 
+        phased to the center of the corresponding LST bin (by index) in dset[dset_index].
 
-        Each integration in all other dsets are phased to the center of the 
-        corresponding LST bin (by index) in dset[dset_index].
-
-        Will only phase if the dataset's phase type is 'drift'.
+        Will only phase if the dataset's phase type is 'drift'. Note that if you intend
+        to use Jy_to_mK(), it must be run after rephase_to_dset().
```
philbull: It would be good to mention why.

hera_pspec/pspecdata.py
```diff
@@ -1597,6 +1598,45 @@ def rephase_to_dset(self, dset_index=0, inplace=True):
         if inplace is False:
             return dsets
 
+    def Jy_to_mK(self, beam=None):
```
philbull: Should the example Jupyter notebook be changed to use this method now?

hera_pspec/uvpspec.py
```diff
@@ -335,6 +335,29 @@ def get_kparas(self, spw, little_h=True):
...
 
+    def get_blpairs(self):
+        """
+        Returns a list of all blpair tuples in the data_array.
+        """
+        return [self.blpair_to_antnums(blp) for blp in np.unique(self.blpair_array)]
+
+    def get_all_keys(self):
+        """
+        Returns a list of all possible tuple keys in the data_array.
```
philbull: It would be good to say what the keys are, i.e. (spectral window, baseline pair, polarisation)

hera_pspec/pspecdata.py
nkern: sure. the inverse average is done b/c the `nsample` is actually a proxy for `1/sigma`, where `sigma` is the std of the noise. so we really should be averaging `sigma` directly, which is equivalent to inversely averaging `nsamples`. I'll make a comment explaining this.

hera_pspec/pspecdata.py
nkern: yup

nkern: except that to run on our test data another small bug fix is required that is addressed in a separate PR open right now. I'll update the notebook at a (near) future date incorporating all the changes we are making in this and a few other simultaneous PRs.

PR #120, opened 2018-05-30
adds pspecdata.pspec_run

nkern:
and a drive script in `scripts/pspec_run.py` to run OQE from the command line.


hera_pspec/pspecdata.py
```diff
@@ -1640,79 +1642,282 @@ def Jy_to_mK(self, beam=None):
...
+        exclude_auto_bls is True, eliminate all instances of a bl crossed 
+        with itself.
+
+    exclude_permutations : boolean
+        If blpairs is None, redundant baseline groups will be formed and
+        all cross-multiplies will be constructed. In doing so, if
+        exclude_permutations is True, eliminates instances of
+        (bl_B, bl_A) if (bl_A, bl_B) also exists.
+
+    group : boolean
```
philbull: Do we need this argument? We can just assume that `group=True` if `Nbls_per_group` is not None, can't we?

hera_pspec/pspecdata.py
```diff
@@ -1640,79 +1642,282 @@ def Jy_to_mK(self, beam=None):
...
+                    _bls2.append(_bl2)
+
+            bls1_list.append(_bls1)
+            bls2_list.append(_bls2)
+
+    # Open PSpecContainer to store all output in
+    psc = container.PSpecContainer(filename, mode='rw')
+
+    # assign group name
+    if groupname is None:
```
philbull: Not sure about this, but should we allow a prefix for groupname to be specified? This would let the user save power spectra into a given top-level group in the PSpecContainer, while maintaining the nice automatic `dset1_dset2` naming, e.g. to store spectra in `toplevel/dset1_dset2/...`.

hera_pspec/pspecdata.py
```diff
@@ -1640,79 +1642,282 @@ def Jy_to_mK(self, beam=None):
...
+        Contains UVData objects or string filepaths to miriad files
+
+    filename : str
+        Output filepath for HDF5 PSpecContainer object
+
+    groupname : str
+        Groupname of the subdirectory in the HDF5 container to store the 
+        UVPSpec objects in. Default is a concatenation the dset_labels.
+
+    dset_labels : list
```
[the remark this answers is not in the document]
nkern: there is a default method. i'll add it to the docstring

hera_pspec/pspecdata.py
nkern: true

hera_pspec/pspecdata.py
nkern: I think that's a good idea. in a different PR I'm going to actually slightly reformat the way I handle groupnames for the pspec pipeline, so I'll wait until then to make this change, but I'll keep it in mind while I work on it.

PR #121, opened 2018-06-02
Data flow pipeline

nkern:
Adds scripts and edits modules for the IDR2.1 data-preprocessing pipeline

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -1650,6 +1650,44 @@ class PSpecData(object):
                 dset.data_array[:, :, :, j] *= factors[p][None, None, :]
             dset.vis_units = 'mK'
 
+    def trim_dset_lsts(self, lst_tol=6):
+        """
+        Assuming all datasets in self.dsets are locked to the same LST grid (but
+        each may have a constant offset), trim LSTs from each dset that aren't found
+        in all other dsets (within some decimal tolerance specified by lst_tol). 
+
+        Warning: this edits the data in dsets in-place, and is not reversible.
+
+        Parameters
+        ----------
+        lst_tol : float
+            Decimal tolerance [radians] for comparing float-valued LST bins.
+        """
+        # ensure each dset has same dLST within tolerance / Ntimes
+        dlst = np.median(np.diff(np.unique(self.dsets[0].lst_array)))
+        for dset in self.dsets:
+            _dlst = np.median(np.diff(np.unique(dset.lst_array)))
+            if not np.isclose(dlst, _dlst, atol=10**(-lst_tol) / dset.Ntimes):
+                print "not all datasets in self.dsets are on the same LST grid, cannot LST trim."
+                return
+
+        # get lst array of each dataset and turn into string and add to common_lsts
+        lst_arrs = []
+        common_lsts = set()
+        for i, dset in enumerate(self.dsets):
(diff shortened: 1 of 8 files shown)


hera_pspec/pspecdata.py
```diff
@@ -1640,6 +1640,44 @@ def Jy_to_mK(self, beam=None):
                 dset.data_array[:, :, :, j] *= factors[p][None, None, :]
             dset.vis_units = 'mK'
 
+    def trim_dset_lsts(self, lst_tol=6):
```
philbull: Is this something that could be done with `select_common` instead?

hera_pspec/pstokes.py
```diff
@@ -269,3 +269,48 @@ def construct_pstokes(dset1, dset2, pstokes='pI', run_check=True, antenna_nums=N
         uvdS.check()
 
     return uvdS
+
+
+def choose_inp_dsets(dsets, pstokes):
```
philbull: I think this function probably needs a `pol` in its name somewhere, as it's all about selecting based on polarisation. Also, can ambiguities arise, when more than one UVData provides the necessary polarizations?

hera_pspec/tests/test_pstokes.py
```diff
@@ -75,5 +75,12 @@ def test_construct_pstokes(self):
         # check except for same polarizations
         nt.assert_raises(AssertionError, pstokes.construct_pstokes, dset1=uvd1, dset2=uvd1, pstokes='pI')
 
+    def test_choose_inp_dsets(self):
```
philbull: It'd be good to add a couple more tests to handle situations where there might be multiple ways of building the same Stokes parameter.

pipelines/idr2_preprocessing/preprocess_data.py
```diff
@@ -0,0 +1,379 @@
...
+    exit_codes = M(xtalk_sub, range(len(datafiles)))
+
+    # print to log
+    hp.utils.log("\nxtalk subtraction exit codes:\n {}".format(exit_codes), f=lf, verbose=verbose)
+
+    input_data_template += file_ext
+
+    time = datetime.utcnow()
+    hp.utils.log("\nfinished crosstalk subtraction: {}\n{}".format(time, "-"*60), f=lf, verbose=verbose)
+
```
philbull: It's not really important right now, but it might be good in future to keep tracking of simple running-time stats for each block here. Runtimes seem to have been a useful diagnostic in the Analysis pipeline.

hera_pspec/pspecdata.py
nkern: that operates on `UVPSpec`, if I'm not mistaken. this is supposed to operate on `UVData` objects. We could make `select_common` work on `UVData`, but it seems like its use-case is really relegated to just `UVPSpec`..

hera_pspec/pstokes.py
nkern: Okay! I'll put a `pol` in the name. Yes that ambiguity can arise and its resolution is stated in the docstring as `If a single UVData object 
    has multiple polarizations, this function only uses its first.`

pipelines/idr2_preprocessing/preprocess_data.py
nkern: for each block do you mean for each call to the run function? or each analysis choice? b/c I do the latter currently..

hera_pspec/tests/test_pstokes.py
nkern: sure thing

PR #122, opened 2018-06-03
Coarse delay binning

acliu:
This PR provides the ability to use thick delay bins, so that there are fewer delay bins than frequency channels.

Note that there is a new adjustment to the normalizing scalar that's implemented as an extra multiplicative factor. I can provide notes to justify the mathematics of this if people are curious. The good news is that I think I can generalize the same function to account for RFI flags, so contrary to earlier expectations, we probably won't need to change Pspecbeam when we add flagging capabilities.

I would particularly appreciate some close scrutiny for the parts in pspec where one deals with multiple spws (and therefore multiple Ndlys).

Thank you!


acliu:
Ok, and here are the promised notes. Section 6.
[main.pdf](https://github.com/HERA-Team/hera_pspec/files/2070501/main.pdf)

PR #124, opened 2018-06-06
RFI

acliu:
Here's the PR for dealing with RFI flags.

Something that occurred to me while I was putting this together: we probably want to enforce that if a channel is flagged in x_1 but not x_2 (or vice versa), the channel is flagged for both. We can either try to enforce this in the pspec code, or we could do it in the preprocessing.

Thank you!

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -580,11 +580,12 @@ class PSpecData(object):
         data covariance matrix (I or C^-1), diagonal flag matrix (Y) and 
         diagonal tapering matrix (T):
 
-        R = T^t Y^t W Y T
+        R = sqrt(T^t) sqrt(Y^t) K sqrt(Y) sqrt(T)
 
-        where T is a diagonal matrix holding sqrt(taper) and Y is a diagonal
-        matrix holding sqrt(flags). The W matrix comes from either I or iC
-        depending on self.data_weighting, and T is informed by self.taper.
+        where T is a diagonal matrix holding the taper and Y is a diagonal
+        matrix holding flag weights. The K matrix comes from either I or iC
+        depending on self.data_weighting, T is informed by self.taper and Y
+        is taken from self.Y().
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


hera_pspec/pspecdata.py
```diff
@@ -517,12 +552,15 @@ def set_R(self, R_matrix):
...
             If set to "identity", sets R = I
             If set to "iC", sets R = C^-1
+            If set to "identity_with_flags", sets R = diagonal according to pspecdata weights
             Otherwise, accepts a user inputted dictionary
         """
         if R_matrix == "identity":
             self.R = self.I
         elif R_matrix == "iC":
             self.R = self.iC
+        elif R_matrix == "diagonal":
```
nkern: the docstring above says `identity_with_flags` but what's actually caught here is `diagonal`.

nkern:
So it seems like we are still (under the hood) setting `R = I` or `R = iC` and now `R = inv_diag`. I was hoping we could do something like `R = I * Y` or `R = iC * Y`, which is what I thought we concluded in our last meeting. If its okay with you, can I make a push to this branch implementing this within the structure that you've already got here? At the same time, I'll implement the RFI-harmonization method in `PSpecData` that we discussed earlier in the week.

hera_pspec/pspecdata.py
acliu: Whoops. Fixed the docstring.

acliu:
My apologies---I owe you an explanation about this.

The thing, though, is that after I typed up a long paragraph for why I came to the conclusion that `R = iC * Y` is very dangerous (and therefore the `Y` structure is not necessary since a diagonal `R` would capture the `R = I * Y` case), I realized that the math argument I was using was just plain wrong :)

So yes, I am fine with what you suggest. Also happy to help out this afternoon with implementation if you like.

nkern:
@acliu okay adrian, take a look at what I've got and let me know if you have any comments. The major things I changed were:

1) got rid of `diag_inv_cov` cache for a general `_R` cache which is `sqrt(T)^t sqrt(Y)^t A sqrt(Y) sqrt(T)` where `A` is either `I` or `iC` depending on `self.data_weighting`. To do this, I got rid of `set_R` but made a similar `set_weighting` method (also just made a `set_taper` method b/c that seems to be the way we are going w/ these things).

2) I also added a `broadcast_dset_flags` method, which does the merging of time-dependent flags that we talked about earlier in the week

In terms of your comment about if a freq channel is flagged in `x_1` we flag it in `x_2`: I actually don't think that is necessary or even desired. why not allow information in `x_2` to influence `FT(x_2)`, even if a channel is flagged in `x_1`?

PR #139, opened 2018-06-25
Added label_*_array to _meta_dsets

duncanjmr:
modified:   hera_pspec/uvpspec.py

I was having problems using UVPSpec.write_to_group because label_1_array and label_2_array are too large to be written as an attr using HDF5, I was getting an "object header message is too large" error. I added them to _meta_dsets in order for them to be saved properly. A small change.

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -105,7 +105,8 @@ class UVPSpec(object):
 
         # define which attributes are considred meta data. Large attrs should be constructed as datasets
         self._meta_dsets = ["lst_1_array", "lst_2_array", "time_1_array", "time_2_array", "blpair_array", 
-                            "bl_vecs", "bl_array", 'lst_avg_array', 'time_avg_array', 'OmegaP', 'OmegaPP']
+                            "bl_vecs", "bl_array", 'lst_avg_array', 'time_avg_array', 'OmegaP', 'OmegaPP',
+                            "label_1_array", "label_2_array"]
         self._meta_attrs = sorted(set(self._all_params) - set(self._dicts) - set(self._meta_dsets))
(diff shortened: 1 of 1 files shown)


nkern:
@duncanjmr interesting you are seeing that. I wasn't anticipating these arrays to contain more than a handful of elements. Can you paste an example print statement of the first couple dozen elements of an array that gives you this error?

duncanjmr:
@nkern The label_1_array is all zeros, they are supposed to be indices pointing to a string that is in the labels array. For a UVPSpec object that has 5460 blpairts, one spectral window, and one polarization, the shape of label_1_array and label_2_array is (1, 5460, 1). However for larger blpairts, this quickly exceeds the 64KB header limit set by HDF5. I should also note that the select function does not reduce the size of this array. Though the uvpspec.py file says that label_1_array should be the same size as data_array, the select function can break this requirement by reducing data_array but not label_1_array.

In the following case, uvpl is just a list of UVPSpecs. Note that all of the entries are equal to zero.

image


And here is the error in full.:
image

And finally, here is the section of code that triggers the error (lines 970-979 ish).

```
        for k in self._meta_attrs:
            if hasattr(self, k):
                # handle cosmo
                if k == 'cosmo':
                    group.attrs['cosmo'] = str(getattr(self, k).get_params())
                    continue
                group.attrs[k] = getattr(self, k)
```

I can confirm that putting label_1_array and label_2_array into _meta_dsets fixes this error, in the case I have found it to arise.

nkern:
@duncanjmr ah you are totally right! It is the `UVPSpec.labels` array that should only have a handful of elements, and indeed the `label_1_array` will have an `Nblpts` axis and should be in the `meta_dsets` list. Good catch!

PR #141, opened 2018-06-28
Stats array

duncanjmr:
Added new class dictionary UVPSpec.stats_array.  This is a dictionary of dictionaries, that is, one key for the statistic, one for the spectral window. Stats_array is an optional housing for errorbars, statistical values, or pretty much anything that matches the shape of the data. One can add as many values to stats_array as one likes, by making each new array under a new key. Integrated with select function and average_spectra function. Average_spectra does a weighted sum over the errors for a requested field to shrink the size.

Added
- UVPSpec.stats_array
- get_stats, set_stats
- Updated average_spectra, select, and write_to_group

Currently the master build is failing, so this will definitely fail as well.


hera_pspec/uvpspec.py
```diff
@@ -42,6 +42,9 @@ def __init__(self):
         desc = "Nsample dictionary, if the pspectra have been incoherently averaged (i.e. averaged after squaring), this is " \
                "the effective number of samples in that average (float type). This is not the same as the pyuvdata.UVData nsample_array."
         self._nsample_array = PSpecParam("nsample_array", description=desc, expected_type=np.float64, form="(Nblpairts, Npols)")
+        desc = ("Power spectrum stats array with stats type and spw integer as keys and values as complex ndarrays with same shape"
+                " as data_array")
+        self._stats_array = PSpecParam("stats_array", description=desc, expected_type=dict, form="(Nblpairts, Ndlys, Npols)")
```
nkern: `expected_type` should be `np.complex128`

hera_pspec/uvpspec.py
```diff
@@ -1090,6 +1165,13 @@ def check(self, just_meta=False):
                             except:
                                 raise AssertionError(err_msg)
 
+                    elif p in self._dicts_of_dicts:
+                        assert isinstance(getattr(self, p), (dict, odict))
+                        for k in getattr(self, p).keys():
+                            assert isinstance(getattr(self, p)[k], (dict, odict))
+                            for j in getattr(self, p)[k].keys():
+                                assert isinstance(getattr(self, p)[k][j], np.ndarray)
+
```
nkern: should add here:
```
try:
    setattr(self, p, a.expected_type(getattr(self, p)))
except:
    raise AssertionError(err_msg)
```

hera_pspec/grouping.py
```diff
@@ -339,7 +346,14 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
             except:
                 raise IndexError("blpair_weights must have the same shape as "
                                  "blpair_groups")
-    
+
+    stat_l = []
```
nkern: can you add a brief comment here just describing what `stat_l` is and what this block of code is doing

hera_pspec/grouping.py
```diff
@@ -339,7 +346,14 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+    if error_field is not None:
+        stat_l = [error_field]
```
nkern: if `error_field` can be a list, this should probably be something like
```
if error_field is not None:
    if isinstance(error_field, (list, tuple, np.ndarray)):
        stat_l = list(error_field)
    elif isinstance(error_field, (str, np.str)):
        stat_l = [error_field]
```

hera_pspec/grouping.py
```diff
@@ -386,6 +405,14 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     wgts = uvp.get_wgts(spw, blp, p)
                     ints = uvp.get_integrations(spw, blp, p)[:, None]
                     w = (ints * np.sqrt(nsmp))
+
+                    # Get error bar weights and set invalid ones to zero.
```
nkern: because you don't use `errs` further down, just set as `errws` to being with so there is no confusion.

What abour errs that are != -99 but are < 0? Seems like the third line should actually read
```
errs['errs < 0'] = 0.0
```

hera_pspec/grouping.py
```diff
@@ -386,6 +405,14 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                    errws = {}
+                    for stat in stat_l:
+                        errs = uvp.get_stats(stat, (spw, blp, p))
+                        errs[np.where(errs > 0.)] = 1. / errs[np.where(errs > 0.)] ** 2.
```
nkern: lot of repeated indexing here, you should do the indexing once, store it and then reference it later on. something like
```
for stat in stat_l:
    errw = uvp.get_stats(stat, (spw, blp, p))
    errw_bools = errs > 0.0
    errw[errw_bools] = 1.0 / errw[errw_bools] ** 2.0
    errw[~errw_bools] = 0.0
    errws[stat] = errw
```

hera_pspec/grouping.py
```diff
@@ -417,20 +448,32 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                 bpg_ints = np.sum(bpg_ints, axis=0) \
                          / np.sum(w_list, axis=0).clip(1e-10, np.inf)
                 w_list = np.sum(w_list, axis=0)
+
+                # For errors that aren't 0, sum weights and take inverse square
+                # root. Otherwise, set to invalid value -99.
+                for stat in stat_l:
+                    arr = np.sum(bpg_stats[stat], axis=0)
+                    arr[np.where(arr == 0.)] = -99.
```
nkern: Should stay away from direct float comparison. Something like
```
arr[np.isclose(arr, 0.0)] = -99.
```
is better

hera_pspec/grouping.py
```diff
@@ -386,6 +412,15 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     wgts = uvp.get_wgts(spw, blp, p)
                     ints = uvp.get_integrations(spw, blp, p)[:, None]
                     w = (ints * np.sqrt(nsmp))
+
+                    # Get error bar weights and set invalid ones to zero.
+                    errws = {}
+                    for stat in stat_l:
+                        errs = uvp.get_stats(stat, (spw, blp, p))
```
nkern: This fix is now more efficient, but it still doesn't address the case when `errs` has negative, but `!=-99` data. This is why I think its better to use `no_data = errs < 0`, which will also capture when `errs == -99`

hera_pspec/uvpspec.py
nkern: Ah yeah you are right, what I think I meant was something like
```
# dict of dicts
elif p in self._dicts_of_dicts:
    assert isinstance(getattr(self, p), (dict, odict))
    for k in getattr(self, p).keys():
        assert isinstance(getattr(self, p)[k], (dict, odict))
        for j in getattr(self, p)[k].keys():
            assert isinstance(getattr(self, p)[k][j], np.ndarray)
            try:
                setattr(self, getattr(self, p)[k][j], a.expected_type(getattr(self, p)[k][j]))
            except:
                raise AssertionError(err_msg)
```
which will work on a dictionary of dictionaries.

hera_pspec/grouping.py
philbull: We should set invalid values to `np.nan`, shouldn't we?

nkern: We could use `nan`s. Perhaps that is a safer solution. In that case, @duncanjmr, your indexing arrays would become `no_data = np.isnan(errs)`.

nkern: Also, @duncanjmr, because you are indexing a mutable variable, you will want to ensure you aren't accidently overwriting the parent `uvp` object by using
```
errs = uvp.get_stats(stat, (spw, blp, p)).copy()
```

nkern:
@anguta ^got it down to one-commit-and-done passing travis :)

hera_pspec/grouping.py
```diff
@@ -416,32 +472,54 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
                 bpg_ints = np.sum(bpg_ints, axis=0) \
                          / np.sum(w_list, axis=0).clip(1e-10, np.inf)
+                if store_cov:
+                    bpg_cov = np.sum(bpg_cov, axis=0) \
+                            / np.sum(w_list, axis=0)[:,None].clip(1e-10, np.inf)
                 w_list = np.sum(w_list, axis=0)
-                
+
+                # For errors that aren't 0, sum weights and take inverse square
+                # root. Otherwise, set to invalid value -99.
```
nkern: this comment needs to get updated to reflect `nan` for invalid values

hera_pspec/grouping.py
```diff
@@ -416,32 +472,54 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                for stat in stat_l:
+                    arr = np.sum(bpg_stats[stat], axis=0)
+                    not_valid = np.isclose(arr, 0., 1e-10)
```
nkern: I still think this should be
```
not_valid = np.isclose(arr, 0, 1e-10) + arr < 0
```
because the error estimation may have done something weird and assigned a non-zero but negative error and that will screw up the averaging downstream...

hera_pspec/grouping.py
```diff
@@ -348,66 +370,100 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
                     ints = uvp.get_integrations(spw, blp, p)[:, None]
                     w = (ints * np.sqrt(nsmp))
-                    
+                    if store_cov:
+                        cov = uvp.get_cov(spw,blp,p)
+                    # Get error bar weights and set invalid ones to zero.
+                    errws = {}
+                    for stat in stat_l:
+                        errs = uvp.get_stats(stat, (spw, blp, p)).copy()
+                        no_data = np.isnan(errs)
```
nkern: I think this should be
```
no_data = np.isnan(errs) + (errs < 0)
```
just to guard against weird cases when `errs` is negative, which I don't think should ever be the case.

hera_pspec/uvpspec.py
```diff
@@ -80,44 +83,91 @@ def __init__(self):
...
+            Baseline-pair key
+
+        Returns
+        -------
+        data : complex ndarray
+            Shape (Ntimes, Ndlys, Ndlys)
+        """
+        spw, blpairts, pol = self.key_to_indices(key, *args)
+        # Need to deal with folded data!
+        # if data has been folded, return only positive delays
```
nkern: this comment is out of place: this return statement doesn't deal w/ folded data. we should add a line that does properly deal w/ folded data

hera_pspec/uvpspec.py
```diff
@@ -394,6 +444,63 @@ def get_spw_ranges(self, spws=None):
...
+
+        if not hasattr(self, "stats_array"):
+            self.stats_array = odict()
+
+        dtype = statistic.dtype
+        if stat not in self.stats_array.keys():
+            self.stats_array[stat] = odict([[i, np.nan * np.ones(self.data_array[i].shape, dtype=dtype)]
+                                             for i in range(self.Nspws)])
+
+        self.stats_array[stat][spw][blpairts, :, pol] = statistic
```
nkern: we should add a conditional to catch if the spectra have been folded. see the `get_data` function for examples

hera_pspec/grouping.py
```diff
@@ -470,50 +548,56 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
     uvp.data_array = data_array
     uvp.integration_array = ints_array
     uvp.wgt_array = wgts_array
     uvp.nsample_array = nsmp_array
+    if store_cov:
+        uvp.cov_array = cov_array
     if hasattr(uvp_in, 'label1'): uvp.label1 = uvp_in.label1
     if hasattr(uvp_in, 'label2'): uvp.label2 = uvp_in.label2
 
+    if error_field is not None:
```
nkern: this logic of the following four lines doesn't make sense to me. can you double check this is indeed doing what you intend?

hera_pspec/uvpspec.py
anguta: I've added covariance matrix folding and get_cov now deals with the folded case so this comment is now appropriate :)

nkern: great thanks!

duncanjmr: @nkern I kind of skirted over the folded data line (and added this comment to the code) because I was unsure of how to deal with folded data. Currently stats array isn't touched by grouping.fold_spectra, so it wouldn't make sense to only use the second half of the errors, as is done in get_data if self.folded is True. What to do to grouping.fold_spectra is something I should talk about with phil or you.

hera_pspec/grouping.py
duncanjmr: This is the logic:

1. Check if input parameter "error" field is None. If it is not None, set attribute stats array to the object just constructed in the long for loops.
2. If error_field is None, but the uvp has an attribute called stats_array, delete it.  In other words, a uvp was given with a stats array but the user did not specify error bars to keep, so stats_array should not be in the output uvp at all.
3. If error_field is None _and_ stats_array is not an attribute of uvp, do nothing.

I hope this makes sense!

duncanjmr:
@anguta I am pushing a new commit that has added store_cov as a PSpecParameter, because of the following error. I traced it to the fact that store_cov is not in _all_params, and therefore not saved, because its never specified as a PSpecParam. However you might have a better idea of what you would like store_cov to be, this is just my temporary fix.

image

hera_pspec/uvpspec.py
duncanjmr: Ah, this response should be for your comment on get_stats, not get_cov. My mistake, and I guess the comment wasn't mine oops.

nkern:
@duncanjmr About `stats_array` access when `folded=True`: Folding is just an averaging step (just requires some fancy indexing gymnastics), so we would treat the `stats_array` the same way we would treat it in the averaging.

Also, I think we decided that `store_cov` should _not_ be a `PSpecParam`, but because `cov_array` is optional, its presence (or lack thereof) is itself a `store_cov`-like parameter.

nkern:
I can make these edits and push to the branch...

hera_pspec/grouping.py
nkern: Okay, that makes sense to me. thanks!

nkern: except this can fail if `uvp` does not have a stats_array

duncanjmr:
Whoa, did I just break something? I don't recognize any of these errors, does anyone else?

philbull:
These errors seem to be associated with `pyuvdata` and `hera_cal` API changes...

philbull:
OK, so it looks like the `pyuvdata` API changed significantly either today or yesterday. This is causing a breakage in `hera_cal.utils.lst_rephase`, which is called from `pspecdata.py`. So we need to wait for `hera_cal` to be fixed before we can move forward...

jsdillon:
Yeah... @nkern is working on it. See https://github.com/HERA-Team/hera_cal/issues/341

philbull:
Since (1) the review comments have been addressed; (2) this is a prerequisite for Nick's pspec pipeline PR; and (3) the errors seem to be due to upstream changes, I'm going to merge this. If it does cause any breakages, we can fix them.

PR #143, opened 2018-06-29
pspecbeam.PSpecBeamUV updates for efield beam

nkern:
Updates the beam files in the `hera_pspec/data` directory, and allows `pspecbeam.PSpecBeamUV` to take an efield beam and convert to power beams.


hera_pspec/pspecbeam.py
```diff
@@ -429,7 +440,7 @@ def power_beam_sq_int(self, pol='pI'):
         primary_beam_area: float, array-like
         """
         if hasattr(self.primary_beam, 'get_beam_area'):
-            return self.primary_beam.get_beam_sq_area(pol)
+            return np.real(self.primary_beam.get_beam_sq_area(pol))
```
philbull: Should this function be returning complex numbers to begin with? Or is there a bug in `get_beam_sq_area`?

nkern: That's what I was asking @bhazelton, and I never got a response, but I think I came to the conclusion that in principle it can be complex (e.g. when asking for `XY` and `YX` areas), but in practice we will always be using `XX` or `YY` or `pI` or `pQ` beams, in which case it is real. But @bhazelton let me know if you have any opinions on that.

hera_pspec/pspecbeam.py
```diff
@@ -358,30 +358,41 @@ def power_beam_sq_int(self, pol='pI'):
...
+        self.beam_freqs = uvb.freq_array[0]
         if cosmo is not None:
             self.cosmo = cosmo
         else:
             self.cosmo = conversions.Cosmo_Conversions()
 
+        # setup primary power beam
+        self.primary_beam = uvb
+        if uvb.beam_type == 'efield':
+            self.primary_beam.efield_to_power(inplace=True)
```
Chuneeta: efield_to_power is called for linear polarizations, for pseudo-stokes you need to use the function efield_to_pstokes.

nkern: So this is done assuming one only wants linear dipole polarization. This function is (purposely) _not_ trying to be smart and _will not_ convert the efield to pstokes. To `pspecbeam` for pstokes parameters, one needs to feed a pstokes power beam. (see the unit tests added where I test for exactly this case). I'll make a comment specifying this in the docstring!

Issue #142, opened 2018-06-28
uvpspec.combine_uvpspec taking very long


duncanjmr:
Calling combine_uvpspec or adding using uvp1 + uvp2 takes very long, Phil suggested that it might be possible to optimize it or something.

nkern:
I'll take a look at this too, b/c i wrote it..

nkern:
Can you give me an example of some `UVPSpec` objects and how many of them you are using? The `test_combine_uvpspec` in `test_uvpspec.py` doesn't take that long at all...

philbull:
This seems to be an issue for UVPSpec objects with lots of blpairs. I did some profiling on an object with 50 blpairs x 200 time samples, and it takes about 9 seconds to run on my computer (which seems like a long time considering the size of the blpairts array index is only 10,000 elements). I'll attach a notebook to demonstrate.

Anyway, the issue seems to be the many calls to the `.index()` method, which seems to be a relatively slow search. We can probably optimise this.

philbull:
[Test_UVPSpec_addition.zip](https://github.com/HERA-Team/hera_pspec/files/2147977/Test_UVPSpec_addition.zip)

philbull:
@duncanjmr and I have been working on several optimisations this afternoon, which should make a significant difference. I'll get those pushed up in the next day or two.

philbull:
OK, see PR #144 for a solution.

philbull:
Solved by PR #144.

PR #144, opened 2018-06-30
Speed up UVPSpec add() operation

philbull:
This PR attempts to address issue #142 -- adding `UVPSpec` objects together takes a long time if there are a lot of blpair-times.

After profiling, the main issue seems to be lots of calls to the `index()` method on long lists. This PR replaces index() calls with `numpy.where()` calls and some `dict`-based lookups, which turn out to be much faster,


nkern:
Overall this looks good and is a nice speedup for the `combine_uvp` functions. I have one qualm with the new "hack" approach to querying the baseline-time axis, and that is by using a `np.where` on a float array, which does direct float-float comparison, which is in general not a good idea. Can we think of a different approach that uses something like an `np.isclose`, or takes the `pyuvdata` approach and turns the `time_array` into a string with a set number of sig figs and does the comparison on strings rather than floats? Let me know what you think @philbull

hera_pspec/uvpspec_utils.py
```diff
@@ -366,3 +364,37 @@ def _conj_blpair(blpair, which='both'):
...
     return conj_blpair
 
+def _fast_lookup_blpairts(src_blpts, query_blpts):
+    """
+    Helper function to allow fast lookups of array indices for large arrays of 
+    blpair-time tuples.
+    
+    Parameters
+    ----------
+    src_blpts : array_like
```
nkern: This says `array_like`, but the description makes it seem like what you actually want is a list, something like
```
src_blpts = [(blp1, time1), (blp2, time2), ...]
```
in which case the data type should be `list` not `array_like`

hera_pspec/uvpspec_utils.py
```diff
@@ -366,3 +364,37 @@ def _conj_blpair(blpair, which='both'):
...
+    Returns
+    -------
+    blpts_idxs : array_like
+        Array of integers of size (M,), which are indices in the source_blpts 
+        array for each item in query_blpts.
+    """
+    # This function works by using a small hack -- the blpair-times are turned 
+    # into complex numbers of the form (blpair + 1.j*time), allowing numpy 
+    # array lookup functions to be used
+    src_blpts = np.array(src_blpts)
```
nkern: perhaps you should use `np.asarray` here, as it will not make a copy of the data if it already conforms to the `array_like`

nkern:
So I think this version of `_fast_lookup_blpairts` passes the float rounding error test I was worried about with the previous implementation.
```
def _fast_lookup_blpairts(src_blpts, query_blpts, time_prec=8):
    """
    Helper function to allow fast lookups of array indices for large arrays of 
    blpair-time tuples.
    
    Parameters
    ----------
    src_blpts : array_like
        Array of size (N, 2), containing a list of (blpair, time) couplets.
    
    query_blpts : array_like
        Array of size (M, 2), containing a list of (blpair, time) couplets that 
        you want to find the indices of in source_blpts.
    
    time_prec : int
        Number of decimals to round time array to when performing float comparision. 

    Returns
    -------
    blpts_idxs : array_like
        Array of integers of size (M,), which are indices in the source_blpts 
        array for each item in query_blpts.
    """
    # This function works by using a small hack -- the blpair-times are turned 
    # into complex numbers of the form (blpair + 1.j*time), allowing numpy 
    # array lookup functions to be used
    src_blpts = np.asarray(src_blpts)
    query_blpts = np.asarray(query_blpts)
    src_blpts = src_blpts[:,0] + 1.j*np.around(src_blpts[:,1], time_prec)
    query_blpts = query_blpts[:,0] + 1.j*np.around(query_blpts[:,1], time_prec)
    
    # Do np.where comparison for all new_blpts
    # (indices stored in second array returned by np.where)
    blpts_idxs = np.where(src_blpts == query_blpts[:,np.newaxis])[1]

    return blpts_idxs
```

nkern:
[PEP8] Also, in general, I noticed you default to `[:,:,:]` style whereas I default to `[:, :, :]` style (i.e. spaces after commas). I'd prefer to use spaces after commas when slicing, seeing as it will trigger a `pep8` run to fail. But seeing as we haven't turned on pep8 checks, its ultimately up to you what you wanna do :)

philbull:
@nkern About the float rounding issue: Normally I would agree with you, but in this case I think it should be safe the way it's currently written. `complex128` is just two `float64`s stuck together, so in principle the conversion from tuple -> complex is lossless. As long as the items in `query_blpts` are ultimately taken from the same source as `src_blpts`, there should be no rounding issues -- the `np.where` comparison is, effectively, just looking for items with the same byte patterns, from two arrays that have been subjected only to the same lossless operations as one another.

The previous implementation, which did tuple comparisons, worked in more or less the same way -- there was no need to test for closeness. So, I think it's fine, but your proposed solution should work too if you'd prefer to err on the side of caution.

philbull:
(I suppose what you're saying is that it could fail if the two input arrays have been subjected to any lossy operations, e.g. if the times were not taken from an existing array but were generated by other means?)

nkern:
@philbull Yes I also agree that in the limit of a lossless comparison your implementation works fine. I'm just trying to think if there is a case in which one is comparing two arrays where some lossy operation has been applied, and I think this could easily be the case, so I think it's better to err on the safe side.

philbull:
@nkern OK, sounds good. I've made your suggested change, so this should be ready to go now.

philbull:
Duncan added a change to speed up one more thing, which looks good to me.

PR #157, opened 2018-07-13
Add flagging functions for power spectra

shaunakmodak:
Functions for generating greedy flagging masks, consolidating LST binned NSamples & Flag data into numpy arrays, and plotting visualizations of the flagging.


hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
...
+    flags : numpy.ndarray
+        binary array with 1 representing flagged, 0 representing unflagged
+        
+    n_threshold : int
+        minimum number of samples needed for a point to remain unflagged
+        
+    greedy : bool
+        greedy flagging is used if true (default is False)
+        
+    axis : int
```
philbull: This would be better off as a string, and maybe with a name change, e.g. `first='row'`.

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Generates a greedy flags mask from input flags and nsamples arrays
+    
+    Parameters
+    ----------
+    nsamples : numpy.ndarray
+        integer array with number of samples available for each frequency channel at a given LST angle
+    
+    flags : numpy.ndarray
+        binary array with 1 representing flagged, 0 representing unflagged
```
philbull: Is this the same as the output of `UVData.get_flags()`, or does that need to be modified in some way before passing it to this function?

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+
+mask_generator(nsamples, flags, n_threshold, greedy=False, axis, greedy_threshold, retain_flags=True):
+    """
+    Generates a greedy flags mask from input flags and nsamples arrays
```
philbull: I think it's worth briefly explaining how the algorithm works here.

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
...
+    nsamples : numpy.ndarray
+        integer array with number of samples available for each frequency channel at a given LST angle
+    
+    flags : numpy.ndarray
+        binary array with 1 representing flagged, 0 representing unflagged
+        
+    n_threshold : int
+        minimum number of samples needed for a point to remain unflagged
+        
+    greedy : bool
```
philbull: What if it's False?

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+
+mask_generator(nsamples, flags, n_threshold, greedy=False, axis, greedy_threshold, retain_flags=True):
```
philbull: Maybe a more descriptive function name, e.g. `construct_factorizable_mask`?

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
...
+        greedy flagging is used if true (default is False)
+        
+    axis : int
+        which axis to flag first if greedy=True (1 is row-first, 0 is col-first)
+        
+    greedy_threshold : float
+        if greedy=True, the threshold used to flag rows or columns if axis=1 or 0, respectively
+        
+    retain_flags : bool
+        LST-Bin Flags are left flagged even if thresholds are not met (default is True)
```
philbull: The data going into this are not necessarily LST binned.

hera_pspec/flags.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+    shape = nsamples.shape
+    flags_output = np.zeros(shape)
+
+    num_exactly_equal = 0
+
+    # comparing the number of samples to the threshold 
+    
+    for i in range(shape[0]):
```
philbull: Can this nested `for` loop perhaps be replaced by a couple of strategic calls to `np.where`? We can discuss this.

philbull (changes requested):
Review of in-progress PR.

nkern:
Have either of you seen the development in `PSpecData.broadcast_dset_flags`? Seems like there is some possible overlap in what these two are trying to accomplish. Perhaps some of the code in `broadcast_dset_flags` should be put in `flags.py` and `broadcast_dset_flags` should then call routines in `flags.py`...

philbull:
I've moved most of the code from this PR into `hera_stats`, where it will be very useful for testing different ways of flagging the data. The relevant PR is here: https://github.com/HERA-Team/hera_stats/pull/22

This code could have lived in `hera_pspec` too, but I think it's worth trying to limit the number of peripheral features in pspec and keep it more or less focused on its core function (it's complicated enough). `hera_stats` is supposed to be a bigger tent, covering many different aspects of the analysis, so I think it's appropriate to move this there.

Issue #155, opened 2018-07-09
Add code to make waterfalls of delay spectra


philbull:
Write generic plotting function to plot delay spectra as a function of LST.

philbull:
This is solved by PR #158.

philbull:
PR merged.

PR #158, opened 2018-07-16
Add ability to plot waterfalls of delay spectra

philbull:
Plot delay spectra ordered by time or baseline-pair. Solves #155.

### DIFF
--- a/hera_pspec/plot.py
+++ b/hera_pspec/plot.py
@@ -274,7 +274,11 @@ def delay_waterfall(uvp, blpairs, spw, pol, average_blpairs=False,
     # Format colorbar labels in log mode
     if log:
         lbls = []
-        for tick in cbar.get_ticks():
+        try:
+            ticks = cbar.get_ticks()
+        except:
+            ticks = cbar.get_yticks() # Needed for matplotlib version < 2.2
+        for tick in ticks:
             # Check to make sure we don't accidentally round the exponent
             if ("%1.1f" % tick)[-1] != "0":
                 lbls.append("")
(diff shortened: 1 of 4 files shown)


nkern:
Thanks for getting a start on this, its much-needed! One comment is, I'm not sure if its right to plot a waterfall of different baseline-pairs. It seems a little odd to do that, and perhaps is just confusing. Seems like when we say "waterfall", we should mean a single baseline-pair with a non-zero time axis which makes it a heatmap. If we want to make multiple blpair heatmaps, instead of concatenating them into a single plot, which I think is just a little confusing, we should be making multiple subplots... I can make this edit if you are ok w/ it and push to the branch

philbull:
The idea was that this could eventually be used to make wedge plots too, which is why I allowed it to take multiple blpairs as an argument. It's not quite there yet though (e.g. it doesn't do the necessary averaging for you), so I agree more work is needed.

acliu:
Is there an option to allow all the subplots to be on a common colour scale?

nkern:
@acliu yes there is. you can specify the `vmin` and `vmax` parameters and all the colorscales are set to those limits

nkern:
@philbull the other important change is that I enforced `matplotlib>=2.2` in the `setup.py` b/c I had `matplotlib=2.0` and it failed due to the `cbar.get_yticks` call (even though I see you had an explicit catch for it, it didn't work for me)... as well as `numpy>1.14` b/c that was causing some problems for me locally due to a `numpy==1.13` installation...

PR #165, opened 2018-07-19
Repo cleanup pretag

nkern:
A few minor changes addressing issues in `UVPSpec` and `PSpecData` before we tag a release

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -310,6 +310,7 @@ class PSpecData(object):
         """
         if keys is None:
             self._C, self._I, self._iC, self._Y, self._R = {}, {}, {}, {}, {}
+            self._identity_G, self._identity_H, self._identity_Y = {}, {}, {} 
         else:
             for k in keys:
                 try: del(self._C[k])
@@ -1097,7 +1098,7 @@ class PSpecData(object):
 
         return H / 2.
 
-    def get_unnormed_E(self,key1,key2):
+    def get_unnormed_E(self, key1, key2):
         """
         Calculates a series of unnormalized E matrices, such that
 
@@ -2001,7 +2002,6 @@ class PSpecData(object):
 
             # clear covariance cache
             self.clear_cache()
-            built_GH = False  # haven't built Gv and Hv matrices in this spw loop yet
 
             # setup emtpy data arrays
             spw_data = []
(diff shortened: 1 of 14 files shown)


nkern:
still working on one last edit...

hera_pspec/pspecdata.py
```diff
@@ -2075,15 +2076,43 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
...
-                        # in this case, all Gv are the same, so skip if already built for this spw!
-                        pass
+                    if input_data_weight == 'identity':
+                        # in this case, all Gv and Hv differ only by flagging pattern
+                        # so check if we've already computed this
+                        # First: get flag weighting matrices given key1 & key2
+                        Y = np.vstack([self.Y(key1).diagonal(), self.Y(key2).diagonal()])
+
+                        # Second: check cache for Y
+                        matches = [np.isclose(Y, y).all() for y in self._identity_Y.values()]
```
philbull: This is probably fine, but an alternative would be to hash the array using md5sum.

hera_pspec/uvpspec_utils.py
```diff
@@ -241,6 +242,10 @@ def _blpair_to_antnums(blpair):
     ant2 = int(np.floor(blpair / 1e6 - ant1*1e3))
     ant3 = int(np.floor(blpair / 1e3 - ant1*1e6 - ant2*1e3))
     ant4 = int(np.floor(blpair - ant1*1e9 - ant2*1e6 - ant3*1e3))
+    ant1 -= 100
```
philbull: Why are these on separate lines? They could just be put into the lines above.

hera_pspec/pspecdata.py
```diff
@@ -2075,15 +2076,43 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
...
+                            match = self._identity_Y.keys()[matches.index(True)]
+                            Gv = self._identity_G[match]
+                            Hv = self._identity_H[match]
+                        else:
+                            # This Y doesn't exist, so compute it
+                            if verbose: print("  Building G...")
+                            Gv = self.get_G(key1, key2)
+                            Hv = self.get_H(key1, key2, sampling=sampling)
+                            # cache it
+                            self._identity_Y[(key1, key2)] = Y
```
philbull: Just for clarity: This is done per baseline-pair, and not per blpair-time? So the flags still have to remain time-independent?

hera_pspec/uvpspec_utils.py
nkern: I don't think so, see how ant2 is dependent on ant1 and so on..

hera_pspec/pspecdata.py
nkern: Yeah, done per-baselinepair (not time). flags still need to remain time-independent

hera_pspec/uvpspec_utils.py
philbull: Oops, my bad.

PR #166, opened 2018-07-27
Plotifying Parts of Preprocess Pipe, Various Bug Fixes & Updates from API Changes etc.

nkern:
This is a PR to "plotify" parts of the pipeline, such that it can produce automated diagnostic plots. In addition, this adds the beginning of a web-interface app that can host the plots for shared accessibility to the plot output in the `pipelines` directory.

This also adds a `uvpspec_utils.subtract_uvp` to take two `UVPSpec` objects and subtract their data (and add their stats_array in quadrature), which is useful for null-testing. Also, I found a bug in `UVPSpec.select` slicing when both `blpairts` and `pols` are selected. So I fixed that and added unittests against it.

This also makes various bug fixes (`pspecdata`, `uvpspec`, `pspecbeam`, `pipelines`, `utils`), and updates to code based on updated `hera_cal` and `pyuvdata` dependencies.


nkern:
I think I fixed most of the tests that were borked: now the failures are coming from the fact that I updated code here anticipating a fix to `pyuvdata`'s `get_blt_inds` function, that actually hasn't been merged into `pyuvdata` master yet. Once that happens (ETA ~couple hours or so), we can re-run tests and, I think, this PR should pass... but we'll have to see..

hera_pspec/pspecbeam.py
```diff
@@ -491,17 +493,16 @@ def __init__(self, OmegaP, OmegaPP, beam_freqs, cosmo=None):
             Cosmology object. Uses the default cosmology object if not 
             specified. Default: None.
         """
-        self.allowed_pols = ['pI', 'pQ', 'pU', 'pV', 
-                             'XX', 'YY', 'XY', 'YX']
         self.OmegaP = {}; self.OmegaPP = {}
+        self.allowed_pols = [1, 2, 3, 4, -5, -6, -7, -8]
```
philbull: These magic numbers need to be documented.

hera_pspec/pspecdata.py
```diff
@@ -2616,11 +2633,15 @@ def pspec_run(dsets, filename, dsets_std=None, groupname=None, dset_labels=None,
     history : str
         String to add to history of each UVPSpec object.
 
-    Returns
+    Returns (psc, ds)
```
philbull: This will break readthedocs

hera_pspec/utils.py
```diff
@@ -895,7 +910,11 @@ def get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False, bl_len_range=(0, 1e4),
         List of bad antenna numbers to exclude
 
     add_autos : bool
-        If True, add into autocorrelation to redundant groups
+        If True, add into autocorrelation group to the redundant group list.
+
+    low_hi : bool
```
philbull: Bit of a strange option...

hera_pspec/uvpspec.py
```diff
@@ -1024,6 +1051,9 @@ def read_from_group(self, grp, just_meta=False, spws=None, bls=None,
         times : float ndarray
             Times from the time_avg_array to keep.
 
+        lsts : float ndarray
+            lsts from the lst_avg_array to keep.
```
philbull: Not clear how `lsts` and `times` are different?

hera_pspec/uvpspec_utils.py
```diff
@@ -70,11 +287,14 @@ def _select(uvp, spws=None, bls=None, only_pairs_in_bls=False, blpairs=None, tim
 
     times : float ndarray of times from the time_avg_array to keep
 
+    lsts : float ndarray of lsts from the lst_avg_array to keep. Cannot be specified if times is specifed.
```
philbull: *specified

hera_pspec/pspecbeam.py
nkern: Not sure what you mean. They are, they are the AIPS convention for polarization integers. See `pyuvdata.utils.polstr2num` and `pyuvdata.utils.polnum2str` for how these are made. I can make a comment pointing to them in the code, though

hera_pspec/pspecdata.py
nkern: oh okay, so I'll just leave it blank then I guess

hera_pspec/utils.py
nkern: Yeah.. it turns out to be a convenience option based on how `redcal.get_reds` works...

PR #167, opened 2018-08-09
travis install pyuvdata commit 9c94be0366eb41692f55

nkern:
currently HEAD of `master` (8/9/2018)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ env:
   - PYUVDATA_VERSION="@9c94be0366eb41692f5501caac99981026a7ffe6"
   - PYUVDATA_VERSION=""
 
+allow_failures:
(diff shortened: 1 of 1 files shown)


philbull:
@nkern I've made this a bit fancier by making Travis build against two versions of pyuvdata: (1) a commit that we specify; and (2) the current version of master. So it will tell us if recent updates to pyuvdata break things for us.

nkern:
@philbull awesome! I'm happy w/ that so long travis knows not to fail a PR check if the build against `master` breaks, but that's not what seems what will happen. Do you know how to specify our PR checks?

philbull:
@nkern I added a line to the Travis config that (hopefully) makes it so that issues with pyuvdata won't block PRs.

PR #168, opened 2018-08-09
added delay_wedge() to plot.py

pchich:
This adds delay_wedge to plot.py.


hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
     if new_plot:
         return fig
     
+def delay_wedge(uvps, blpairs, fold=False, cosmo=True, center_line=True, horizon_lines=True, cosmo_cbar=True, suptitle=''):
+    """
+    Plot a 2D delay spectrum (or spectra) for a group of baselines.
+    
+    Parameters
+    ----------
+    uvps : UVPspec
```
nkern: Let's make this a single `UVPSpec` or a list of `UVPSpec`, which should be concatenated inside the function.

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        times, polarizations, and spectral windows.
+        
+    blpairs : list of tuples
+        List of baseline-pair tuples.
+        
+    fold : bool, optional
+        Whether to fold the power spectrum in :math:`|k_\parallel|`. 
+        Default: False.
+        
+    cosmo : bool, optional
```
nkern: this kwarg `cosmo` is a little confusing b/c we also call `cosmo` an object that holds cosmologies. Let's change this to `delay`, which is `True` if plotting tau and `False` if plotting k

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
     if new_plot:
         return fig
     
+def delay_wedge(uvps, blpairs, fold=False, cosmo=True, center_line=True, horizon_lines=True, cosmo_cbar=True, suptitle=''):
```
nkern: This function should optionally take an `ax=None` kwarg, if the user wants to instantiate their own figure and subplots themselves

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        Suptitle for the plot.  If not provided, no suptitle will be plotted.
+        Default: ''.
+        
+    Returns
+    -------
+    f : matplotlib.pyplot.Figure
+        Matplotlib Figure instance.
+    """
+    
+    if (type(uvps) != list) and (type(uvps) != np.ndarray):
```
nkern: This should check if its a list, and if so, concatenate all uvp via `hp.uvpspec_utils.uvpspec_combine`

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        raise AttributeError("The uvps paramater should be a list of UVPSpec objects.")
+        
+    #Initialize axes objects
+    ncols = len(uvps)
+    f, axes = plt.subplots(
```
nkern: only create fig and axes if `ax=None`, see other plotting functions in this script for example

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        sharex=True,
+        sharey=True,
+        squeeze=False,
+        figsize=(20, 10))
+    
+    #Plot each uvp
+    pols = []
+    plt.subplots_adjust(wspace=0, hspace=0.1)
+    for uvp, ax in zip(uvps, axes.flatten()):
+        #Find redundant-length baseline groups and their baseline lengths
```
nkern: use the `reds, lens, angs = .utils.get_reds(uvp, ...)` function for getting redundant baseline groups and their lengths and angles

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+            else:
+                bllens.append(bllen)
+                sorted_blpairs[bllen] = [antnums]
+        bllens = sorted(bllens)
+        
+        #Average the spectra along redundant baseline groups and time
+        uvp.average_spectra(blpair_groups=sorted_blpairs.values(), time_avg=True)
+        
+        #Grab polarization string for naming the plot later on
+        pol_int_to_str = {1: 'pI', 2: 'pQ', 3: 'pU', 4: 'pV', -5: 'XX', -6: 'YY', -7: 'XY', -8: 'YX'}
```
nkern: use `pyuvdata.utils.polnum2str` instead of this

hera_pspec/plot.py
nkern: Additionally, you should request a `spw` and a `pol` as *arguments*. See other plotting functions for example

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        pol = pol_int_to_str[uvp.pol_array[0]]
+        pols.append(pol)
+        
+        #Format data array
+        data = uvp.data_array[0][:, :, 0]
```
nkern: Remove this line in favor of one below

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+        if fold:
+            uvp.fold_spectra()
+            data = uvp.data_array[0][:, data.shape[1] // 2:, 0]
```
nkern: remove this line, in favor of one below

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+
```
nkern: here you should access the data via our `get_data` method, something like
```
data = uvp.get_data((spw, blpair, pol))
```

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+    if cosmo:
+        f.text(0.5, 0.05, r'$k_{\parallel}\ [h\ Mpc^{-1}]$', fontsize=15, ha='center')            
+        f.text(0.07, 0.5, r'$k_{\perp}\ [\rm h\ Mpc^{-1}]$', fontsize=15, va='center', rotation='vertical')
+    else:
+        f.text(0.5, 0.05, r'$\tau\ [ns]$', fontsize=15, ha='center')
+        f.text(0.07, 0.5, r'Baseline Group $[m]$', fontsize=15, va='center', rotation='vertical')
+        
+    #add colorbar label
+    if cosmo_cbar:
+        cbar.set_label(r"$P(k_{\parallel})\ [mK^2h^{-3}Mpc^3]$", fontsize=15, ha='center')
```
nkern: this should use the units found inside the `uvp` object. see other plotting functions for examples

hera_pspec/plot.py
```diff
@@ -494,3 +496,176 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='real', average_blpairs=Fa
...
+    else:
+        cbar.set_label(r'$P(\tau)\ [mK^2]$', fontsize=15, ha='center')
```
nkern: same here

nkern:
Thanks for all the hard work on this Paul and Penn group! I have some major feedback regarding the design of the function that breaks down into three main categories: 1. using more pre-built `hera_pspec` routines to do basic things, 2. bugs and 3. slimming the function down to operating on only a single `UVPSpec` object.

1. The `UVPSpec` object contains a lot of functions that makes life easier: getting k-para and k-perps, delays, data arrays etc. I added these routines in where necessary to be self-consistent w/ the rest of the module.

2. There were some bugs in the function that probably weren't caught b/c there wasn't any unit-tests pushed with the function. I've fixed those bugs and also added unit-tests.

3. I like the ambition behind trying to plot a list of `UVPSpec`, but it ended up making the function a little hard to read and understand, and a little cumbersome to expand upon. If we limit the function to operating on a single subplot, we can allow the user to make their own Figure and arrange the subplots how they want, and then make their own wrapper function for personalized multi-subplot figures--thus making it more flexible in my mind.

I went ahead and made these changes so we can start making wedge plots sooner rather than later! @pchich if you have any comments on what fell out in the wash, please let me know. To get back your multi-`UVPSpec` plotting capabilities, something like this should suffice
```
fig, axes = plt.subplots(1, 4, figsize=(15, 6))
for ax, uvp in zip(axes, uvp_list):
    delay_wedge(uvp, spw, pol, ax=ax)
```
In the meantime if @philbull could have a quick look at it and give us your feedback / approve when ready, that'd be great. Thanks!

nkern:
Here are some examples of the plotting script as it currently stands

dly_wdge1

dly_wdge2

PR #179, opened 2018-11-09
Lazy covariance

anguta:
Added some new weighting modes.


hera_pspec/pspecdata.py
```diff
@@ -13,6 +13,8 @@
 import ast
 import glob
 
+SPEED_OF_LIGHT = 299792458.
```
nkern: we have a `conversions.utils` object specifically for these values

anguta: I've removed explicit references to speed of light. New implementation allows for arbitrary "clean" windows to be specified in which foregrounds are down-weighted.

hera_pspec/utils.py
```diff
@@ -12,6 +12,45 @@
 from datetime import datetime
 
 
+
+def clean_inv_mat(nchan, df, filter_centers, filter_widths, filter_factors):
+    """
+    Computes inverse of clean weights for a baseline.
+    This form of weighting is diagonal in delay-space and down-weights tophat regions
+    Parameters
```
nkern: insert line breaks (PEP8)

hera_pspec/utils.py
```diff
@@ -12,6 +12,45 @@
 from datetime import datetime
 
 
+
+def clean_inv_mat(nchan, df, filter_centers, filter_widths, filter_factors):
```
nkern: lets think of a more apt name for the matrix...

anguta: I'm going with sync_downweight_mat

hera_pspec/utils.py
anguta: I now have two blank lines between the functions I've added. Is this the right number?

nkern: Yes, but I actually meant for the doc string. Look at other docstrings to see how this is done, but you should keep a space between the function explanation and Parameters section. Ex:
"""
Computes inverse ...
This form of ...

Parameters
-----------
"""
and the same for Returns. I know its nitpicky but I'm just trying to keep everything consistent :)  If you just look at all the other doc strings in this file you'll see the pattern

hera_pspec/pspecdata.py
nkern: okay but, then can we get rid of this line then?

hera_pspec/pspecdata.py
```diff
@@ -774,6 +779,25 @@ def R(self, key):
             elif self.data_weighting == 'iC':
                 self._R[Rkey] = sqrtT.T * sqrtY.T * self.iC(key) * sqrtY * sqrtT
 
+            elif self.data_weighting == 'sinc_downweight':
+                r_param_key = (self.data_weighting,) + key
+                if not r_param_key in self.r_params:
+                    raise_warning("Warnging: no filter params specified for "
+                                    "clean weights! Defaulting to Identity!")
```
nkern: warnging typo, also lets say "sinc weights" here

hera_pspec/pspecdata.py
```diff
@@ -335,6 +338,7 @@ def clear_cache(self, keys=None):
         if keys is None:
             self._C, self._I, self._iC, self._Y, self._R = {}, {}, {}, {}, {}
             self._identity_G, self._identity_H, self._identity_Y = {}, {}, {}
+            self._RegFactor = {}
```
nkern: I don't see where `RegFactor` is used elsewhere in the code. Let's get rid of this if its not being used

hera_pspec/pspecdata.py
```diff
@@ -343,6 +347,8 @@ def clear_cache(self, keys=None):
                 except(KeyError): pass
                 try: del(self._iC[k])
                 except(KeyError): pass
+                try: del(self._RegFactor[k])
+                except(KeyError): pass
```
nkern: you probably also want to clear `self.r_params` here too

hera_pspec/pspecdata.py
```diff
@@ -2230,15 +2284,19 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
...
 
                     # Generate the covariance matrix if error bars provided
                     if store_cov:
                         if verbose: print(" Building q_hat covariance...")
-                        cov_qv = self.cov_q_hat(key1, key2)
+                        if cov_method == 'empirical':
+                            cov_qv = self.get_unnormed_V(key1, key2)
+                            cov_qv = np.array([cov_qv for m in range(self.Ntimes)])
+                        elif cov_method == 'propagated':
+                            cov_qv = self.cov_q_hat(key1, key2)
```
nkern: i don't understand why this stuff is new.. was this from an old PR that somehow found its way into here?

hera_pspec/utils.py
```diff
@@ -12,6 +12,50 @@
...
+    """
+    Computes inverse of clean weights for a baseline.
+    This form of weighting is diagonal in delay-space and down-weights tophat regions
+    Parameters
+    ----------
+    nchan: integer
+        Number of channels on baseline
+    df: float
+        channel width (Hz)
+    weights: array-like (complex or float)
```
nkern: let's specify `shape=(Nfreqs,)`

nkern:
It seems like what you've done in `sinc_downweight` is to apply a fourier filter, which is done via a discrete convolution whose functional form is dictated by a tophat in the fourier domain, or a sinc function in real space.. I'm just curious as to whether this is something we want to replace the `K` matrix in our `R` weighting, or have as a separate step in our QE chain. perhaps we can discuss this in person or on the next pspec telecon

nkern:
@anguta Here is a neat example for you to test out.
```
import numpy as np
from hera_pspec import utils
from scipy import stats

# generate a bunch of white noise vectors
x = np.array([stats.norm.rvs(0, 1, 32) for i in range(100)]).T

# generate a sinc downweight matrix
weights = np.ones(32)
cmat1 = utils.sinc_downweight_mat_inv(32, 100e3, weights, filter_centers = 0., filter_widths = 1500e-9, filter_factors = 1e-9)

# apply to data
xfilt = cmat1.dot(x)

# get covariances of data, filtered data, and then dot C with sinc matrix
C = np.cov(x)
CF = np.cov(xfilt)
CF2 = cmat1.dot(C).dot(cmat1.T)
```
You'll find that CF and CF2 are nearly identical, which I think suggests that this is not necessarily a replacement for covariance weighting..

hera_pspec/utils.py
anguta: I've added the white space.

hera_pspec/pspecdata.py
anguta: I've removed the typo and changed "clean" to "since" :).

hera_pspec/pspecdata.py
anguta: Now clears r_params.

hera_pspec/pspecdata.py
anguta: I've removed all mentions of RegFactor.

hera_pspec/utils.py
anguta: I've added a comment that the shape should be (Nfreqs,)

hera_pspec/pspecdata.py
anguta: I actually tweaked this because 

> It seems like what you've done in `sinc_downweight` is to apply a fourier filter, which is done via a discrete convolution whose functional form is dictated by a tophat in the fourier domain, or a sinc function in real space..

Minor comment: Its not technically a convolution unless we define it as a diagonal in delay space and derive it through a DFT. As implemented, the inverse of the sinc matrix is not quite translation invariant (not purely a convolution). 

> I'm just curious as to whether this is something we want to replace the `K` matrix in our `R` weighting, or have as a separate step in our QE chain. In other words, this matrix isn't really a covariance matrix, its a toeplitz matrix that is used for a discrete convolution, so not sure if its fair to replace the slot in QE that would normally be used for covariance matrices with this matrix... perhaps we can discuss this in person or on the next pspec telecon.

I think that the main reason to include this matrix in our power-spectrum stage is that its action is most easily incorporated into the calculation of power-spectrum covariances and error bars if its added as the K matrix. That said, we could propogate it into C either empirically or through direct calculation (though for the calculation, we'd need to develop some extra machinery outside of pspec).

We should discuss this on the next telecon.

anguta: OK. This might be a little tangential to the PR. I ended up adding it because there are now two independent different ways of calculating power spectrum covariances but no way of specifying which one to use. I will get rid of these lines for now and create an issue because it really should be added as a different PR!

anguta:
> It seems like what you've done in `sinc_downweight` is to apply a fourier filter, which is done via a discrete convolution whose functional form is dictated by a tophat in the fourier domain, or a sinc function in real space.. I'm just curious as to whether this is something we want to replace the `K` matrix in our `R` weighting, or have as a separate step in our QE chain. In other words, this matrix isn't really a covariance matrix, its a toeplitz matrix that is used for a discrete convolution, so not sure if its fair to replace the slot in QE that would normally be used for covariance matrices with this matrix... perhaps we can discuss this in person or on the next pspec telecon.


I think that the main reason to include this matrix in our power-spectrum stage is that its action is most easily incorporated into the calculation of power-spectrum covariances and error bars if its added as the K matrix. That said, we could propogate it into C either empirically or through direct calculation (though for the calculation, we'd need to develop some extra machinery outside of pspec).

We should discuss this on the next telecon.

anguta:
> Can you also update the doc-string of `PSpecData.R()` to note that the `K` matrix can be either `I`, `iC` or `sinc_downweight` based on the value of `self.data_weighting`?

I've added that sinc_downweight can be used as an option

anguta:
> @anguta Here is a neat example for you to test out.
> 
> ```
> import numpy as np
> from hera_pspec import utils
> from scipy import stats
> 
> # generate a bunch of white noise vectors
> x = np.array([stats.norm.rvs(0, 1, 32) for i in range(100)]).T
> 
> # generate a sinc downweight matrix
> weights = np.ones(32)
> cmat1 = utils.sinc_downweight_mat_inv(32, 100e3, weights, filter_centers = 0., filter_widths = 1500e-9, filter_factors = 1e-9)
> 
> # apply to data
> xfilt = cmat1.dot(x)
> 
> # get covariances of data, filtered data, and then dot C with sinc matrix
> C = np.cov(x)
> CF = np.cov(xfilt)
> CF2 = cmat1.dot(C).dot(cmat1.T)
> ```
> 
> You'll find that CF and CF2 are nearly identical, which I think suggests that this is not necessarily a replacement for covariance weighting..

I don't expect you'd find significant differences when inverse covariance filtering of white noise since all of the modes are equally represented (unless you subtract a large fraction of the modes). I think a better example would be to add a tone in the filtering window. I'll look into this more.

hera_pspec/pspecdata.py
nkern: seems like this line is still here. let's get rid of this if we arent' using it

hera_pspec/pspecdata.py
```diff
@@ -774,6 +779,25 @@ def R(self, key):
             elif self.data_weighting == 'iC':
                 self._R[Rkey] = sqrtT.T * sqrtY.T * self.iC(key) * sqrtY * sqrtT
 
+            elif self.data_weighting == 'sinc_downweight':
+                r_param_key = (self.data_weighting,) + key
+                if not r_param_key in self.r_params:
+                    raise_warning("Warning: no filter params specified for "
+                                    "sinc weights! Defaulting to Identity!")
```
nkern: this is a situation where in the `history` of the output `uvp` object, the data weighting will be `'sinc_downweight'` but the power spectra will actually have `'identity'` weighting. I think we may be better off having this raise an Error rather than a warning. Perhaps @philbull has some thoughts on this.

Also, in theory we would want all of `self.r_params` to be propagated to the file history (b/c each baseline can have a different `r_param`) but the number of baselines is such that this will likely become too large. We should think of how we can pass through a minimum amount of information on these parameters through to the file history...

nkern:
@anguta could you also link your lazy covariance PDF or overleaf url for posterity here? thanks

hera_pspec/pspecdata.py
philbull: As a minimum, you could include a hash of `self.r_params` in the history, as this would at least allow you to check whether the same settings are being used or not. It doesn't help you figure out what's different though...
And yes, it would be safer to raise an error instead of using Identity as default. If the user has mis-specified something, and it's not obvious what they meant to do, then we should stop them from proceeding :)

hera_pspec/pspecdata.py
```diff
@@ -783,10 +807,36 @@ def set_weighting(self, data_weighting):
...
+        r_params: dictionary with parameters for weighting matrix.
+                  Proper fields
+                  and formats depend on the mode of data_weighting.
+                data_weighting == 'sinc_downweight':
+                                dictionary with fields
+                                'filter_centers', list of floats (or float) specifying the centers of clean windows
+                                                  in units of 1/(channel index)
+                                'filter_widths', list of floats (or float) specifying the width of clean windows
+                                                  in units of 1/(channel index)
+                                'filter_factors', list of floats (or float) specifying how much power within each clean window
```
philbull: How is this number propagated into an amount of suppression? Is it a fraction of something?

hera_pspec/pspecdata.py
```diff
@@ -1898,16 +1948,20 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
             Each tuple should contain a start (inclusive) and stop (exclusive)
             channel used to index the `freq_array` of each dataset. The default
             (None) is to use the entire band provided in each dataset.
-        
+
+        cov_method: string, specify whether covariance matrices should be computed
+        empirically ('empircal') from the data or through propagation of error bars provided by
```
philbull: Typo 'empircal'. Also, please put the docstring on a new line, so it doesn't break the ReadTheDocs output, e.g.
```py
cov_method : str
    Specify whether...
```

hera_pspec/pspecdata.py
```diff
@@ -783,10 +807,36 @@ def set_weighting(self, data_weighting):
...
+             pol is an index of polarization of r_matrix.
+
+        r_params: dictionary with parameters for weighting matrix.
+                  Proper fields
+                  and formats depend on the mode of data_weighting.
+                data_weighting == 'sinc_downweight':
+                                dictionary with fields
+                                'filter_centers', list of floats (or float) specifying the centers of clean windows
+                                                  in units of 1/(channel index)
+                                'filter_widths', list of floats (or float) specifying the width of clean windows
```
philbull: Isn't `channel_index` a dimensionless integer? So this just means the width is specified in an integer number of channels?

hera_pspec/pspecdata.py
```diff
@@ -55,16 +57,17 @@ def __init__(self, dsets=[], wgts=None, dsets_std=None, labels=None,
         self.spw_range = None
         self.spw_Nfreqs = None
         self.spw_Ndlys = None
-
+        self.r_params = {}
```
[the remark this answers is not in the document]
anguta: r_params is an optional dictionary that stores parameters for any parameteric R matrices that are used. What these parameters are depends on the specific R matrix that is used. In the case of lazy covariance weights, the R matrix is parameterized by a list of filtering window centers, filtering window widths, and the level of filtering to be applied. I have added more documentation.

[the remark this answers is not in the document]
anguta: I've added documentation to the line. It would be nice if there were a more appropriate place -- general documentation describing the PspecData class and its parameters. I've added a separate issue to address this.

hera_pspec/pspecdata.py
```diff
@@ -505,7 +510,7 @@ def set_C(self, cov):
         self.clear_cache(cov.keys())
         for key in cov: self._C[key] = cov[key]
 
-    def C_model(self, key, model='empirical'):
+    def C_model(self, key, model='empirical', model_params = None):
```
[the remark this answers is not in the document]
anguta: I think this was orphan code. I've reverted to the original version.

hera_pspec/pspecdata.py
```diff
@@ -774,6 +779,25 @@ def R(self, key):
             elif self.data_weighting == 'iC':
                 self._R[Rkey] = sqrtT.T * sqrtY.T * self.iC(key) * sqrtY * sqrtT
 
+            elif self.data_weighting == 'sinc_downweight':
+                r_param_key = (self.data_weighting,) + key
+                if not r_param_key in self.r_params:
+                    raise_warning("Warning: no filter params specified for "
+                                    "sinc weights! Defaulting to Identity!")
+                    r_params = {'filter_centers':[],
```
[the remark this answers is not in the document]
anguta: I agree here but the data volume of this dictionary is dwarfed by the size of the data set itself along with the covariances [accessed by C() ] which are addressed per baseline. I think that there should be a general way of tracking redundant baseline groups in the pspecdata object that can be applied to covariance matrices along with r prameters and I think it should have its own pull request.

hera_pspec/pspecdata.py
```diff
@@ -783,10 +807,36 @@ def set_weighting(self, data_weighting):
...
+        key: tuple (dset, bl, [pol]), where dset is the index of the dataset
+             bl is a 2-tuple
+             pol is an index of polarization of r_matrix.
+
+        r_params: dictionary with parameters for weighting matrix.
+                  Proper fields
+                  and formats depend on the mode of data_weighting.
+                data_weighting == 'sinc_downweight':
+                                dictionary with fields
+                                'filter_centers', list of floats (or float) specifying the centers of clean windows
```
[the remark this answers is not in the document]
anguta: The effective impact of this weighting is very close to what is achieved by CLEAN. However, it isn't the same so I've removed this language.

hera_pspec/pspecdata.py
anguta: It's essentially the factor by which modes within the filtering window are multiplied. filter_factor = [1e-9] will multiply modes within the filtering window by a factor of roughly 1e-9.

hera_pspec/pspecdata.py
anguta: This is orphan code for something I've decided to move to a separate pull request. I've removed it.

hera_pspec/pspecdata.py
anguta: It doesn't have to be an integer. I've made the language a little clearer and explicitly stated that fractional channel numbers can be used.

hera_pspec/pspecdata.py
```diff
@@ -783,10 +807,36 @@ def set_weighting(self, data_weighting):
...
 
+    def set_r_param(self, key, r_params):
+        """
+        Set the weighting parameters for baseline at (dset,bl, pol)
+
+        Parameters
+        ----------
+        key: tuple (dset, bl, [pol]), where dset is the index of the dataset
+             bl is a 2-tuple
+             pol is an index of polarization of r_matrix.
```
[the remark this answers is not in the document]
anguta: pol is supposed to be the usual. I've fixed the confusing wording and run the key through parse_blkey.

hera_pspec/pspecdata.py
nkern: Yes that is the point, but I think that instead of a warning and defaulting to `identity` we should raise an error: otherwise the output `uvp` objects will have `data_weighting='sinc_downweight'` in their metadata but will really have been weighted with `identity`. 

And also @anguta one way to propagate the `r_params` is to just propagate each unique `r_param` combination with its baseline-pair key: in a sense only propagating `r_params` for each unique redundant group, which would cut down on the size of the string you would be appending to the history. something to consider.

anguta: I know throw an error when no `r_params` are specified.

anguta:
> @anguta could you also link your lazy covariance PDF or overleaf url for posterity here? thanks

I'm working on a tutorial notebook that also describes some of the underlying math and why the filter is effective.

nkern:
@anguta ok sounds good. do we want to wait on this PR so that you just call `uvtools` capabilities instead of duplicate code here? I think that's probably the best approach.

Also we still need to figure out how to propagate some amount of `r_params` metadata to the file history without overloading it, perhaps making it optional.

anguta:
> @anguta ok sounds good. do we want to wait on this PR so that you just call `uvtools` capabilities instead of duplicate code here? I think that's probably the best approach.
> 
> Also we still need to figure out how to propagate some amount of `r_params` metadata to the file history without overloading it, perhaps making it optional.

Yeah, lets wait until the uvtools code is merged in so I can reference it from hera_pspec. 

cheers,

-Aaron

anguta:
I've added linear cleaning to uvtools and removed duplicate code here.

hera_pspec/pspecdata.py
```diff
@@ -858,7 +910,7 @@ def cov_q_hat(self, key1, key2, time_indices=None):
 
         Returns
         -------
-        cov_q_hat: matrix with covariances between un-normalized band powers
+        cov_q_hat: matrix with covariances between un-normalized band powers (Ntimes, Nfreqs, Nfreqs)
```
philbull: This breaks readthedocs -- description should be indented, on the second line, e.g.
```
cov_q_hat: array_like
    Matrix with covariances between un-normalized band powers (Ntimes, Nfreqs, Nfreqs)
```

hera_pspec/tests/test_pspecdata.py
```diff
@@ -719,8 +730,13 @@ def test_q_hat(self):
 
         self.ds.spw_Ndlys = Nfreq
         # Check that the slow method is the same as the FFT method
-        for input_data_weight in ['identity', 'iC']:
+        for input_data_weight in ['identity', 'iC','sinc_downweight']:
             self.ds.set_weighting(input_data_weight)
+            #nt.assert_raises(ValueError,self.ds.R, key1)
+            #if input_data_weight == 'sinc_downweight':
```
philbull: Commented-out code

philbull:
@anguta Also: the new notebook `examples/PS_estimation_example_inverse_weights.ipynb` seems to be a duplicate of the existing `examples/PS_estimation_example.ipynb`, but with a couple of small changes. I don't think it should duplicate the whole thing, as this will make it hard for users to see what the differences are! So, perhaps it should either focus on showing a specific example of how to use the new weighting scheme, or fully replace the old PS estimation example. (I have a preference for the former; it would be nice to show a simple comparison, e.g. using Nick's example code!)

anguta:
I have no idea where the notebook comes from or how it got added to the PR. I've removed it.

anguta:
> @anguta Also: the new notebook `examples/PS_estimation_example_inverse_weights.ipynb` seems to be a duplicate of the existing `examples/PS_estimation_example.ipynb`, but with a couple of small changes. I don't think it should duplicate the whole thing, as this will make it hard for users to see what the differences are! So, perhaps it should either focus on showing a specific example of how to use the new weighting scheme, or fully replace the old PS estimation example. (I have a preference for the former; it would be nice to show a simple comparison, e.g. using Nick's example code!)

I have no idea where the notebook comes from or how it got added to the PR. I've removed it.

anguta:
I've added the notebook `examples/Inverse_sinc_weighting_example.ipynb` to demonstrate how to use inverse sinc weighting.

anguta:
I've also added a new r_params parameter for uvpspec objects that stores r_parameters as a string (in a compressed format that avoids storing many copies of the same dictionary) along with a decompression method to retrieve the original dictionary.

hera_pspec/pspecdata.py
```diff
@@ -775,6 +781,25 @@ def R(self, key):
             elif self.data_weighting == 'iC':
                 self._R[Rkey] = sqrtT.T * sqrtY.T * self.iC(key) * sqrtY * sqrtT
 
+            elif self.data_weighting == 'sinc_downweight':
+                r_param_key = (self.data_weighting,) + key
+                if not r_param_key in self.r_params:
+                    raise ValueError("Error: no filter params specified for "
+                                    "sinc weights! ")#Defaulting to Identity!")
+                    #r_params = {'filter_centers':[],
```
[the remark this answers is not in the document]
anguta: deleted

hera_pspec/pspecdata.py
```diff
@@ -775,6 +781,25 @@ def R(self, key):
...
+                    #            'filter_widths':[],
+                    #            'filter_factors':[]}
+                else:
+                    r_params = self.r_params[r_param_key]
+                #weights = copy.copy(self.w(key))
+                #weights = np.mean(weights,axis=1) #weights are calculated by averaging weights matrix in time.
+                self._R[Rkey] = sqrtT.T * np.linalg.pinv(sqrtY.T * \
```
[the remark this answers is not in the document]
anguta: I've added the comment.

hera_pspec/pspecdata.py
anguta: fixed.

hera_pspec/tests/test_pspecdata.py
anguta: removed.

hera_pspec/uvpspec.py
```diff
@@ -32,7 +33,7 @@ def __init__(self):
         self._Nfreqs = PSpecParam("Nfreqs", description="Number of unique frequency bins in the data.", expected_type=int)
         self._Npols = PSpecParam("Npols", description="Number of polarizations in the data.", expected_type=int)
         self._history = PSpecParam("history", description='The file history.', expected_type=str)
-
+        self._r_params = PSpecParam("r_params", description = 'r_params.', expected_type = str)
```
philbull: Important to have a clearer description here I think!

hera_pspec/uvpspec.py
```diff
@@ -301,6 +302,23 @@ def get_integrations(self, key, omit_flags=False):
 
         return self.integration_array[spw][blpairts, polpair]
 
+    def get_r_params(self):
+        """
+        decompress r_params dictionary (so it can be readily used).
```
philbull: Docstring could be a little more helpful here -- what are r_params, how is it stored etc.

hera_pspec/tests/test_uvpspec.py
```diff
@@ -304,6 +304,43 @@ def test_clear(self):
...
+        for bl in baselines:
+            key1 = (0,) + bl + ('xx',)
+            key2 = (1,) + bl + ('xx',)
+            r_params[key1] = rp
+            r_params[key2] = rp
+
+        r_params_unique = {}
+        r_params_unique_bls = {}
+        r_params_index = -1
+        #build a compressed string to store r_params
```
philbull: It would be more robust to create a new `UVPSpec` object from data that uses a set of `r_params`, and then check that those params have been stored successfully in the file. This tests seems to just add them manually to the `UVPSpec` object, which isn't the usual workflow.

hera_pspec/tests/test_uvpspec.py
```diff
@@ -467,20 +504,20 @@ def test_combine_uvpspec(self):
         # setup uvp build
```
philbull: How are `r_params` dealt with by `combine_uvpspec`, e.g. when two UVPSpec objects are added together?

philbull: (or `select` for that matter)

anguta: combine_uvpspec now combines` r_params` but refuses to combine two uvpspec objects with a conflict in `r_params` (dictionaries are not equal for same baseline in two different uvpspec objects). 

`select` trims `r_params` to only contain entries that refer to baselines in the down selected boject.

hera_pspec/tests/test_uvpspec.py
anguta: I have adjusted the testing methods that create `UVpspec` objects from data to include a realistic set of `r_params`.

hera_pspec/uvpspec.py
anguta: I've lengthened this docstring to make it more helpful.

hera_pspec/uvpspec.py
anguta: I've improved this description.

anguta:
> @anguta Here is a neat example for you to test out.
> 
> ```
> import numpy as np
> from hera_pspec import utils
> from scipy import stats
> 
> # generate a bunch of white noise vectors
> x = np.array([stats.norm.rvs(0, 1, 32) for i in range(100)]).T
> 
> # generate a sinc downweight matrix
> weights = np.ones(32)
> cmat1 = utils.sinc_downweight_mat_inv(32, 100e3, weights, filter_centers = 0., filter_widths = 1500e-9, filter_factors = 1e-9)
> 
> # apply to data
> xfilt = cmat1.dot(x)
> 
> # get covariances of data, filtered data, and then dot C with sinc matrix
> C = np.cov(x)
> CF = np.cov(xfilt)
> CF2 = cmat1.dot(C).dot(cmat1.T)
> ```
> 
> You'll find that CF and CF2 are nearly identical, which I think suggests that this is not necessarily a replacement for covariance weighting..

I'm a little bit confused by this example. The filtering matrix is supposed to be the inverse of sinc_downweight_inv. At any rate, I'd naively  expect CF2 and CF to be the same since the first is the covariance of the data after applying a matrix multiplication (basically a change of basis with a matrix)  and the second is applying the transformation matrix to the covariance of the original basis (equivalent to calculating the covariance in the new basis).

anguta:
> Looks good, but a few loose ends to tie up:
> 
>     * a couple of docstrings could be a little clearer/more detailed
> 
>     * it's not clear how `r_params` is treated when UVPSpec objects are added together
> 
>     * I think the `get_r_params` test should be a bit more real-worldy
I've addressed all of the issues mentioned here. 

cheers,

-Aaron

hera_pspec/uvpspec_utils.py
```diff
@@ -100,12 +101,92 @@ def subtract_uvp(uvp1, uvp2, run_check=True, verbose=False):
...
+    Decompress json format r_params string into an r_params dictionary.
+
+    Parameters
+    ----------
+    r_params_str: String
+        string with compressed r_params in json format
+
+    Returns
+    -------
+    dictionary with parameters for weighting matrix. Proper fields
```
philbull: Should follow proper format for parameter docstrings. Otherwise it breaks readthedocs!
e.g.
```
Returns
-----------
r_params : dict
    Dictionary with...
```

hera_pspec/uvpspec.py
```diff
@@ -2038,6 +2048,28 @@ def combine_uvpspec(uvps, verbose=True):
     # Construct dict of label indices, to be used for re-ordering later
     u_lbls = {lbl: ll for ll, lbl in enumerate(u.labels)}
 
+    # Concatenate r_params arrays
+    #check that r_params are either all empty or all full.
+    _r_param_strs = [_uvp.r_params for _uvp in uvps]
+    if '' in _r_param_strs:
+        if not np.all(np.asarray([rp == '' for rp in _r_param_strs])):
+            raise ValueError("All r_params must be set or empty."
+                             "Combinging empty with non-empty r_params"
```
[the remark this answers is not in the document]
anguta: This has been fixed :)

hera_pspec/uvpspec_utils.py
anguta: This should be fixed now :).

PR #180, opened 2018-11-11
Adding the output for covariance matrix and  analytic variance

JianrongTan:
Adding function analytic_variance() in pspecdata.py ;
Modifying function C_model() and cross_covar_model() in pspecdata.py;
Adding outputs for cov_array_q, var_array, var_array_q in pspecdata.py;
Adding get_cov_q, get_var, get_var_q in uvpspec.py.


nkern:
@JianrongTan thanks for getting this up. It seems like something when seriously wrong when you merged in the master branch recently, b/c it looks like you erased and then re-comitted _all_ of `uvpspec.py`. This is a little dangerous b/c we can't see the true git diff: i.e. you may have changed something by accident and we can't actually tell b/c the git diff is the entire file. We will probably need to take your changes and re-branch off of master. I can help you with this if you'd like, just let me know...

PR #181, opened 2018-11-21
Adding the output for analytic variance and covariance matrix

JianrongTan:
1. The analytic variance is calculated in pspecdata.analytic_variance();
2. The output functions named in "get ..." for both the analytic variance and covariance matrix are added in uvpspec.py.


hera_pspec/grouping.py
```diff
@@ -240,35 +240,75 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
 
     # Create new data arrays
-    data_array, wgts_array = odict(), odict()
+    data_array, data_array_q, wgts_array = odict(), odict(), odict()
     ints_array, nsmp_array = odict(), odict()
     stats_array = odict([[stat, odict()] for stat in stat_l])
 
     # will average covariance array if present
-    store_cov = hasattr(uvp, "cov_array")
+    ##store_cov = hasattr(uvp, "cov_array")
```
nkern: let's get rid of these commments

hera_pspec/grouping.py
```diff
@@ -310,43 +362,227 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                    bpg_cov_imag = np.array(bpg_cov_imag)
+                    bpg_cov_q_real = np.array(bpg_cov_q_real)
+                    bpg_cov_q_imag = np.array(bpg_cov_q_imag)
+                    bpg_var_real = np.array(bpg_var_real)
+                    bpg_var_imag = np.array(bpg_var_imag)
+                    bpg_var_q_real = np.array(bpg_var_q_real)
+                    bpg_var_q_imag = np.array(bpg_var_q_imag)
+                    w_list = np.array(w_list)
+
+                    times = len(bpg_cov_real[0])
```
nkern: The next hundred lines or so is an example where you can make an in-line function that does a general average, and then just feed it `bpg_cov_real`, or `bpg_cov_imag` etc. to do the operation on each array you want..

hera_pspec/grouping.py
```diff
@@ -494,20 +769,91 @@ def fold_spectra(uvp):
...
+            #    leftright = uvp.cov_array[spw][:, 1:Ndlys//2, Ndlys//2+1:, :][:, ::-1, :, :]
+            #    rightleft = uvp.cov_array[spw][:, Ndlys//2+1: , 1:Ndlys//2, :][:, :, ::-1, :]
+            #    rightright = uvp.cov_array[spw][:, Ndlys//2+1:, Ndlys//2+1:, :]
+            #    uvp.cov_array[spw][:, Ndlys//2+1:, Ndlys//2+1:, :] = .25*(leftleft\
+            #                                                             +leftright\
+            #                                                             +rightleft\
+            #                                                             +rightright)
+            #    uvp.cov_array[spw][:, :Ndlys/2, :, :] = 0.0
+            #    uvp.cov_array[spw][:, :, :Ndlys/2, : :] = 0.0
+            if hasattr(uvp,'cov_array_real'):
```
nkern: another example where an in-line function would help.

hera_pspec/plot.py
```diff
@@ -1009,3 +1010,1272 @@ def _round_sigfig(x, up=True):
...
+        if fit_curve == 'Exponential':
+            y = stats.laplace.pdf(bins, np.mean(data), np.std(data)/np.sqrt(2.))
+            ax.plot(bins, y, lw=2, ls='--', label='Exponential', color='blue')
+    
+    ax.legend(fontsize=12, loc='lower center',framealpha=0)
+    ax.set_yscale('log')
+    ax.grid()
+
+
+def plot_error(uvp, uvp_td, key, time_index, Tsys, normalized=True, extra_error_types=[], **kwargs):
```
nkern: This is slightly redundant with `plot.delay_spectrum`, which can plot errorbars by using the `plot_stats` kwarg. Can you see if that function satisfies what you want to do here, just to minimize redundant code? If not we can keep this function and get rid of `delay_spectrum(plot_stats)` kwarg.

hera_pspec/plot.py
```diff
@@ -1009,3 +1010,1272 @@ def _round_sigfig(x, up=True):
...
+        ax3.plot(dlys, analytic_imag_min, ls='-.', label='analytic-min', c='m')
+    if normalized==True:
+        ax3.plot(dlys, noise[time_index], c='k', lw=0.5, label='thermal noise')
+    ax3.legend(loc='best', framealpha=0)
+    ax3.semilogy()
+    ax3.yaxis.tick_right()
+    ax3.set_xticklabels('')
+
+
+def plot_zscore_hist(uvp, uvp_td, key, wedge, normalized=True, inside_wedge=True, extra_error_types=[], fit_curve='Gaussian', plot_mode='normal', show_robust=True, **kwargs):
```
nkern: We have code in `hera_stats.stats.uvp_zscore` to compute z-scores. I think probably this plotting code should go in `hera_stats`, and should use `hera_stats.stats.uvp_zscore` to compute the zscores. Let me know if you have questions on that code.

hera_pspec/utils.py
```diff
@@ -1002,3 +1004,72 @@ def get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False, bl_len_range=(0, 1e4),
         reds, lens, angs = _reds, _lens, _angs
 
     return reds, lens, angs
+
+def pspecdata_time_difference(ds, time_diff):
+    """
+    Given a PSpecData object and a time difference,
```
nkern: seems like you didn't finish your sentence here

hera_pspec/utils.py
```diff
@@ -1002,3 +1004,72 @@ def get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False, bl_len_range=(0, 1e4),
...
+    data21 = uvd21.data_array
+    data20 -= data21
+    uvd20.data_array = data20
+
+    ds_td = hp.PSpecData(dsets=[uvd10, uvd20], wgts=ds.wgts, beam=ds.primary_beam)
+    return ds_td
+
+def uvd_time_difference(uvd, time_diff):
+    """
+    Given a UVData object and a time difference,
```
nkern: same here, unfinished sentence

hera_pspec/uvpspec.py
```diff
@@ -179,6 +198,369 @@ def get_cov(self, key, omit_flags=False):
...
+                    result_list = np.array(result_list)
+                    return result_list
+                else:
+                    return self.cov_array_q_imag[spw][blpairts, pol]
+            else:
+                raise AttributeError("No covariance array has been calculated.")
+        else:
+            raise ValueError("No types besides real and imag.")
+
+    def get_cov(self, key, type="real", omit_flags=False):
```
nkern: let's avoid using special variables like `type`. perhaps `dtype` is a better kwarg name

hera_pspec/grouping.py
JianrongTan: I have tried on that. Will be more succinct in future commits. Thanks.

hera_pspec/grouping.py
JianrongTan: ok.

hera_pspec/plot.py
JianrongTan: plot_error() tries to plot different kinds of error bars without delay spectra at the same time, and asks for a outer time-difference power spectrum and cov_arrays. While plot_stats asks for the data stored in stats_array. Might be better to keep both?

hera_pspec/plot.py
JianrongTan: Thanks. hera_stats.stats.uvp_zscore relies on error bars stored in uvp,stats_array. And for plot_zscore_hist, it uses the information in cov_array and a outer time-difference power spectrum. Should we include these information in stats_array?

hera_pspec/utils.py
JianrongTan: Would fix it.

hera_pspec/utils.py
JianrongTan: Would fix this.

hera_pspec/uvpspec.py
JianrongTan: Good suggestions!

hera_pspec/grouping.py
JianrongTan: Thanks!

hera_pspec/pspecdata.py
```diff
@@ -514,12 +517,32 @@ def C_model(self, key, model='empirical'):
         # check cache
         if not self._C.has_key(Ckey):
             # calculate covariance model
+            if model == 'time_average':
+                spw = slice(self.spw_range[0], self.spw_range[1])
```
nkern: Let's use the objects `self.x()` method to get this data, instead of doing it yourself. That's what this method is for.

Also I don't see you weight the data before taking the time-average: you don't want flagged data in the model I presume. Try using the `self.w()` method to get the data weights and instead of `np.average`, try something like
```
np.sum(d * w, axis=0) / np.sum(w, axis=0).clip(1e-10, np.inf)
```
to create the weighted average.

hera_pspec/pspecdata.py
```diff
@@ -560,6 +586,34 @@ def cross_covar_model(self, key1, key2, model='empirical', conj_1=False, conj_2=
...
+            if conj_2:
+                x2 = x2.conj()
+            x1x2 = np.zeros((x1.shape[0], x1.shape[1], x1.shape[1]), dtype=np.complex128)
+            #(Ntimes, Nspw_freqs, Nspw_freqs)
+            for time in range(x1.shape[0]):
+                x1x2[time, :, :] = np.einsum('i,j', x1[time, :], x2[time, :])
+            covar = np.average(x1x2, axis=0)
+            #(Nspw_freqs, Nspw_freqs)
+            if subtracted == True:
+                x1_average = np.average(x1, axis=0)
```
nkern: I think my comment above also applies here: you might consider weighting the data with `self.w(key)` before taking the time average, and then performing the time-average as described above.

hera_pspec/pspecdata.py
```diff
@@ -480,7 +480,7 @@ def set_C(self, cov):
         self.clear_cache(cov.keys())
         for key in cov: self._C[key] = cov[key]
 
-    def C_model(self, key, model='empirical'):
+    def C_model(self, key, model='time_average', subtracted=True):
         """
         Return a covariance model having specified a key and model type.
```
nkern: I think you should add this comment to the line below
```
Return a covariance model having specified a key and model type.

Note: Time-dependent flags that differ from frequency channel-to-channel
can create spurious spectral structure. Consider factorizing the flags with 
self.broadcast_dset_flags() before using model='time_average'

Parameters
------------
```

hera_pspec/pspecdata.py
```diff
@@ -514,12 +517,32 @@ def C_model(self, key, model='empirical'):
...
             if model == 'empirical':
                 self.set_C({Ckey: utils.cov(self.x(key), self.w(key))})
-
+                #(Nspw_freqs, Nspw_freqs)
         return self._C[Ckey]
 
-    def cross_covar_model(self, key1, key2, model='empirical', conj_1=False, conj_2=True):
+    def cross_covar_model(self, key1, key2, model='time_average', conj_1=False, conj_2=True, subtracted=True):
         """
         Return a covariance model having specified a key and model type.
```
nkern: I think the doc-string comment above also applies here, so something like
```
Return a covariance model having specified a key and model type.

Note: Time-dependent flags that differ from frequency channel-to-channel
can create spurious spectral structure. Consider factorizing the flags with 
self.broadcast_dset_flags() before using model='time_average'

Parameters
------------
```

hera_pspec/pspecdata.py
```diff
@@ -1225,6 +1279,204 @@ def get_unnormed_V(self, key1, key2, model='empirical'):
...
+
+        Note that
+        
+        .. math ::
+            E^{12,a}_{ij}.conj = E^{21,a}_{ji}
+
+        This function estimates C^1, C^2, P^{12}, and S^{12} empirically by 
+        default. (So while the pointy brackets <...> should in principle be 
+        ensemble averages, in practice the code performs averages in time.)
+
```
nkern: I think the doc-string comment also should be inserted here, above the `Parameters` line.

hera_pspec/pspecdata.py
```diff
@@ -1851,6 +2103,15 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
             given an input visibility noise model, and store the output
             in the UVPSpec object.
 
+        save_q : boolean, optional
+            If True, store the results (delay spectra and covariance matrices) for 
+            the unnormalized bandpowers in the UVPSpec object.
```
nkern: Can we edit this slightly to read
```
        save_q : boolean, optional
            If True, store and return the results (delay spectra and covariance matrices) for 
            the unnormalized bandpowers in a separate UVPSpec object.
 ```

File: hera_pspec/pspecdata.py [hunk 15]
```diff
@@ -1851,6 +2103,15 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
...
+
+        cov_types : list of strs, optional
+            A list of the models on the input covariance matrix (the covariance matrix
+            for data stored in UVData objects). 
+            Default: ['original']
```
nkern: Let's change this to
```Default = 'original'. Options = ['original', 'time_average', ...]```
where in the `...` you should fill in all the possible options

File: hera_pspec/pspecdata.py [hunk 16]
```diff
@@ -1860,7 +2121,10 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
         Returns
         -------
         uvp : UVPSpec object
-            Instance of UVPSpec that holds the output power spectrum data.
+            Instance of UVPSpec that holds the normalized output power spectrum data.
+
+        uvp_q : UVPspec object
+            Instance of the UVPspec that holds the output unnormalized power spectrum data.
```
nkern: forgot to eliminate `output` here

nkern:
This looks like a much improved PR: great work @JianrongTan! I've gone over the core code and have only minor comments. I'm going to spend more time later to review the tests to make sure we have our bases covered, but so far looks good! 

I'm going to wait to approve until I take a look at the tests and until we hear if @philbull has any comments.

acliu:
Thanks @nkern! And to confirm, @anguta, you were ok with @JianrongTan's algebra, correct?

JianrongTan:
Thanks @nkern!

philbull:
**IMPORTANT:** I had to rewrite part of the commit history on this branch. To sync up your local copy, do `git fetch`, then `git reset origin/error_estimation --hard`, then `git pull`. This will delete any commits you made after the previous (failed) merge!

JianrongTan:
The conflicts with the master branch has been fixed. And I add an example notebook to show how to calculate the error bars in hera pspec, also includes the new feature of using the autobhaseline visibilities as input covariance when carrying on the analytic method.

philbull:
Sorry @JianrongTan, since other PRs have been merged there are new conflicts...

JianrongTan:
@philbull Yes! Now I fix the conflicts arouse from average_spectra().

JianrongTan:
@anguta I found my PR has some conflicts with your recent PR, mostly in our changes with functions related to the covariance matrix. Would you like to have some discussion here on how to fix them?

hera_pspec/testing.py
```diff
@@ -103,7 +103,10 @@ def build_vanilla_uvpspec(beam=None):
         integration_array[s] = np.ones((Nblpairts, Npols), dtype=np.float)
         nsample_array[s] = np.ones((Nblpairts, Npols), dtype=np.float)
         window_function_array[s] = np.ones((Nblpairts, Ndlys, Ndlys, Npols), dtype=np.float64)
-        cov_array[s] = np.moveaxis(np.array([[np.identity(Ndlys,dtype=np.complex) \
+        cov_array_real[s] = np.moveaxis(np.array([[np.identity(Ndlys,dtype=np.float) \
```
anguta, line 106: I'm not sure I see the prudence in splitting up our covariances into a real and imaginary part (and reworking all of our code to support this). uvp.data_array is complex. To me, the most economical solution is to just make the covariances of these complex arrays also complex.

hera_pspec/uvpspec.py
```diff
@@ -40,8 +40,9 @@ def __init__(self):
         self._cov_model = PSpecParam("cov_model", description=desc, expected_type=str)
         desc = "Power spectrum data dictionary with spw integer as keys and values as complex ndarrays."
         self._data_array = PSpecParam("data_array", description=desc, expected_type=np.complex128, form="(Nblpairts, spw_Ndlys, Npols)")
-        desc = "Power spectrum covariance dictionary with spw integer as keys and values as complex ndarrays. "
-        self._cov_array = PSpecParam("cov_array", description=desc, expected_type=np.complex128, form="(Nblpairts, spw_Ndlys, spw_Ndlys, Npols)")
+        desc = "Power spectrum covariance dictionary with spw integer as keys and values as float ndarrays, stored separately for real and imaginary parts."
```
anguta, line 43: Again, I'd like to see a good rational for doing this.

hera_pspec/pspecdata.py
```diff
@@ -2401,6 +2693,11 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
             given an input visibility noise model, and store the output
             in the UVPSpec object.
 
+        save_q : bool, optional
```
anguta, line 2696: I think it would be more economical to add a store_M option and store the M-matrices in the uvpspec object so that users can recover q by applying and inverted M to data_array.

hera_pspec/utils.py
```diff
@@ -1120,4 +1120,54 @@ def get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False, bl_len_range=(0, 1e4),
 
     return reds, lens, angs
 
+def pspecdata_time_difference(ds, time_diff):
+    """
+    Given a PSpecData object and a time difference, give the time difference PSpecData object.
```
anguta, line 1125: This seems like it could be very useful. Is the idea to estimate the covariance from time difference? I'm all for it. I think expanding the description here would be helpful since I'm not 100% sure if that's what this function is supposed to do.

hera_pspec/utils.py
```diff
@@ -1120,4 +1120,54 @@ def get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False, bl_len_range=(0, 1e4),
...
+    Parameters
+    ----------
+    ds : PSpecData object
 
+    time_diff : float
+        The time difference in seconds. 
+
+    Returns
+    -------
+    ds_td : PSpecData object
```
anguta, line 1136: Explain what is in this new pspecdata object that is returned.

hera_pspec/pspecdata.py
nkern: Well, I think that the idea here is that q_hat is already computed en-route to getting p_hat, so if by economical you mean the least amount of computation, then I think the answer is to return the pre-computed q_hat if requested (default should be False b/c this is a non-standard situation I think).

nkern: If we are going to keep this, @JianrongTan let's change the kwarg to `return_q`

hera_pspec/pspecdata.py
```diff
@@ -566,16 +566,27 @@ def set_C(self, cov):
...
         cov : dict
             Dictionary containing new covariance values for given datasets and
-            baselines. Keys of the dictionary are tuples, with the first item
-            being the ID (index) of the dataset, and subsequent items being the
-            baseline indices.
+            baselines. Keys of the dictionary are tuples. 
+            The format of a key is like: (dset_idx1, bl1, dset_idx2, bl2, model, conj_1, conj_2, subtracted), 
+            dset_idx1 and bl1 has the same form as the returns of parse_blkey. 
+            `model' should be a str, refers to the name of covariance model. 
+            `conj_1', `conj_2' and `subtracted' should all be boolean.
```
nkern, line 573: @JianrongTan the conj_1 and conj_2 seems like a perhaps unnecessary proliferation of arguments--but please correct me if I am wrong. Is it possible to adopt a convention, say x1 is conj and x2 is non-conj and then force the user to adopt those conventions? In other words, is there any compelling case for why we would want x1 and/or x2 to deviate from that convention?

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
-                       only supported if mode == 'dsets'
+                       supported if mode == 'dsets' or 'autos'
 
+        known_cov : dicts of covariance matrices
+            Covariance matrices that are not calculated internally.
+        
         Returns
         -------
-        C : array-like
+        C : array-like, (Ntimes x spw_Nfreqs x spw_Nfreqs)
```
nkern, line 618: the more standard convention is something like this
`ndarray, (Ntimes, spw_Nfreqs, spw_Nfreqs)`

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
             channel-channel covariance of each baseline over time and
             then applying the appropriate linear transformations to these
-            frequency-domain covariances.
+            frequency-domain covariances. If 'autos' is set, the covariances of the input data
+            over a baseline is estimated from the autocorrelations of the two antennas over channel bandwidth 
+            and integration time. 
 
         time_index : integer, compute covariance at specific time-step in dset
-                       only supported if mode == 'dsets'
+                       supported if mode == 'dsets' or 'autos'
```
nkern, line 611: too much indentation here

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
 
+        known_cov : dicts of covariance matrices
+            Covariance matrices that are not calculated internally.
```
nkern, line 614: you should include `see set_C()` for more details

hera_pspec/pspecdata.py
anguta: I think keeping `return_q` to False would make me happy.

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
             # add model to key
             Ckey = key + (model,)
 
+        elif model == 'autos':
+            assert isinstance(time_index, int), "time_index must be integer if cov-model==autos"
+            # add model to key
+            Ckey = key + (model, time_index)
+
+        else:
+            if known_cov == None:
```
nkern, line 643: more standard convention is `known_cov is None`

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
+                raise ValueError("didn't recognize model {}".format(model))
+            else:
+                assert isinstance(time_index, int), "time_index must be integer if cov-model=autos"
```
nkern, line 646: what is this catching? `model` (typo) is not autos b/c it passed over that conditional. and why is the Ckey so different here? See my general comment about standardizing Ckey

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
         if model == 'dsets':
-            assert isinstance(time_index, int), "time_index must be integer if cov-model=dsets"
+            assert isinstance(time_index, int), "time_index must be integer if cov-model==dsets"
+            # add model to key
             Ckey = key + (model, time_index)
 
         elif model == 'empirical':
-            assert time_index is None, "time_indexing not supported in empirical covariances"
             # add model to key
             Ckey = key + (model,)
```
nkern, line 635: I think we should include time_index in Ckey even though it is None to conform to a standard.

nkern:
@JianrongTan @anguta Seems like there may be conflicting standards for the `Ckey` we are using for caching covariances. I think we need to be very standard about this b/c it can easily lead to repeated calculations and memory bloat if we are using this in different ways (similar issues have come up with `Rkey`).

I think we should define our standard clearly in the `set_C()` docstring, and abide by that for all future functionality that wants to use covariance caching. @JianrongTan perhaps something like this will suffice
```
    def set_C(self, cov):
        """
        Set the cached covariance matrix to a set of user-provided values.

        Parameters
        ----------
        cov : dict
            Covariance keys and ndarrays.
            The key should conform to 
            (dset_index, blpair_int, model, time_index, <anything else @jianrong wants>)
            while the ndarrays should have shape (@JianrongTan whatever shape is appropriate)
        """
```

hera_pspec/pspecdata.py
```diff
@@ -585,54 +596,105 @@ def C_model(self, key, model='empirical', time_index=None):
...
         # check cache
         if Ckey not in self._C:
             # calculate covariance model
             if model == 'empirical':
                 self.set_C({Ckey: utils.cov(self.x(key, filter_extension=True), self.w(key, filter_extension=True))})
             elif model == 'dsets':
                 self.set_C({Ckey: np.diag( np.abs(self.w(key, filter_extension=True)[:,time_index] * self.dx(key, filter_extension=True)[:,time_index]) ** 2. )})
+            elif model == 'autos':
+                """
+                Pick a baseline $b=(alpha,beta)$ where $alpha$ and $beta$ are antennas,
```
nkern, line 659: this I think needs its own function with a clearer docstring explanation.
can we put this in `hera_pspec.utils` and then call it here?

hera_pspec/pspecdata.py
```diff
@@ -658,31 +725,54 @@ def cross_covar_model(self, key1, key2, model='empirical',
             Whether to conjugate second copy of data in covar or not.
             Default: True
 
+        known_cov : dicts of covariance matrices
+            Covariance matrices that are not calculated internally from data.
+
         Returns
         -------
-        cross_covar : array-like, spw_Nfreqs x spw_Nfreqs
+        cross_covar : array-like, (spw_Nfreqs x spw_Nfreqs)
```
nkern, line 733: again should probably be `(spw_Nfreqs, spw_Nfreqs)`

hera_pspec/pspecdata.py
```diff
@@ -658,31 +725,54 @@ def cross_covar_model(self, key1, key2, model='empirical',
...
-        if model == 'empirical':
-            # parse key
-            dset, bl = self.parse_blkey(key1)
-            key1 = (dset,) + (bl,)
-            dset, bl = self.parse_blkey(key2)
-            key2 = (dset,) + (bl,)
 
+        internal_models = ['empirical', 'dsets', 'autos']
+        # internal_models refer to the covariance can be calculated intrinsically. 
+        if known_cov == None:
```
nkern, line 743: `known_cov is None`

hera_pspec/pspecdata.py
```diff
@@ -658,31 +725,54 @@ def cross_covar_model(self, key1, key2, model='empirical',
...
+        if model == 'empirical':
             covar = utils.cov(self.x(key1, filter_extension=True), self.w(key1, filter_extension=True),
                               self.x(key2, filter_extension=True), self.w(key2, filter_extension=True),
                               conj_1=conj_1, conj_2=conj_2)
+        
         elif model == 'dsets':
             covar = np.zeros((self.spw_Nfreqs,
                               self.spw_Nfreqs),
-                              dtype=complex)
+                              dtype=float)
```
nkern, line 760: lets be explicit about dtype when possible, perhaps use np.float64

hera_pspec/pspecdata.py
JianrongTan: So do you mean `return_q` refers to store the unnormalized one other than the normalized one? So we don't have double outputs for `uvp` and `uvp_q`, instead, we have output for either `uvp` or `uvp_q`?

hera_pspec/pspecdata.py
JianrongTan: Sure. Thanks!

hera_pspec/pspecdata.py
JianrongTan: Sure.

hera_pspec/testing.py
anguta: OK. Now I realize that a single complex array is not enough information to describe the error bars on both the real and imaginary component of a power-spectrum estimate, essentially because there are 2 x (N+1) x N / 2 degrees of freedom for the imag/real covariances but a single complex covariance only has (N-1) x N / 2 + N degrees of freedom (since the diagonal has to be real). I think it does make sense to move to real/imag arrays. Thank you @JianrongTan / @acliu for explaining this.

hera_pspec/pspecdata.py
JianrongTan: Sure. Thanks!

hera_pspec/uvpspec.py
anguta: This has been satisfactorily explained. Perhaps consider adding some block comments about why this is.

anguta:
> This pull request does some major reorganization of outputs.
> 
> This includes splitting up the complex covariance into a real and imaginary float part in `uvpspec` and saving the unnormalized bandpowers (and their covariances) as a separate uvdata object. The former is already achieved by storing the covariance as a complex array and the latter could be done more economically by just saving the M-matrix.
> 
> Since this major reorganization will break downstream code, I think there needs to be a good rational for implementing these changes. I'm not sure this exists since real/imag covariances are already stored in the complex covariance variable and q-matrices can be backed out if we just store the M-matrix instead. We could even add a function to `uvpspec` that recalculates `q` from `M` and `data_array` along with the covariance of `q`.
> 
> Finally, I think that this does come into conflict with the covariance infrastructure that is aready being built up wherein the type of covariance (analytic, empirical etc...) is specified by a model keyword in get_unnormed_V, C rather then creating entirely separate functions for these.
> 
> I'm open to supporting more specific functions but I think we should make sure that the routing through get_unnormed_V, C, is preserved.

After some explanation, I'm happy with real/imag covariances and am convinced that these are necessary given how our estimator works. I can sign off on storing the q arrays if as long as by default, they are not returned (an easy compromize). 

The last item is how to organize covariance calculations.

JianrongTan:
> @JianrongTan @anguta Seems like there may be conflicting standards for the `Ckey` we are using for caching covariances. I think we need to be very standard about this b/c it can easily lead to repeated calculations and memory bloat if we are using this in different ways (similar issues have come up with `Rkey`).
> 
> I think we should define our standard clearly in the `set_C()` docstring, and abide by that for all future functionality that wants to use covariance caching. @JianrongTan perhaps something like this will suffice
> 
> ```
>     def set_C(self, cov):
>         """
>         Set the cached covariance matrix to a set of user-provided values.
> 
>         Parameters
>         ----------
>         cov : dict
>             Covariance keys and ndarrays.
>             The key should conform to 
>             (dset_index, blpair_int, model, time_index, <anything else @jianrong wants>)
>             while the ndarrays should have shape (@JianrongTan whatever shape is appropriate)
>         """
> ```
Sure. I suggest Ckey follows a general format (dset_index, blpair_int, model, time_index, conj_1, conj_2).

hera_pspec/pspecdata.py
JianrongTan: Sure thing.

anguta:
> > @JianrongTan @anguta Seems like there may be conflicting standards for the `Ckey` we are using for caching covariances. I think we need to be very standard about this b/c it can easily lead to repeated calculations and memory bloat if we are using this in different ways (similar issues have come up with `Rkey`).
> > I think we should define our standard clearly in the `set_C()` docstring, and abide by that for all future functionality that wants to use covariance caching. @JianrongTan perhaps something like this will suffice
> > ```
> >     def set_C(self, cov):
> >         """
> >         Set the cached covariance matrix to a set of user-provided values.
> > 
> >         Parameters
> >         ----------
> >         cov : dict
> >             Covariance keys and ndarrays.
> >             The key should conform to 
> >             (dset_index, blpair_int, model, time_index, <anything else @jianrong wants>)
> >             while the ndarrays should have shape (@JianrongTan whatever shape is appropriate)
> >         """
> > ```
> 
> Sure. I suggest Ckey follows a general format (dset_index, blpair_int, model, time_index, conj_1, conj_2), where conj_1 and conj_2 are boolean values.

@JianrongTan. I think this is a great idea.

hera_pspec/pspecdata.py
nkern: Yeah that could be confusing--here's a question, do you ever find yourself needing both p_hat and q_hat? If not, then you could make `return_q=False` just return uvp and `return_q=True` just return uvp_q. Otherwise you can make it `return uvp, uvp_q` and if `return_q=False` set `uvp_q=None`

hera_pspec/pspecdata.py
```diff
@@ -1621,6 +1686,208 @@ def get_unnormed_V(self, key1, key2, model='empirical', exact_norm=False, pol =
...
+            P11.append(self.cross_covar_model(key1, key1, model=model, conj_1=False, conj_2=False, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+            S11.append(self.cross_covar_model(key1, key1, model=model, conj_1=True, conj_2=True, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+            P22.append(self.cross_covar_model(key2, key2, model=model, conj_1=False, conj_2=False, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+            S22.append(self.cross_covar_model(key2, key2, model=model, conj_1=True, conj_2=True, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+            P21.append(self.cross_covar_model(key2, key1, model=model, conj_1=False, conj_2=False, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+            S21.append(self.cross_covar_model(key2, key1, model=model, conj_1=True, conj_2=True, known_cov=known_cov, time_index=time_index)[np.newaxis,:,:])
+        # (Ntimes, 1, spw_Nfreqs, spw_Nfreqs)
+        
+        if (C11[0] == C11[-1]).all():
```
nkern, line 1819: not good to use `==` for comparing data that is floating point--unless you expect it to be *exactly* the same. try using `np.isclose(a, b, atol=xx).all()` instead

hera_pspec/pspecdata.py
```diff
@@ -1621,6 +1686,208 @@ def get_unnormed_V(self, key1, key2, model='empirical', exact_norm=False, pol =
...
+        # if the covariance matrix is uniform along the time axis
```
nkern, line 1820: indent

hera_pspec/uvpspec_utils.py
```diff
@@ -676,16 +684,17 @@ def _select(uvp, spws=None, bls=None, only_pairs_in_bls=False, blpairs=None,
...
-        cov = odict()
+        cov_real = odict()
+        cov_imag = odict()
         stats = odict()
         window_function = odict()
 
         # determine if certain arrays are stored
         if h5file is not None:
-            store_cov = 'cov_spw0' in h5file
+            store_cov = 'cov_real_spw0' in h5file
```
nkern, line 694: below this I think we should add
```
if 'cov_array_spw0' in h5file:
    store_cov = False
    warnings.warn("uvp.cov_array is no longer supported and will not be loaded. Please update this to be uvp.cov_array_real and uvp.cov_array_imag. See hera_pspec PR #181 for details.")

hera_pspec/pspecdata.py
JianrongTan: Sure.

hera_pspec/pspecdata.py
JianrongTan: Nice catch!

hera_pspec/uvpspec_utils.py
JianrongTan: Nice suggestion!

JianrongTan:
> I think this looks good--thanks for the hard work. I think you should be set to merge if you are fine with it

Thanks! While it says I am not authorized to merge this pull request yet, you may help to merge it.

PR #182, opened 2018-11-30
Fix problems with spw select and polstr

philbull:
(1) Fixes a problem with `UVPSpec.select()` assigning data arrays from the wrong spectral windows when `spws` is not `None`.
(2) Fixes issue with polstr conversion in `select()`.
(3) Fixes unit test failure caused by change in ordering of bl tuple elements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
     pip: true
 
 env:
-  - PYUVDATA_VERSION="@fae18f1a97e6c521b822b2ffc3c1afef4e2ac3b3"
+  - PYUVDATA_VERSION="@d6e1c2f80d7b5b59d7794c8f28d867a7030a7c75"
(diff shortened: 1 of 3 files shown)


nkern:
Thanks for the fixes @philbull ! Looking at the code I see now that (2) was just a plain oversight on my part (and the case of not having a unit-test to check that the relevant module wasn't imported).

(3) is weird b/c I don't see this failure locally. I'm looking into (2) now and why this wasn't caught in our units tests...

philbull:
@nkern: Did you empty the cache for `master` on Travis?

nkern:
Just for posterity, after emptying the travis cache the master branch still passes its tests, so something seems weird w/ Phil's local setup. But there is some legit bug fixes in this PR, so we will still merge it in.

PR #189, opened 2019-03-03
Enable cross-pol power spectra *and* Python 3 support

philbull:
This PR enables cross-polarisation support, subject to the limitation that the beam scalar can't be computed for cross-pol yet. (In other words, this PR only allows unnormalised cross-pol spectra to be computed.)

The main change in the PR is to store polarisation pairs inside `UVPSpec`, instead of individual polarisations. Pol-pairs can be specified as tuples, e.g. `('xx', 'xx')`, or as specially-constructed integers (there are helper functions to deal with these). Individual polarisations (e.g. `'xx'`) are no longer accepted.

I also got a bit carried away and enabled Python 3 support in this PR. This required more extensive changes than first expected, as functions like `map()` and `reduce()` have been deprecated (now replaced with list comprehensions and `np.reduce`), some string handling has changed (especially annoying for HDF5), and methods like `dict.keys()` now return iterators instead of lists, and so can't be indexed directly (fixed by calling `list()` on the iterators).

### DIFF
--- a/hera_pspec/pspecbeam.py
+++ b/hera_pspec/pspecbeam.py
@@ -277,10 +277,11 @@ class PSpecBeamBase(object):
             if isinstance(pol2, (int, np.integer)):
                 pol2 = uvutils.polnum2str(pol2)
             
-            # Check for cross-pol; only auto-pol calculation currently supported
+            # Check for cross-pol; only same-pol calculation currently supported
             if pol1 != pol2:
                 raise NotImplementedError(
-                        "get_Omegas does not support cross-polarized beams yet. "
+                        "get_Omegas does not support cross-correlation between "
+                        "two different visibility polarizations yet. "
                         "Could not calculate Omegas for (%s, %s)" % (pol1, pol2))
             
             # Calculate Omegas
--- a/hera_pspec/pspecbeam.py
+++ b/hera_pspec/pspecbeam.py
@@ -1,6 +1,7 @@
 import numpy as np
 import os
 import hera_pspec.conversions as conversions
+import hera_pspec.uvpspec_utils as uvputils
 import scipy.integrate as integrate
 from scipy.interpolate import interp1d
 from pyuvdata import UVBeam, utils as uvutils
@@ -71,7 +72,7 @@ def _compute_pspec_scalar(cosmo, beam_freqs, omega_ratio, pspec_freqs,
     
     # Get redshifts and cosmological functions
     redshifts = cosmo.f2z(integration_freqs).flatten()
-    X2Y = np.array(map(lambda z: cosmo.X2Y(z, little_h=little_h), redshifts))
+    X2Y = np.array([cosmo.X2Y(z, little_h=little_h) for z in redshifts])
 
     # Use linear interpolation to interpolate the frequency-dependent 
     # quantities derived from the beam model to the same frequency grid as the 
@@ -238,14 +239,15 @@ class PSpecBeamBase(object):
         return 1e-20 * conversions.cgs_units.c**2 \
                / (2 * conversions.cgs_units.kb * freqs**2 * Op)
 
-    def get_Omegas(self, pols):
+    def get_Omegas(self, polpairs):
         """
-        Get OmegaP and OmegaPP across beam_freqs for requested polarizations.
+        Get OmegaP and OmegaPP across beam_freqs for requested polarization 
+        pairs.
 
         Parameters
         ----------
-        pols : list
-            List of polarization strings or integers.
+        polpairs : list
+            List of polarization-pair tuples or integers.
 
         Returns
         -------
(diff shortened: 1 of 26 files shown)


philbull:
The tests are only failing because of a weird bug with the print statement in `version.py` in Python 2.x.

hera_pspec/uvpspec.py
```diff
@@ -53,8 +53,8 @@ def __init__(self):
         self._spw_freq_array = PSpecParam("spw_freq_array", description="Spw integer array for the freq_array.", form="(Nspwfreqs,)", expected_type=np.uint16)
         self._freq_array = PSpecParam("freq_array", description="Frequency array of the original data in Hz.", form="(Nfreqs,)", expected_type=np.float64)
         self._dly_array = PSpecParam("dly_array", description="Delay array in seconds.", form="(Nspwdlys,)", expected_type=np.float64)
-        desc = "Polarization integers of power spectra. Stokes 1:4 (I,Q,U,V); circular -1:-4 (RR,LL,RL,LR); linear -5:-8 (XX,YY,XY,YX)"
-        self._pol_array = PSpecParam("pol_array", description=desc, form="(Npols,)", expected_type=np.int32)
+        desc = "Polarization pair integer, made up of two "
```
nkern: unfinished sentence here, looks like

hera_pspec/uvpspec_utils.py
```diff
@@ -199,73 +210,164 @@ def select_common(uvp_list, spws=True, blpairs=True, times=True, pols=True,
...
+    else:
+        return (pol1, pol2)
+    
+
+def polpair_tuple2int(polpair):
+    """
+    Convert a tuple pair of polarization strings/integers into 
+    an pol-pair integer.
+    
+    The polpair integer is formed by adding 10 to each standardized
```
nkern: Seems like we might want to make this `adding 20 to each standardized polarization integer and then concatenating them`, because the dipole polarizations are `-5, -6, -7, -8`, which means that in the former case `XX,XX` would be `55`, which wouldn't follow a standard convention of a len-4 number, but in the latter case would be `1515`. LMK what you think.

We did a similar thing to `blpair` integers recently to make this more standardized.

philbull: I should have pointed out that there is a leading zero, so `('xx', 'xx')` -> `505` (unless you think it should always be a len-4 number).

hera_pspec/uvpspec.py
philbull: Good catch.

hera_pspec/uvpspec_utils.py
nkern: I see. Yeah I guess its not a big deal because there is still a 1-to-1 mapping, so its up to you. For whatever reason, I find the integers themselves more readable when there is a set number of integers, like `1515` over `505`, but b/c you provide convenience functions for mapping them to strings its not a big deal to me!

philbull: You've convinced me! I'm adding 20 now, so they will all be len-4 integers.

philbull:
OK, I managed to fix the test failure (it was due to a change in `io.StringIO` behaviour between Py2 and Py3, used by the test suite). I also made some improvements in test coverage and caught a bug, so coveralls passes now too.

hera_pspec/pspecbeam.py
```diff
@@ -255,24 +257,38 @@ def get_Omegas(self, pols):
...
-        for p in pols:
-            OmegaP.append(self.power_beam_int(pol=p))
-            OmegaPP.append(self.power_beam_sq_int(pol=p))
+        for pol1, pol2 in polpairs:
+            if isinstance(pol1, (int, np.integer)):
+                pol1 = uvutils.polnum2str(pol1)
+            if isinstance(pol2, (int, np.integer)):
+                pol2 = uvutils.polnum2str(pol2)
+            
+            # Check for cross-pol; only auto-pol calculation currently supported
```
nkern: looks like auto-pol is used here ;) hehe. perhaps same-pol?

hera_pspec/pspecbeam.py
```diff
@@ -255,24 +257,38 @@ def get_Omegas(self, pols):
...
+            if pol1 != pol2:
+                raise NotImplementedError(
+                        "get_Omegas does not support cross-polarized beams yet. "
```
nkern: I think `cross-polarized beams` is slightly different, something like `xy` is I think also referred to as a cross-polarized beam. I think here we mean, `get_Omegas does not support cross correlation between two different visibility polarizations yet`.

hera_pspec/pspecdata.py
```diff
@@ -2027,33 +2076,41 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
...
                 pol_ints = []
                 pol_cov=[]
 
                 # Compute scalar to convert "telescope units" to "cosmo units"
                 if self.primary_beam is not None:
-                    # using zero'th indexed poalrization as cross polarized beam are not yet implemented
+                    
+                    # Raise error if cross-pol is requested
+                    if (p[0] != p[1]):
+                        raise NotImplementedError("Cross-polarized beams are "
```
nkern: same comment on `cross-polarized` beams

hera_pspec/pspecdata.py
```diff
@@ -1805,9 +1838,11 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None, input_data_weight='identit
...
             is used for the Right-Hand dataset (see above).
 
-        pols : length-2 tuple of strings or integers or list of length-2 tuples of strings or integers
+        pols : length-2 tuple of strings or integers or list of length-2 
+            tuples of strings or integers
             Contains polarization pairs to use in forming power spectra
-            e.g. ('XX','XX') or [('XX','XX'),('XY','YX')] or list of polarization pairs.
+            e.g. ('XX','XX') or [('XX','XX'),('pI','pI')] or list of 
+            polarization pairs.
             Only auto/equal polarization pairs are implemented at the moment.
```
nkern: although you didn't add this, maybe good to get this consistent: per UW requests perhaps we change this to `Only identical polarization pairs...`?

hera_pspec/uvpspec.py
```diff
@@ -853,29 +903,29 @@ def key_to_indices(self, key, omit_flags=False):
         """
         Convert a data key into relevant slice arrays. A data key takes the form
 
-        (spw_integer, ((ant1, ant2), (ant3, ant4)), pol_string)
+        (spw_integer, ((ant1, ant2), (ant3, ant4)), (pol12, pol34))
```
nkern: small comment: I think `pol12, pol34` is a little confusing b/c throughout the code we adopt `pol1, pol2`, where `pol` is a visibility polarization, not a feed polarization. So I'd stick to the convention that `pol` refers to visibility not feed polarization, such that this reads `(pol1, pol2)`.

hera_pspec/uvpspec.py
```diff
@@ -853,29 +903,29 @@ def key_to_indices(self, key, omit_flags=False):
...
-        and pol is either a polarization string (ex. 'XX') or integer (ex. -5).
+        and polpair is either a polarization-pair integer or tuple.
 
         The key can also be a dictionary in the form::
 
           key = {
             'spw' : spw_integer,
             'blpair' : ((ant1, ant2), (ant3, ant4))
-            'pol' : pol_string
+            'polpair' : (pol12, pol34)
```
nkern: same here

nkern:
@philbull just for posterity: on the hera_pspec telecon 3/5/18 I think we decided to punt on cross-visibility-polarization scalars, and that cross-visibility-pol jacknives should be done in "data units" rather than "cosmo units"--so no scalar.

Also, I take back my support for interpreting `polpairs = ['XX']` as `polpairs = [('XX', 'XX')]`, because I think that can easily lead to confusion over 'XX' being a "polarization-pair", which it isn't in our definition, although it is technically a "feed-polarization-pair". Up to you @philbull what you want to do about it though..

nkern:
@philbull so I've finished my skim of the PR. it looks good to me with minor requests for revision. ping me when you are done w/ edits and i'll approve!

hera_pspec/pspecbeam.py
philbull: Done!

hera_pspec/pspecbeam.py
philbull: Great, done.

hera_pspec/pspecdata.py
philbull: Good catch, done.

hera_pspec/pspecdata.py
philbull: I rewrote this as it's a bit confusing anyway.

philbull:
Thanks Nick! I've made all the changes you suggested, and improved test coverage slightly to get that coveted green checkmark.

PR #193, opened 2019-03-25
Concurrent access to PSpecContainer

philbull:
Add a 'transactional' mode and enable HDF5's "single writer, multiple reader" (SWMR) mode to allow concurrent access to `PSpecContainer` objects.

The transactional mode is enabled when the container is instantiated with the `keep_open=False` kwarg. This keeps the HDF5 file closed except for when actions are being performed on it, when it is briefly opened and then closed again. This is helpful if multiple processes need to write out to the same file for example, as HDF5 files cannot be opened in read-write mode by more than one process.

The SWMR mode allows multiple processes to open the file for reading at the same time, as long as at most one process has opened it in read-write mode. Generally, the read-write process should open the file first; if a read-only version has been opened beforehand, opening in read-write mode may fail.


hera_pspec/container.py
```diff
@@ -23,24 +52,63 @@ def __init__(self, filename, mode='r'):
...
+        *first* one that is created; if a read-only instance is already open 
+        when a rw instance is created, an error will be raised by h5py. 
         """
+        if self.data is not None: return
+        
         # Convert user-specified mode to a mode that HDF5 recognizes. We only
         # allow non-destructive operations!
         mode = 'a' if self.mode == 'rw' else 'r'
-        self.data = h5py.File(self.filename, mode)
+        swmr = True if self.mode == 'r' else False
```
plaplant: Here's a case where I think it'd be much more obvious to break this out into a traditional multi-line if/else block. The line above also uses the ternary conditional, which (for me) is hard to parse. You might even be able to handle these different cases within a single if/elif/else series of conditionals.

hera_pspec/container.py
```diff
@@ -23,24 +52,63 @@ def __init__(self, filename, mode='r'):
...
+        
+        try:
+            self.data = h5py.File(self.filename, mode, libver='latest', swmr=swmr)
```
plaplant: It's important to pass the `libver="latest"` keyword here, but before that you should do a check that the version of the underlying HDF5 library is >= 1.10. I think one way forward is to call `h5py.version.hdf5_version_tuple` (which will return something like `(1, 10, 4)`), and just check that it's `>= (1, 10, 0)`. (No one should be using an ancient installation of HDF5 these days, but you never know!)

hera_pspec/container.py
```diff
@@ -23,24 +52,63 @@ def __init__(self, filename, mode='r'):
...
+            if self.mode == 'rw':
+                try:
+                    # Enable single writer, multiple reader mode on HDF5 file. 
+                    # This allows multiple handles to exist for the same file 
+                    # at the same time, as long as only one is in rw mode
+                    self.data.swmr_mode = True
+                except ValueError:
```
plaplant: Is this the symptom of opening a file using an older version of the HDF5 library? I wonder if it might be better to warn the user that SWMR mode is unavailable due to their HDF5 library version, rather than silently passing.

hera_pspec/container.py
```diff
@@ -50,7 +118,15 @@ def _open(self):
             # Denote as Container
             if 'pspec_type' not in list(self.data.attrs.keys()):
                 self.data.attrs['pspec_type'] = self.__class__.__name__
-
+    
+    def _close(self):
+        """
+        Close HDF5 file. DOes nothing if file is already closed.
+        """
+        if self.data is None: return
```
plaplant: Single-line if

hera_pspec/container.py
```diff
@@ -356,13 +435,16 @@ def combine_psc_spectra(psc, groups=None, dset_split_str='_x_', ext_split_str='_
...
+    # Load container
     if isinstance(psc, (str, np.str)):
         psc = PSpecContainer(psc, mode='rw')
     else:
         assert isinstance(psc, PSpecContainer)
-
-    # get groups
+    
+    # Check whether PSpecContainer is in transactional mode and a lock is defined
+
```
plaplant: Is there supposed to be some code to do this checking underneath this comment?

hera_pspec/pspecdata.py
```diff
@@ -2902,7 +2905,7 @@ def pspec_run(dsets, filename, dsets_std=None, groupname=None,
             bls2_list.append(_bls2)
 
     # Open PSpecContainer to store all output in
-    psc = container.PSpecContainer(filename, mode='rw')
+    psc = container.PSpecContainer(filename, mode='rw', keep_open=False)
```
plaplant: Is it best to hard-code `keep_open=False` here? (I'm genuinely asking, and I don't know what the larger context here is.)

hera_pspec/container.py
nkern: Because we implicitly set swmr=True for mode=r, raising an exception would mean ppl with old HDF5 installations can't even use the basic functionality of PSpecContainer, let alone its swmr features.

What I've done is to raise a warning if swmr=True and their HDF5 installation is outdated.

hera_pspec/pspecdata.py
nkern: yes i think its right to hardcode it as false

hera_pspec/container.py
nkern: I don't think this is the case, as one can set an attribute of the file descriptor without error even if the attribute `swmr_mode` didnt' originally exist..

Issue #194, opened 2019-03-26
Current master is not backwards compatible with old hdf5 files


plaplant:
It seems that the current version of master is no longer backwards compatible with old files. I am not sure when the transition happened, but there are no clues to the user what is wrong or how to read in older files. Trying to read a pspec object generated with an older version of the code (located at `/lustre/aoc/projects/hera/plaplant/HERA19Golden/CalibratedData/LSThrs_10.5_23.0/FREQrng_530_730/zen.2457548.16700_66112.pI.HH.uvcRPCS.uvh5.UVPavg`) yields this traceback:

```
In [14]: uvp.read_hdf5("zen.2457755.58922_11048.pI.HH.uvCP.uvh5.UVPavg")
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-14-ee8853d6d1eb> in <module>
----> 1 uvp.read_hdf5("zen.2457755.58922_11048.pI.HH.uvCP.uvh5.UVPavg")

/lustre/aoc/projects/hera/plaplant/anaconda2/envs/py3/lib/python3.6/site-packages/hera_pspec-0.1.0-py3.6.egg/hera_pspec/uvpspec.py in read_hdf5(self, filepath, just_meta, spws, bls,
 blpairs, times, lsts, polpairs, only_pairs_in_bls)
   1221             self.read_from_group(f, just_meta=just_meta, spws=spws, bls=bls,
   1222                                  times=times, lsts=lsts, polpairs=polpairs,
-> 1223                                  only_pairs_in_bls=only_pairs_in_bls)
   1224
   1225

/lustre/aoc/projects/hera/plaplant/anaconda2/envs/py3/lib/python3.6/site-packages/hera_pspec-0.1.0-py3.6.egg/hera_pspec/uvpspec.py in read_from_group(self, grp, just_meta, spws, bls
, blpairs, times, lsts, polpairs, only_pairs_in_bls)
   1141         assert 'pspec_type' in grp.attrs, "This object is not a UVPSpec object"
   1142         assert grp.attrs['pspec_type'] == 'UVPSpec', \
-> 1143             "This object is not a UVPSpec object"
   1144
   1145         # Clear all data in the current object

AssertionError: This object is not a UVPSpec object

In [15]: uvp = hera_pspec.UVPSpec()

In [16]: uvp.read_hdf5("zen.2457755.58922_11048.pI.HH.uvCP.uvh5.UVPavg")
[...]

PR #195, opened 2019-03-26
Fix backwards compatibility bugs with UVPSpec data files

philbull:
Fix two backwards compatibility bugs related to loading old HDF5 files into `UVPSpec` objects.

The first is related to HDF5 in Py3, which returns `bytes` instead of strings. A string comparison was failing as a result. All string attributes in `UVPSpec` files are now converted from `bytes` to strings.

The second is due to the recent change from `pol_array` to `polpair_array` in the `UVPSpec` data model. The old `pol_array` attribute is now detected and automatically converted to a `polpair_array` when a file is loaded. A `UserWarning` is shown when this happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,13 +40,13 @@ install:
   - conda install -c conda-forge healpy aipy
   - pip install coveralls
   - pip install h5py
+  - pip install scikit-learn
   - pip install git+https://github.com/HERA-Team/pyuvdata.git$PYUVDATA_VERSION
   - pip install git+https://github.com/HERA-Team/omnical.git
   - pip install git+https://github.com/HERA-Team/linsolve.git
   - pip install git+https://github.com/HERA-Team/hera_qm.git
   - pip install git+https://github.com/HERA-Team/uvtools.git
   - pip install git+https://github.com/HERA-Team/hera_cal.git
-  - pip install scikit-learn
   - pip install pyyaml
   - pip install multiprocess
   - python setup.py install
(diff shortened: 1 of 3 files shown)


philbull:
This should fix issue #194.

.travis.yml
```diff
@@ -35,18 +35,18 @@ install:
   - conda info -a
 
   # create environment and install dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy nose pip matplotlib coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy>=1.15.0 scipy nose pip matplotlib coverage
   - source activate test-environment
-  - conda install -c conda-forge healpy aipy scikit-learn
+  - conda install -c conda-forge healpy aipy
```
nkern: hera_qm and hera_cal, while not strictly required, have some advanced functionality that relies on `scikit-learn`, so it'd be nice to keep this if possible. was something breaking with the installation?

plaplant:
Thanks for the quick turnaround @philbull!

.travis.yml
philbull: Thanks Nick. The reason from removing `scikit-learn` from this line is that it was pulling in an old version of `numpy`, which is now incompatible with `pyuvdata` (thus causing the tests to fail). It *is* installed via `pip` later on in the Travis script however (pip allows it to install with a more recent version of `numpy`).

PR #196, opened 2019-04-06
fix for PSpecData.pspec to pass the little_h option when calculating …

zacharymartinot:
…the normalization scalar. I didn't look extensively for any other instances of this bug, but this change produced the expected (non-little-h) output power spectrum amplitude.


philbull:
This is good. I did notice this, but wherever I fixed it hasn't been merged yet, so let's just merge this PR.

Issue #152, opened 2018-07-08
Validation tests for averaging covariance matrices


anguta:
There are not tests validating that the averaged covariance matrices match our expectations for the weighted average of independent measurements.  These should be added.

PR #200, opened 2019-04-22
Update covariance averaging in grouping.average_spectra

JianrongTan:
An update on how to average the covariance with weights. Should be like $cov = \sum{w_i**2 cov_i } /  (\sum w_i)**2$, where $w_i$ are weights.


nkern:
@JianrongTan can you look into propagating these changes to how we average errorbars? I think the averaging of errorbars should be the same way we average covariances, so if we update one we should do so for the other. I believe that our errorbars are generally stored as standard deviations, so you'll probably want to square them into a variance before doing the averaging, if that's not already done...

nkern:
@anguta this might be a good time to address #152 if you want to contribute to this PR. Or we can put it off for later..

JianrongTan:
@nkern Thanks for comments! If I remember correctly, the information of other error bars is just stored in uvpspec.stats_array, so I will update that accordingly...

PR #202, opened 2019-05-10
Use older version of scikit-learn for Python 2.7 tests

philbull:
Scikit-learn has dropped Py2.7 support; use older versions for Py2.7 tests.


saurabh-astro (approved):
Looks good!

PR #208, opened 2019-06-15
Handle bytes/str comparisons in PSpecContainer

philbull:
Simple change to improve bytes vs string handling in HDF5 strings.


nkern:
@philbull when this passes tests I'm happy w/ it

philbull:
Awesome, thanks. I think the test failure is intermittent due to the IERS table not updating in `astropy`.

philbull:
@nkern This is now passing its tests.

PR #210, opened 2019-06-21
Get redundant groups contained within a UVPSpec object

philbull:
Adds two new methods on `UVPSpec` to return a list of redundant baselines (or baseline-pairs), grouped according to redundant bl group.

For the case of baseline-pairs, only pairs that come from the same redundant group are listed.


philbull:
This is currently waiting on a bugfix in `uvtools`: https://github.com/HERA-Team/uvtools/pull/37

philbull:
Bugfix merged.

hera_pspec/uvpspec_utils.py
```diff
@@ -945,3 +946,145 @@ def _fast_lookup_blpairts(src_blpts, query_blpts, time_prec=8):
     blpts_idxs = np.where(src_blpts == query_blpts[:,np.newaxis])[1]
 
     return blpts_idxs
+
+
+def _get_red_bls(uvp, bl_len_tol=1., bl_ang_tol=1.):
```
nkern: seems like no need for these to be "hidden" functions? These seem well-documented and useful, so might as well make the un-underscored?

philbull: They're exposed as methods in the `UVPSpec` object, so they should be reasonably visible. I just put them here to offload some code into `uvpspec_utils` (like some similar functions in this file).

nkern: ok great

hera_pspec/uvpspec_utils.py
```diff
@@ -945,3 +946,145 @@ def _fast_lookup_blpairts(src_blpts, query_blpts, time_prec=8):
...
+    grp_lens : list of float
+        Average length of the baselines in each group.
+        
+    grp_angs : list of float
+        Average angle of the baselines in each group.
+    """
+    # Calculate length and angle of baseline vecs
+    bl_vecs = uvp.get_ENU_bl_vecs()
+    
+    lens, angs = utils.get_bl_lens_angs(bl_vecs, bl_error_tol=1.0)
```
nkern: this should be passed from `bl_len_tol`

philbull: Good catch, thanks!

PR #211, opened 2019-06-29
Allow pspec pipeline to read other file types

philbull:
Some parts of the pspec pipeline are currently hard-coded to use miriad files. This PR allows other filetypes to be used.


philbull:
Great, thanks!

PR #217, opened 2019-08-01
Refactoring exact norm

Ronan1879:
In pspecdata.py, in the definition of q_hat (line 956), there is a for loop present (line 1052) in which within this loop, we calculate Q as Q = del_tau * self.get_Q(i, pol). This for loop is going over each delay mode "i" in which get_Q takes as a parameter. If we go to the function get_Q (around line 1557), it computes the exponential term and the integral of the primary beam. However, the integral does not depend on the delay mode. So back in the q_hat function, the integral is being calculated in the loop for every delay mode when in fact its result is constant. To change this, I created a separate function in pspecdata.py called get_integral_beam where the code that was in get_Q to compute the integral has been transferred to this new function. Outside of the for loop, I call get_integral_beam and store its result in a variable called integral_beam and then multiply it inside the loop as Q = del_tau * self.get_Q(i, pol) * integral_beam. This way we are always using the same result for each delay mode. This makes generating the power spectrum faster using exact_norm = True as a parameter in pspec. This change does not affect the time to generate the power spectrum when exact_norm is set to "False", as it doesn't use this for loop in q_hat. I also wrote a test function called test_get_integral_beam in test_pspecdata.py for my new function get_integral_beam.


hera_pspec/pspecdata.py
```diff
@@ -1515,7 +1517,43 @@ def get_Q_alt(self, mode, allow_fft=True):
...
+
+        pol : str/int/bool, optional
+            Which beam polarization to use. If the specified polarization doesn't exist, 
+            a uniform isotropic beam (with integral 4pi for all frequencies) is assumed. 
+            Default: False (uniform beam).
+
+        Return
+        -------
+        Q : array_like
+            Response matrix for bandpower p_alpha.
```
saurabh-astro: This should be changed to returning the beam integral.

hera_pspec/tests/test_pspecdata.py
```diff
@@ -1115,7 +1138,7 @@ def test_pspec(self):
         bls_Q   = [(24, 25)]
         uvp = ds_Q.pspec(bls_Q, bls_Q, (0, 1), [('xx', 'xx')], input_data_weight='identity',
                                        norm='I', taper='none', verbose=True, exact_norm=False)
-        Q_sample = ds_Q.get_Q((ds_Q.spw_range[1] - ds_Q.spw_range[0])/2, 'xx') #Get Q matrix for 0th delay mode
+        Q_sample = ds_Q.get_integral_beam('xx') #Get Q matrix for 0th delay mode
```
saurabh-astro: This is now independent of mode.

hera_pspec/pspecdata.py
```diff
@@ -1545,21 +1583,9 @@ def get_Q(self, mode, pol=False):
...
-            integral_beam = (np.pi/(3.0*(N)**2))* \
-                                      np.dot(beam_prod, beam_prod.T) #beam_prod has omega subsumed, but taper is still part of R matrix
-                                                                     # the nside terms is dtheta^2, where dtheta is the resolution in healpix map
-        except(AttributeError):
-            warnings.warn('The beam response could not be calculated. PS will not be normalized!')
-            integral_beam = np.ones((len(nu), len(nu)))
-
         eta_int = np.exp(-2j * np.pi * tau * nu) #exponential part of the expression
         Q_alt   = np.einsum('i,j', eta_int.conj(), eta_int) # dot it with its conjugate
-        Q       = Q_alt * integral_beam
```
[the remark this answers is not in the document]
saurabh-astro: Related to this, it would also be useful to mention the improvement in speed after this refactoring (may be at the documentation of pspec fucntion (exact_norm parameter).

philbull:
There was a merge conflict, so I fixed that and did a bit of tidy-up. @Ronan1879 don't forget to pull before making any more changes!

PR #219, opened 2019-08-28
Makes r_params backwards compatible for UVPSpec

nkern:
Any new parameter added to our file formats (including `PSpecContainer` and `UVPSpec`) must have a catch in their `read()` functions to catch for the case that the parameter is not defined.

This wasn't caught in our normal review process b/c we don't store static `UVPSpec` objects in our testing directly, and create them on-the-fly in our testing suite. This was done b/c our file format and its API was actively changing, but now that things have settled down I think its a good idea to make one simple static `UVPSpec` file in our data directory to specifically catch these backwards compatibility errors. Therefore, an example `uvp` object is added to the data dir and a simple `read()` test is added to `test_uvpspec`.

addresses #220


steven-murray:
Looks good to me!

Issue #221, opened 2019-08-29
move from aipy taper functions to uvtools taper functions


nkern:
We should move from using aipy tapering functions, which are somewhat limited, to those provided in uvtools, which provides a much larger library of tapering functions, including 7-term blackman-harris functions and all standard window functions in `scipy.signal.windows`

PR #222, opened 2019-08-29
pspec tapering functions draw from uvtools

nkern:
uvtools provides a much larger library of tapering functions than aipy, so our `PSpecData.taper` kwarg now draws from uvtools. One particular feature is the 7-term blackman harris window, which achieves more sidelobe suppression than the standard blackman harris window (aka 4-term blackman harris window). For users wanting to make power spectra of simulated FG + EoR datasets this may be necessary to achieve the required dynamic range. Addresses #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,9 @@
 [![Coverage Status](https://coveralls.io/repos/github/HERA-Team/hera_pspec/badge.svg?branch=master)](https://coveralls.io/github/HERA-Team/hera_pspec?branch=master)
 [![Documentation](https://readthedocs.org/projects/hera-pspec/badge/?version=latest)](https://readthedocs.org/projects/hera-pspec/badge/?version=latest)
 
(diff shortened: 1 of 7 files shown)


philbull:
There were a couple more instances of `aipy` tapering functions being used. I've now removed them, and now we no longer have `aipy` as a dependency!

nkern:
@philbull awesome! feel free to approve and merge whenever you get a chance

nkern:
@philbull so although we don't explicitly call `aipy` anymore, we are still reliant on packages that themselves have an `aipy` dependence, so maybe we should keep the `aipy` installation in our travis script?

philbull:
Interestingly, the tests still seem to be passing... I'll re-run to see if anything breaks.

nkern:
@philbull looks like this checks out. I'll merge it by the end of the day if you are okay with it, thanks!

PR #224, opened 2019-09-19
Fixes wedge plotting issue w/ baseline ordering

nkern:
When wedge plotting sometimes the baseline ordering is mixed up, so the wedge well doesn't really look wedge-like. Realized this happens because in `uvp.get_blpairs()` we do `np.unique()` which automatically resorts the baselines not by their length but by their id, which can mess things up. So I made `uvp.get_blpairs()` do a unique without re-sorting and also enforced this ordering anyways in the wedge plotting function.

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -395,7 +395,7 @@ class UVPSpec(object):
         blp_avg_sep = np.empty(self.Nblpairts, np.float)
 
         # construct blpair_bls
-        blpairs = _unique_ordered(self.blpair_array)
+        blpairs = _ordered_unqique(self.blpair_array)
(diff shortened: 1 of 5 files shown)


hera_pspec/uvpspec.py
```diff
@@ -2406,3 +2407,12 @@ def get_uvp_overlap(uvps, just_meta=True, verbose=True):
 
     return uvps, concat_ax, unique_spws, unique_blpts, \
            unique_polpairs, static_meta
+
+
+def _ordered_unique(arr):
+    """
+    Get the unique of an array while preserving order.
```
philbull: "unique" -> "unique elements"

PR #234, opened 2019-11-27
remove python2 support

anguta:
removed python 2 options from .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
-  - "2.7"
   - "3.6"
 
 # Cache pip-installed dependencies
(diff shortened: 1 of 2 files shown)


anguta:
I haven't found any other references to python 2 in hera_pspec (or its documentation).

anguta:
> It looks like the only other thing to change is `version.py`, for which the check for PY2 can be removed.
> 
> Also, I don't know if this is the right place for it, but the `aipy` package version specification can now be changed to `aipy>=3`.
> 
> Another slightly off-topic thing: there's apparently a README _inside_ the package?



> It looks like the only other thing to change is `version.py`, for which the check for PY2 can be removed.

Fixed this stuff. 

> 
> Also, I don't know if this is the right place for it, but the `aipy` package version specification can now be changed to `aipy>=3`.

I believe that it is >=3 as is (>=3.12)?  
> 
> Another slightly off-topic thing: there's apparently a README _inside_ the package?

The README is for another pull request.

PR #241, opened 2020-01-09
updated calls to polstr2num w/ x_orientation

nkern:
A recent `pyuvdata` update switched 'xx' and 'yy' to 'ee' and 'nn' with a definition of `x_orientation`, which H1C files now conform to. This caused some of our calls of `polstr2num` to break without propagating the `x_orientation` attribute. This PR propagates that attribute to fix this in the various regions where `polstr2num` is required.

### DIFF
--- a/hera_pspec/pspecbeam.py
+++ b/hera_pspec/pspecbeam.py
@@ -452,6 +452,7 @@ class PSpecBeamUV(PSpecBeamBase):
             Frequencies [Hz] to interpolate onto.
         x_orientation: str, optional
             Orientation in cardinal direction east or north of X dipole.
+            Default keeps polarization in X and Y basis.
 
         Returns
         -------
@@ -583,6 +584,7 @@ class PSpecBeamFromArray(PSpecBeamBase):
 
         x_orientation : str, optional
             Orientation in cardinal direction east or north of X dipole.
+            Default keeps polarization in X and Y basis.
         """
         self.OmegaP = {}; self.OmegaPP = {}
         self.x_orientation = x_orientation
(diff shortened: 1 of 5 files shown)


hera_pspec/pspecbeam.py
```diff
@@ -448,6 +448,10 @@ def beam_normalized_response(self, pol='pI', freq=None):
             'pI', 'pQ', 'pU', 'pV', 'XX', 'YY', 'XY', 'YX' 
             The output shape is (Nfreq, Npixels)
             Default: 'pI'
+        freq: array, optional
+            Frequencies [Hz] to interpolate onto.
+        x_orientation: str, optional
```
jsdillon: I'd recommend clarifying what the default None does.

hera_pspec/pspecbeam.py
```diff
@@ -576,8 +580,12 @@ def __init__(self, OmegaP, OmegaPP, beam_freqs, cosmo=None):
         cosmo : conversions.Cosmo_Conversions object, optional
             Cosmology object. Uses the default cosmology object if not 
             specified. Default: None.
+
+        x_orientation : str, optional
+            Orientation in cardinal direction east or north of X dipole.
```
jsdillon: I'd recommend clarifying what the default None does.

hera_pspec/pstokes.py
```diff
@@ -89,6 +89,9 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI'):
         Pseudo stokes polarization to form, can be 'pI' or 'pQ' or 'pU' or 'pV'.
         Default: pI
 
+    x_orientation: str, optional
```
jsdillon: I'd recommend clarifying what the default None does.

hera_pspec/uvpspec_utils.py
```diff
@@ -377,6 +377,9 @@ def polpair_tuple2int(polpair):
         A length-2 tuple containing a pair of polarization strings
         or integers, e.g. ('XX', 'YY') or (-5, -5).
 
+    x_orientation: str, optional
```
jsdillon: I'd recommend clarifying what the default None does.

PR #246, opened 2020-03-01
Upgrade grouping.average_spectra()

JianrongTan:
Upgrade grouping.average_spectra() by allowing using error bars stored in UVData.stats_array as weights when averaging power spectra. Below is the comparison on average power spectra constructing from using integration_time as weights and using analytic noise as weights.
average_ps


hera_pspec/grouping.py
```diff
@@ -153,6 +154,15 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
         does this for every specified key. Every stats_array key that is
         not specified is thrown out of the new averaged object.
 
+    if_error_weighted: bool, optional
+        If True, power spectra will be averaged in a way weighed by
+        error bars. Default: False
+
+    name_error_weights: string, optional
```
philbull: Instead of having two arguments needed to specify the weights, perhaps we could only use one? e.g. one kwarg called `error_weights` that is `None` by default (the same as setting `if_error_weighted = False`), but can be set to a string that tells the function which key of `stats_array` to use.

hera_pspec/grouping.py
```diff
@@ -153,6 +154,15 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+         name_error_weights specify which kind of errors we use for weights
```
philbull: Should also mention that the weights are defined as $w_i = 1/\sigma_i^2$, where $\sigma_i$ is taken from the relevant field of `stats_array`.

hera_pspec/grouping.py
```diff
@@ -288,74 +307,135 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
 
+                    if if_error_weighted:
+                    # If if_error_weighted==True, all arrays are weighted by a specified kind of errors.
+                    # For each power spectrum P_i with error sigma_i, 
+                    # P_avg = \sum{ P_i / (sigma_i)^2 } / \sum{ 1 / (sigma_i)^2 } 
+                    # while for terms like variance or covariance,
+                    # epsilon_avg = \sum{ (epsilon_i / (sigma_i)^4 } / ( \sum{ 1 / (sigma_i)^2 } )^2
+                    # For reference: Tegmark 1997, The Astrophysical Journal Letters, 480, L87, Table 1, #3
+                    # or Dillon 2014, Physical Review D, 89, 023002 , Equation 34. 
+                        error_weights = uvp.get_stats(name_error_weights, (spw, blp, p)).copy()
```
philbull: Perhaps these two lines can be merged into one, so `error_weights` is the right quantity immediately (then there would be no need for the `.copy()` either).

hera_pspec/grouping.py
```diff
@@ -288,74 +307,135 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                        error_weights = 1./ error_weights**2
+                        # shape of error_weights: (Ntimes, Ndlys)
+                        
+                        # Get error bar
+                        errws = {}
+                        for stat in stat_l:
+                            errs = uvp.get_stats(stat, (spw, blp, p)).copy()
+                            # shape of errs: (Ntimes, Ndlys)
+                            errws[stat] = errs**2
```
philbull: These two lines can also be merged into one, without a `.copy()`

hera_pspec/grouping.py
```diff
@@ -288,74 +307,135 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                            wgts = (np.sum(wgts * error_weights[:, :, None], axis=0) \
+                                / np.sum(error_weights, axis=0)[:, None])[None]
+                            ints = (np.sum(ints * error_weights, axis=0) \
+                                / np.sum(error_weights, axis=0))[None]
+                            nsmp = np.sum(nsmp, axis=0)[None]
+                            if store_cov:
+                                cov = (np.sum(cov * error_weights[:, :, None]**2, axis=0) \
+                                    / (np.sum(w,axis=0))[:, None]**2)[None]
+                            for stat in stat_l:
+                                errws[stat] = (np.sum(errws[stat]*error_weights**2, axis=0) \
```
philbull: I think this is doing a weighted sum (in quadrature) of all of the error fields, where the weights are from the field specified by `name_error_weights` rather than each error field itself. Perhaps this should be mentioned in the docstring at the top, as it's slightly non-trivial.

hera_pspec/grouping.py
```diff
@@ -288,74 +307,135 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                            for stat in stat_l:
+                                bpg_stats[stat].append(errws[stat]*error_weights**2)
+                            w_list.append(error_weights)
+                            if store_cov:
+                                bpg_cov.append(cov * error_weights[:, :, None]**2)
+
+                    else:
+                    # Otherwise all arrays are averaged in a way weighted by the integration time
+                        w = (ints * np.sqrt(nsmp))
+                        # shaoe of w: (Ntimes, 1)
```
philbull: shaoe -> shape

hera_pspec/grouping.py
```diff
@@ -288,74 +307,135 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                        # to the weighting/multiplicity
+                        for m in range(int(blpg_wgts[k])):
+                            bpg_data.append(data * w)
+                            bpg_wgts.append(wgts * w[:, None])
+                            bpg_ints.append(ints * w)
+                            bpg_nsmp.append(nsmp)
+                            [bpg_stats[stat].append(errws[stat]) for stat in stat_l]
+                            w_list.append(w)
+                            if store_cov:
+                                bpg_cov.append(cov * w[:, None]**2)
```
philbull: Are these two blocks of code in the `if` statement almost the same, except that the first block uses the specified `error_weights` and the second uses the integration time as the weights? 

If (and only if) possible, it would be good to just have one block of code that does the averaging/time averaging/multiplicity (to avoid having two copies of very similar code in both parts of the `if` statement). The weights could be calculated in a shorter `if` statement above this block.

philbull:
P.S. Would it be worth adding a new example notebook showing how the averaging works? I think we already have some averaging examples in another notebook, but we could add another that shows a plot similar to the one @JianrongTan posted above, amongst other things. (It could also include an example of how to use the multiplicity feature perhaps?)

hera_pspec/grouping.py
```diff
@@ -153,6 +154,15 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
         does this for every specified key. Every stats_array key that is
         not specified is thrown out of the new averaged object.
 
+    if_error_weighted: bool, optional
```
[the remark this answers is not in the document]
JianrongTan: Sure.

hera_pspec/grouping.py
JianrongTan: Sure.

hera_pspec/grouping.py
JianrongTan: Sure.

hera_pspec/grouping.py
JianrongTan: Sure.

hera_pspec/grouping.py
JianrongTan: They are not the same on how to average the error fields. In the original code, each error field is weighted by itself (1/sigma_i^2), while other arrays are weighted by the integration time; In the new code, every array is weighted by the specific error. I hope we could have more clarification here.

JianrongTan:
Thanks @philbull! I have already had a notebook to show how to use the new feature and plot the figure above. If possible, I could easily include that notebook in my pull request.

hera_pspec/grouping.py
JianrongTan: Sure.

hera_pspec/grouping.py
JianrongTan: Sure.

hera_pspec/grouping.py
philbull: OK, if it can't be simplified, then that's fine. (May be worth mentioning in an additional in-line comment though.)

philbull:
Sounds good! Please do add the notebook; it doesn't have to be anything complicated, just a simple demonstration of how to use the different modes.

hera_pspec/grouping.py
JianrongTan: From the discussion on today's pspec telcon, people agree us to do the average in a consistent way in which every array is either weighted by the integration time or the error bars.

hera_pspec/grouping.py
```diff
@@ -153,6 +154,14 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
         does this for every specified key. Every stats_array key that is
         not specified is thrown out of the new averaged object.
 
+    error_weights: string, optional
+         error_weights specify which kind of errors we use for weights 
+         during averaging power spectra.
+         The weights are defined as $w_i = 1/ sigma_i^2$, 
+         where $sigma_i$ is taken from the relevant field of stats_array.
+         If `error_weight' is set to None, which means we 
+         just use the integration time as weights. Default: None
```
nkern: should add that if `error_weights` is specified then it also gets assigned as `error_field`, and if one specified both then `error_weights` supercedes as `error_field`. lmk if that makes sense.

hera_pspec/grouping.py
```diff
@@ -288,74 +308,88 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                        errws[stat] = (uvp.get_stats(stat, (spw, blp, p)))**2
+                        # shape of errs: (Ntimes, Ndlys)
+
+                    if use_error_weights:
+                    # If use_error_weights==True, all arrays are weighted by a specified kind of errors,
+                    # including the error_filed in stats_array and cov_array. 
+                    # For each power spectrum P_i with error sigma_i, 
+                    # P_avg = \sum{ P_i / (sigma_i)^2 } / \sum{ 1 / (sigma_i)^2 } 
+                    # while for terms like variance or covariance,
+                    # epsilon_avg = \sum{ (epsilon_i / (sigma_i)^4 } / ( \sum{ 1 / (sigma_i)^2 } )^2
```
nkern: is epsilon a standard dev or variance? should be a variance, no? so perhaps Sigma is more appropriate for this comment?

hera_pspec/grouping.py
```diff
@@ -288,74 +308,88 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
-                           / np.sum(w, axis=0).clip(1e-10, np.inf))[None]
-                      nsmp = np.sum(nsmp, axis=0)[None]
-                      w = np.sum(w, axis=0)[None]
-                      if store_cov:
-                          cov = (np.sum(cov * w[:, None], axis=0) \
-                          / np.sum(w,axis=0).clip(1e-10, np.inf))[None]
-
-                      for stat in stat_l:
-                        errws[stat] = np.sum(errws[stat], axis=0)[None]
+                        data = (np.sum(data * w, axis=0) \
```
nkern: why are the `.clip` methods being removed? they are important for weights that are all zeros (which can happen)

hera_pspec/grouping.py
```diff
@@ -288,74 +308,88 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                        nsmp = np.sum(nsmp, axis=0)[None]
+                        if store_cov:
+                            cov = (np.sum(cov * w[:, :, None]**2, axis=0) \
+                                / (np.sum(w,axis=0))[:, None]**2)[None]
+                        for stat in stat_l:
+                            errws[stat] = (np.sum(errws[stat]*w**2, axis=0) \
+                            / (np.sum(w, axis=0))**2)[None]
+                        w = np.sum(w, axis=0)[None]
                     # Add multiple copies of data for each baseline according
                     # to the weighting/multiplicity
```
nkern: should add this note for new developers
`: multiple copies are only added when bootstrap resampling`

hera_pspec/grouping.py
```diff
@@ -288,74 +308,88 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
-                            bpg_cov.append(cov * w[:, None])
+                            bpg_cov.append(cov * w[:, :, None]**2)
 
+                # Average over baseline-pairs
                 # Take integration-weighted averages, with clipping to deal
                 # with zeros
+                # Or take error_weighted averages
                 bpg_data = np.sum(bpg_data, axis=0) \
-                         / np.sum(w_list, axis=0).clip(1e-10, np.inf)
+                         / np.sum(w_list, axis=0)
```
nkern: same question here on removal of `clip` methods

hera_pspec/tests/test_grouping.py
```diff
@@ -76,6 +76,54 @@ def test_group_baselines(self):
...
+            blp = uvp.antnums_to_blpair(key[1])
+            error = Pn[blp]
+            uvp.set_stats("noise", key, error)
+        blpair_groups = [[((24, 25), (37, 38)),((24, 25), (38, 39)), ((37, 38), (38, 39))]]
+        uvp_avg_ints_wgts = grouping.average_spectra(uvp, blpair_groups=blpair_groups,
+                                                error_field="noise", time_avg=True,inplace=False)
+        uvp_avg_error_wgts = grouping.average_spectra(uvp, time_avg=True, blpair_groups=blpair_groups,
+                                                 error_field="noise", error_weights="noise",
+                                                 inplace=False)
+        assert(abs(uvp_avg_ints_wgts.stats_array["noise"][0][0,0,0]) < abs(uvp.stats_array["noise"][0][0,0,0]))
```
nkern: can you add a comment here on what this is testing and to what tolerance?

hera_pspec/grouping.py
```diff
@@ -234,6 +243,17 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
             if stat not in uvp.stats_array.keys():
                 raise KeyError("error_field \"%s\" not found in stats_array keys." % stat)
 
+    # pre-check for choos average weighted by error bars
```
nkern: awkard sentence

hera_pspec/grouping.py
```diff
@@ -234,6 +243,17 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+    
+    if error_weights is None:
```
nkern: this block should be moved up 20 lines, before the creation of `stat_l`

hera_pspec/grouping.py
```diff
@@ -288,74 +308,88 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                    # while for terms like variance or covariance,
+                    # epsilon_avg = \sum{ (epsilon_i / (sigma_i)^4 } / ( \sum{ 1 / (sigma_i)^2 } )^2
+                    # For reference: M. Tegmark 1997, The Astrophysical Journal Letters, 480, L87, Table 1, #3
+                    # or J. Dillon 2014, Physical Review D, 89, 023002 , Equation 34. 
+                        w = 1./(uvp.get_stats(error_weights, (spw, blp, p)))**2
+                        # shape of w: (Ntimes, Ndlys)
+                    else:
+                    # Otherwise all arrays are averaged in a way weighted by the integration time,
+                    # including the error_filed in stats_array and cov_array. 
+                        w = (ints * np.sqrt(nsmp))
```
nkern: If we are inverse variance weighting I think this should be `w = (ints * np.sqrt(nsmp))**2`

nkern:
@JianrongTan thanks for making this PR. I want to clarify that if you are inverse variance weighting the power spectrum average, than I think you actually need to weight by `t_int^2` in the case of integration-time weighting, see below:
```
sigma_n = P_N ~ Tsys^2 / t_int (see eqn 12 of Kern+2020b)
therefore
w = 1/sigma_n^2 ~ t_int^2
```
however, correct me if this is wrong

hera_pspec/grouping.py
JianrongTan: My understanding is that the elements in error_field include error_weights, which allow us to also weight and average other error bars with the specified one.

hera_pspec/grouping.py
JianrongTan: epsilon is the variance or covariance. Since here we only use sigma as weights to average all stats_array and cov_array, so I name the other variance and covariance terms as epsilon to avoid confusion.

hera_pspec/grouping.py
JianrongTan: I remove the line since I found a cut-off at 1e-10 is not appropriate when we use inverse variance as weights, for they could be as small as 1e-20.

hera_pspec/tests/test_grouping.py
JianrongTan: The intuition here is the error bar on the average result should be smaller than on one sample.

hera_pspec/grouping.py
JianrongTan: It is much better in this way!

hera_pspec/grouping.py
JianrongTan: I agree.

hera_pspec/grouping.py
nkern: I see good point!

hera_pspec/grouping.py
nkern: Okay, then let's keep the `clip` methods and just change the tolerance to `1e-20`.

hera_pspec/grouping.py
JianrongTan: And nsmp here refers to N_coherent? So maybe we should have  w = (ints * nsmp)**2

nkern: actually, Nsmp is Nincoherent, so it is correct the way it is I believe

JianrongTan: Yes, I have checked that and followed w = (ints * np.sqrt(nsmp))**2.

hera_pspec/tests/test_grouping.py
nkern: I see, given that we have known input, can we assert that the result is equal to an analytic result? In this case, we are averaging 3 blpairs, so the averaged noise should be 1/sqrt(N) times the single sample errorbar, no?

JianrongTan: Since the noise power spectrum for each single baseline-pair power spectrum are not equal,  the average result will not be the exactly 1/sqrt(N) times the single sample errorbar. If we want to be confirmed on this, I did a similar thing in  test_stats_array, where all the error bars are set to be one, and I have a check like  nt.assert_true(np.all(np.isclose(u.get_stats("errors", keys[0])[0], np.ones(u.Ndlys)/np.sqrt(len(blpairs))))).

hera_pspec/grouping.py
JianrongTan: To be more cautious, I set the tolerance as 1e-40.

hera_pspec/grouping.py
```diff
@@ -153,6 +154,17 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
         does this for every specified key. Every stats_array key that is
         not specified is thrown out of the new averaged object.
 
+    error_weights: string, optional
+         error_weights specify which kind of errors we use for weights 
+         during averaging power spectra.
+         The weights are defined as $w_i = 1/ sigma_i^2$, 
+         where $sigma_i$ is taken from the relevant field of stats_array.
+         If `error_weight' is set to None, which means we just use the 
+         integration time as weights. If error_weights is specified,
```
nkern: Actually the way you have implemented, this should read
```
If error_weights is specified then it gets appended to error_field as a list.
```

hera_pspec/tests/test_grouping.py
nkern: let's add that. once that's in I'm all set with this PR!

nkern:
@JianrongTan thanks for the edits. two things
1. let's add the simple errorbar average test you described (setting them to 1 and making sure the final errorbar == 1/sqrt(Nblps)
2. docstring edit 

then I'm good w/ this PR 👍

JianrongTan:
@philbull @nkern Thanks for all your comments! I have added the parts as you suggested. Also, in the previous commits I forgot to update uvpspec.average_spectra() the same as  grouping.average_spectra(). Now the most up-to-date PR includes all the changes.

PR #247, opened 2020-03-17
extra_info kwarg to utils.calc_blpair_reds

nkern:
Brief PR: enables extra redundant group output from `utils.calc_blpair_reds` if requested. Usable by pspec pipe for redundant group handling.

To not break existing API, I made this output optional (default False)

### DIFF
--- a/hera_pspec/utils.py
+++ b/hera_pspec/utils.py
@@ -972,6 +972,7 @@ def job_monitor(run_func, iterator, action_name, M=map, lf=None, maxiter=1,
 
     return failures
 
+
 def get_bl_lens_angs(blvecs, bl_error_tol=1.0):
     """
     Given a list of baseline vectors in ENU (TOPO) coords, get the
--- a/hera_pspec/utils.py
+++ b/hera_pspec/utils.py
@@ -240,7 +240,8 @@ def calc_blpair_reds(uvd1, uvd2, bl_tol=1.0, filter_blpairs=True,
         to keep in baseline selection
 
     extra_info : bool, optional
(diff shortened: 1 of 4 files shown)


hera_pspec/utils.py
```diff
@@ -246,6 +253,16 @@ def calc_blpair_reds(uvd1, uvd2, bl_tol=1.0, filter_blpairs=True,
 
     xants1, xants2 : lists
         List of bad antenna integers for uvd1 and uvd2
+
+    if extra_info:
```
philbull: I think this might break the ReadTheDocs formatting.

hera_pspec/utils.py
```diff
@@ -235,6 +239,9 @@ def calc_blpair_reds(uvd1, uvd2, bl_tol=1.0, filter_blpairs=True,
         len-2 tuple containing (minimum, maximum) baseline angle in degrees
         to keep in baseline selection
 
+    extra_info : bool, optional
+        If True, return extra info on redundant group length and angle
```
philbull: Could be slightly more specific.

hera_pspec/utils.py
```diff
@@ -327,6 +349,8 @@ def calc_blpair_reds(uvd1, uvd2, bl_tol=1.0, filter_blpairs=True,
             bls1, bls2 = _bls1, _bls2
             blps = list(zip(bls1, bls2))
 
+        rinds = [j] * len(blps)
```
philbull: Perhaps an inline comment here just to explain what this is keeping track of?

PR #251, opened 2020-03-30
spherical average spectra and window function

JianrongTan:
We allow the output for spherical average spectra and window function in this PR (run over all baseline-pairs and integrations and call the error_weighted method in average_spectra()). Here are examples of final outputs:
window_function k-k: 
sph_window
1D power spectrum
sph_ps


nkern:
Thanks for the work @JianrongTan -- this is a much-needed advancement. I have one high level comment that I'd appreciate hearing from you and @acliu and others on

 Your `cylindrical_spherical_average` function outputs dictionaries of power spectra, but do we instead want to output `UVPSpec` objects? One of the key features of using uvpspec is our ability to write them to disk in a standard format with associated metadata. For the cylindrical average you should actually able to output it in `UVPSpec` with existing infrastructure, but the spherical average is more complicated and will require some deep infrastructure changes in the `UVPSpec` object (I've taken a brief look at doing this). The answer I think comes down to the time vs. benefit ratio: is the benefit of keeping spherically averaged power spectra in UVPSpec format worth the time investment of UVPSpec plumbing? Note that the benefit of having data in UVPSpec is not just easy read-write ability and its self-consistent handling of metadata, but also automatic interfacing with our plotting functions, folding functions, and P->DelSq functions.

JianrongTan:
Thanks @nkern! Yes it is the reason I am hesitated to put spherical outputs as intrinsic attributes of UVPspec. The other arrays inside UVPspec share similar shape related to (Nblpairts,Ndlys, Npols), while the spherical averaged ones has definitely different data shape from them, so I wonder if it is worth doing more dumb works here just for changing infrastructures, and I prefer just keeping all array shapes in UVPspec consistent and intuitive.

nkern:
@JianrongTan Yeah but it might be worth thinking about how much work is actually needed to do it properly--because we will eventually want to do this in uvpspec, so its more of just putting if off than not doing it entirely.

JianrongTan:
[Mapping_cover_array_and_window_function_array_when_running_average.pdf](https://github.com/HERA-Team/hera_pspec/files/4405347/Mapping_cover_array_and_window_function_array_when_running_average.pdf)
I feel the average on cov_array and window_function_array is not as trivial as on other arrays. I attach a memo here for the math I used for averaging cov_arrays and window_function_arrays. Hope reviewers can also check this with the corresponding code.

nkern:
@JianrongTan @acliu okay revamp of `grouping.spherical_average` is done I think. I made the notation and the code conform to Dillon+2013 eqn. 33 - 38 standards and all seems to check out. I added a number of unit tests, such as ensuring that the row-sum of W_spherical == 1.0 for a number of different weighting options (uniform weight, covariance diagonal weight and full covariance weigh), and ensuring that the resultant errors, C_sph^1/2, average down as expected.

@acliu This seems to have fixed the problem I was having before, where if I gave infinite variance to certain delay modes then the resultant W_sph did not row-sum to unity: this is true now.

@JianrongTan I removed your `cylindrical_and_spherical_average` function, as it is now not needed.

Please let me now if you have have any comments on the added code here!

hera_pspec/grouping.py
```diff
@@ -496,13 +511,312 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+    # perform time and cylindrical averaging upfront if requested
+    if blpair_groups is not None or time_avg:
+        uvp.average_spectra(blpair_groups=blpair_groups, time_avg=time_avg,
+                            blpair_weights=blpair_weights, error_weights=error_weights,
+                            inplace=True)
+
+    # initialize blank arrays and dicts
+    Nk = len(kbins)
+    dlys_array, spw_dlys_array = [], []
+    (data_array, window_function_array, wgt_array, integration_array,
```
JianrongTan, line 614: It is maybe better to define window_function in a `if store_window:' statement.

hera_pspec/grouping.py
```diff
@@ -496,13 +511,312 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
+                for stat in stats_array:
+                    stats_array[stat][spw][:, dslice] = uvp.stats_array[stat][spw][blpt_inds]
+            if store_cov:
+                cov_array[spw][:, dslice, dslice] = uvp.cov_array[spw][blpt_inds]
+
+            # fill weighting matrix E
+            if weight_by_cov:
+                # weight by inverse (real) covariance
+                for p in range(uvp.Npols):
+                    E[:, dslice, dslice, p] = np.linalg.pinv(cov_array[spw][:, dslice, dslice, p].real)
```
JianrongTan, line 706: I think here we could add some notations to emphasize the inverse covariance matrix C^{-1} used as weights here, has a shape of (N_cyl, Ncyl). Because we only have the covariance information for bandpowers between different delays on the same baseline-pair, our C here is block diagonal, so C^{-1} is equal to the direct sum of the inverse of each block matrix.

nkern: sure thing

nkern:
@JianrongTan give me a verbal go-ahead to merge.

PR #255, opened 2020-04-21
revert test_uvp to 8/2019. made window_func non-req

nkern:
This reverts the `test_uvp.h5` file back to its form in 8/2019, and makes the `window_function_array` attribute of `UVPSpec` a non-required parameter. 

This also addresses #256 : I was getting weird H matrices b/c I was not setting Ndlys after setting spw_range. We now automatically set Ndlys upon a `ds.set_spw()` call so this does not happen.

Another thing I'm finding in adding these tests is a failure in `get_MW` with `mode='V^-1/2'`. The eigenvector decomposition performed leads to some small but negative eigenvalues. There is a warning put in when this happens, but it doesn't do anything about the negative values--and the result is that `1/sqrt(eigvals)` creates nans that then get propagated to the M and W matrices. I fix this by truncating these eigenvectors before taking `1/sqrt(eigvals)`.

### DIFF
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -875,17 +875,17 @@ def fold_spectra(uvp):
             uvp.data_array[spw][:, Ndlys//2+1:, :] = np.mean([left, right], axis=0)
             uvp.data_array[spw][:, :Ndlys//2, :] = 0.0
             uvp.nsample_array[spw] *= 2.0
-            if store_window
-            leftleft = uvp.window_function_array[spw][:, 1:Ndlys//2, 1:Ndlys//2, :][:, ::-1, ::-1, :]
-            leftright = uvp.window_function_array[spw][:, 1:Ndlys//2, Ndlys//2+1:, :][:, ::-1, :, :]
-            rightleft = uvp.window_function_array[spw][:, Ndlys//2+1: , 1:Ndlys//2, :][:, :, ::-1, :]
-            rightright = uvp.window_function_array[spw][:, Ndlys//2+1:, Ndlys//2+1:, :]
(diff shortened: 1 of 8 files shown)


nkern:
Another (minor) thing this does is clean up the `ds._R` key caching syntax. The way the keys were being stored was not consistent with the `ds.set_R` docstring, such that one could not easily figure out how to actually store arbitrary R matrices without having to hack the `ds._R` dictionary directly.

@anguta I think this was because of the addition of dayenu keywords to the `Rkey` parameter in `ds.R()` method, even when dayenu wasn't being invoked. I made it such that the dayenu params are only appended to `Rkey` if dayenu is enabled, and then made sure this was reflected in the `ds.set_R` docstring. In the future if you add additional caching features it would be good to keep this in mind--I imagine your extra caching PR also open would benefit from this :)

hera_pspec/grouping.py
```diff
@@ -412,21 +423,26 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     arr = np.sum(bpg_stats[stat], axis=0) \
                         / (np.sum(w_list, axis=0).clip(1e-40, np.inf)**2)
                     bpg_stats[stat] = np.sqrt(arr)
-                bpg_window_function = np.sum(bpg_window_function, axis=0) \
-                            / (np.sum(w_list, axis=0).clip(1e-40, np.inf)[:,:,None])
+                    if store_window:
```
JianrongTan, line 426: if store_window has wrong indentation.

nkern: great catch thanks!

nkern:
@acliu @philbull I'm going to merge this, but maybe it would be good to spend a few minutes next pspec telecon to share some of the stuff in here for major pspec developers so we are all on the same page.

Issue #260, opened 2020-05-01
combine stokes bug


anguta:
I suspect that `construct_pstokes` may not be doing what it is supposed to 

In the following lines 272 to 273 of `pstokes.py`

a pstokes is formed from `req_pol1` in `uvd2` and `req_pol1` in `uvd2`. This is because the last select line selects `req_pol1` in `uvd2` instead of `req_pol2`. 

```
assert req_pol1 in uvd1.polarization_array, \
        "Polarization {} not found in dset1 object".format(req_pol1)
    if uvd1.Npols > 1:
        uvd1 = uvd1.select(polarizations=req_pol1, inplace=False)
    assert req_pol2 in uvd2.polarization_array, \
        "Polarization {} not found in dset2 object".format(req_pol2)
    if uvd2.Npols > 1:
        uvd2 = uvd2.select(polarizations=req_pol1, inplace=False)
```

`req_pol1` should be changed to `req_pol2` in the last line.

nkern:
@anguta can we make this fix a separate PR, just so we can avoid a big pileup of features into one massive PR? I also have a tendency to push a lot of features into one PR, but this bug seems truly orthogonal to the other stuff, so perhaps its a good candidate for splitting out? Also Ridhima should review this bug fix, and I don't think we want to make her fish it out of everything else we've got going on in your other PR ;)

anguta:
That makes sense. I actually started running into bugs in my bug fix so rather then let things snowball, I've restored the bug in #236 and will make a new PR to deal with it once this gets merged.

nkern:
okay great--i'd like to get this bug fixed asap as it currently impacts our bottom line. this bug probably should not have made it through our unit tests, so I'll make sure I add a test to catch this. @Chuneeta I'll have a PR for you shortly

PR #262, opened 2020-05-11
fixed pstokes multipol files construct_pstokes bug

nkern:
There was a pretty serious bug in `pstokes.construct_pstokes` where the right-hand polarization was not being selected properly from the input UVData object, see here #260.

Turns out this was only a problem when operating on multi-polarization files, which, while the code supported, was not being tested on, as at the time we used 1-pol files.
This PR fixes that bug and adds tests for the `construct_pstokes` method on our multi-pol files to catch this kind of bug going forward. Thanks @anguta for spotting this!


nkern:
@Chuneeta thanks! @anguta let me know if you have any comments on this, seeing as you raised the issue originally.

PR #264, opened 2020-05-12
fixed stats_array propagation bug in combine_uvpspec

nkern:
In `uvpspec.combine_uvpspec`, the optional attribute `stats_array` was not being propagated, even if it was filled by all input UVPSpec objects. This fixes that, and also asserts that for propagation of cov_array, all cov_model attributes of incoming uvps must match. Fixes #264

### DIFF
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -338,7 +338,8 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     # Get squared statistic
                     errws = {}
                     for stat in stat_l:
-                        errws[stat] = (uvp.get_stats(stat, (spw, blp, p)))**2
+                        errws[stat] = uvp.get_stats(error_weights, (spw, blp, p))
+                        np.square(errws[stat], out=errws[stat], where=np.isfinite(errws[stat]))
                         # shape of errs: (Ntimes, Ndlys)
 
                     if use_error_weights:
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -338,7 +338,7 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     # Get squared statistic
                     errws = {}
                     for stat in stat_l:
-                        errws[stat] = uvp.get_stats(stat, (spw, blp, p))
+                        errws[stat] = uvp.get_stats(stat, (spw, blp, p)).copy()
                         np.square(errws[stat], out=errws[stat], where=np.isfinite(errws[stat]))
                         # shape of errs: (Ntimes, Ndlys)
 
@@ -351,7 +351,7 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     # epsilon_avg = \sum{ (epsilon_i / (sigma_i)^4 } / ( \sum{ 1 / (sigma_i)^2 } )^2
                     # For reference: M. Tegmark 1997, The Astrophysical Journal Letters, 480, L87, Table 1, #3
                     # or J. Dillon 2014, Physical Review D, 89, 023002 , Equation 34.
-                        stat_val = uvp.get_stats(error_weights, (spw, blp, p))
+                        stat_val = uvp.get_stats(error_weights, (spw, blp, p)).copy()
                         np.square(stat_val, out=stat_val, where=np.isfinite(stat_val))
                         w = np.real(1. / stat_val.clip(1e-40, np.inf))
                         # shape of w: (Ntimes, Ndlys)
@@ -732,7 +732,7 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
 
             elif error_weights is not None:
                 # fill diagonal with by 1/stats_array^2 as weight
-                stat_weight = stats_array[error_weights][spw][:, dslice].real
+                stat_weight = stats_array[error_weights][spw][:, dslice].real.copy()
                 np.square(stat_weight, out=stat_weight, where=np.isfinite(stat_weight))
                 E[:, range(dstart, dstop), range(dstart, dstop)] = 1 / stat_weight.clip(1e-40, np.inf)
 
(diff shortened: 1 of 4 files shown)


hera_pspec/uvpspec.py
```diff
@@ -2088,9 +2089,23 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
...
+    if store_stats:
+        # get shared stats keys
+        stored_stats = [set(uvp.stats_array.keys()) for uvp in uvps]
+        for i in range(len(stored_stats)):
+            stored_stats[0].intersection_update(stored_stats[i])
+        stored_stats = stored_stats[0]
+        # if nothing in common, set to False
+        if len(stored_stats) == 0:
+            store_stats = False
+        u.stats_array = odict([(stat, odict()) for stat in stored_stats])
```
JianrongTan, line 2107: Maybe better to add an `else' block here to avoid constructing a redundant u.stats_array when len(stored_stats) == 0

hera_pspec/tests/test_uvpspec.py
```diff
@@ -532,48 +542,41 @@ def test_combine_uvpspec(self):
...
-        uvp2 = copy.deepcopy(uvp1)
-        uvp2.polpair_array[0] = 1414
-        uvp2 = uvpspec.combine_uvpspec([uvp1, uvp2], verbose=False)
-        nt.assert_raises(AssertionError, uvpspec.combine_uvpspec, [uvp1, uvp2])
+        # optionals
+        for spw in out.spw_array:
+            ndlys = out.get_spw_ranges(spw)[0][-1]
+            assert out.cov_array[spw].shape == (30, ndlys, ndlys, 2)
+            assert out.stats_array['noise_err'][spw].shape == (30, ndlys, 2)
+            assert out.window_function_array[spw].shape == (30, ndlys, ndlys, 2)
```
JianrongTan, line 550: Should we check the cov_model is also as expected?

hera_pspec/tests/test_uvpspec.py
```diff
@@ -615,6 +620,50 @@ def test_combine_uvpspec(self):
         out = uvpspec.combine_uvpspec([uvp_a, uvp_b], merge_history=False, verbose=False)
         assert 'batwing' in out.history and not 'foobar' in out.history
```
JianrongTan, line 621: Do we need another test for uvps have different cov_array and stats_array keys so that the combined uvp doesn't have  such two arrays?

hera_pspec/uvpspec.py
nkern: ah, good catch

hera_pspec/tests/test_uvpspec.py
nkern: sure thing

hera_pspec/tests/test_uvpspec.py
nkern: Hmm, not quite sure I know what you mean. cov_array is an ndarray while `stats_array` is a dictionary, so it is expected that they can in principle hold different kinds of information--even if we are using stats_array as the sqrt(diagonal(cov_array)) in practice. Can you give me an example?

JianrongTan: Yes. I mean for uvp1 we have a cov_model 'model1', while for uvp2 we have a cov_model 'model2', thus their combined uvp should not have the attribute cov_array as required by the code.

nkern: Ah, I see what you are saying. Yes, that is probably a good test! thanks

nkern:
@JianrongTan great thanks. I pushed an additional small edit to fix a minor bug, which was when we were averaging stats_array that were fully flagged (i.e. stats=inf), the resultant stat was 0 (because the weights were 0 = 1/stat = 1/inf), but of course the average of inf stats should also be inf. Lmk if you have any comments on this, thanks

hera_pspec/grouping.py
```diff
@@ -363,29 +365,33 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
...
                         for stat in stat_l:
-                            # clip errws because it may be inf, and inf * 0 is nan
-                            errws[stat] = (np.sum(errws[stat].clip(0, 1e40) * w**2, axis=0) \
-                                        / (np.sum(w, axis=0).clip(1e-40, np.inf))**2)[None]
+                            # clip errws to eliminate nan: inf * 0 yields nans
+                            weighted_errws = errws[stat].clip(0, 1e40) * w**2
+                            errws[stat] = (np.sum(weighted_errws, axis=0) \
+                                           / wsum**2)[None]
+                            # set near-zero errws to inf, as they should be
+                            errws[stat][np.isclose(errws[stat], 0)] = np.inf
```
JianrongTan, line 392: If we have zero input for errws[stat] , then weighted_errws = errws[stat] * w**2 also goes to zero, but we could not set it to be np.inf here. Maybe we should just apply a cut (1e-40, 1e40) on both the lower end and higher end of weights w, in this way 1e-40*np.inf = inf and 1e40*0=0 leads to no bugs too. Since we have cuts on w on both ends, we don't need to apply cuts on other attributes wsum and weighted_errws here.

nkern: Yeah so what you suggest is effectively what we do in computing the weights abovev: we clip the errws at (0, 1e-40). However, the errorbars are intrinsically allowed to be inf (e.g. for data w/ zero integration), so when we exit the errorbar calculation (as done here) we are setting errorbars that should be inf to inf (ones with zero integration), which is needed because with zero weights but clipped errws, the errorbars would otherwise be 0, which is incorrect.

JianrongTan: Yes, I mean there are two cases lead to a weighted errws[stat] to be 0, while the code  errws[stat][np.isclose(errws[stat], 0)] = np.inf seems to set both of them back to np.inf. Imagine errws[stat] to be 0 and np.inf, the clipped weights w is 1e40 and 0; thus weighted_errws = errws[stat].clip(0, 1e40) \* w\^2 is 0\*1e40=0 and 1e40\*0 = 0; wsum is 1e40 and 1e-40; errws[stat] = (np.sum(weighted_errws, axis=0) \
                                            / wsum^2) is 0/1e80 and 0/1e-80 = 0.

nkern: Ah yes, but we would never expect `errws` to be 0 intrinsically. This is an unachievable scenario (t_int=>infinity) that is also somewhat pathological, as errws that are zero will lead to an inf weight for that data point, which mean the final average is informed solely from that data point. So I think its not an issue..

JianrongTan:
Thanks Nick! I think it's ok to merge.

PR #266, opened 2020-05-28
pspec_run blpair match typo

nkern:
a small typo that causes conjugated baseline to not be caught in pspec_run(). This is not a problem w/ any HERA data b/c baselines follow "low_hi" convention, but for simulated RIMEz data that do not conform to this, we found this bug. A unit test is added to ensure this is caught.


bhazelton:
You might look at using UVData's `conjugate_bls` method instead. It will make an object conform to a particular conjugation convention (only conjugating baselines that do not conform). And it properly updates all the metadata as well as conjugating the visibilities.

PR #267, opened 2020-06-05
Allow QE formalism to produce foreground/systematics dependent error bars

JianrongTan:
The foreground/systematics coupling term with noise in the analytic covariance outputs takes the form of
Screen Shot 2020-06-04 at 10 52 08 PM,
or
Screen Shot 2020-06-04 at 10 53 28 PM,
where C_total is the outer product of input visibilities like C_{total,ij}^{11} = x^1_i x_j^{1*} while C_n is the noise covariance estimated from auto-correlation amplitudes. I find such analytic covariance output is roughly equivalent to power spectrum product
Screen Shot 2020-06-04 at 11 00 32 PM, where P_n^2 is the estimated noise power spectrum variance, P_{dd} is the cross-baseline power spectrum directly constructed from visibilities. 

Full maths are provided in a PDF in `examples/internal_memos/errorbar_choices`, and in Tan et al. 2020 in prep.


hera_pspec/pspecdata.py
```diff
@@ -662,6 +666,16 @@ def C_model(self, key, model='empirical', time_index=None, known_cov=None):
             # add model to key
             Ckey = ((dset, dset), (bl,bl), ) + (model, time_index, False, True,)
 
+        elif model == 'outer_product':
```
nkern, line 669: It seems like the last three conditionals can be merged into a single statement: `elif model in ['autos', 'outer_product', 'foreground_dependent']:` unless you suspect that in the future they may break apart in their Ckey definition

hera_pspec/pspecdata.py
```diff
@@ -670,22 +684,30 @@ def C_model(self, key, model='empirical', time_index=None, known_cov=None):
                 # add model to key
                 Ckey = ((dset, dset), (bl,bl), ) + (model, time_index, False, True,)
 
+        # Update self._C with known_cov
+        if known_cov is not None:
+            spw = slice(self.spw_range[0]-self.filter_extension[0], self.spw_range[1]+self.filter_extension[1])
```
nkern, line 689: why is filter extension needed here? I think this is a dayenu specific attribute, and should probably not be used unless dayenu filtering is specifically requested. Or if by default filter_extension is set to [0, 0]. please clarify for me, thanks

nkern:
@JianrongTan can you clarify the new `cov_model` settings of `outer_product` and `foreground_dependent`? Also, if it isn't too much work, I think there should be a memo detailing all of the new math in `get_analytic_covariance` method. Although some of this math will be in your paper, it may be good to have a separate memo that we can date and then continue to update if the math changes. This could be a HERA memo, or just an internal hera_pspec memo we keep in the repo. If its easy enough to put that math into a separate latex, that would be great.

hera_pspec/pspecdata.py
```diff
@@ -1978,6 +2026,8 @@ def get_analytic_covariance(self, key1, key2, M=None, exact_norm=False, pol=Fals
         # Collect all the relevant pieces
 
         C11, C22, C21, C12, P11, S11, P22, S22, P21, S21 = [], [], [], [], [], [], [], [], [], []
+        if model == 'foreground_dependent':
```
nkern, line 2029: you should describe in the method docstring what `foreground_dependent` and `outer_product` do

hera_pspec/tests/test_pspecdata.py
```diff
@@ -1585,6 +1585,14 @@ def test_get_analytic_covariance(self):
...
             nt.assert_true((abs(cov_q_imag[time_index].real) / abs\
                 (cov_q_imag[time_index].imag) > 1e8).all())
+        # test foreground_dependent error bar estimation
+        uvp_autos = ds.pspec(bls1[:2], bls2[:2], (0, 1), ('xx','xx'), spw_ranges=(100,120), store_cov=True, cov_model='autos', verbose=False)
+        uvp_foreground_dependent = ds.pspec(bls1[:2], bls2[:2], (0, 1), ('xx','xx'), spw_ranges=(100,120), store_cov=True, cov_model='foreground_dependent', verbose=False)
+        key = (0, blpairs[0], "xx")
+        var_foreground_dependent = np.abs(np.diagonal(uvp_foreground_dependent.get_cov(key), axis1=1, axis2=2))
+        ps_products = np.sqrt(2)*np.real(uvp_autos.get_data(key))*np.sqrt(np.abs(np.diagonal(uvp_autos.get_cov(key), axis1=1, axis2=2))) + np.abs(np.diagonal(uvp_autos.get_cov(key), axis1=1, axis2=2))
+        # Compare the analytic variance from QE formalism with rough estimation from power spectrum products
+        nt.assert_true(np.isclose(var_foreground_dependent[:,2:-2], np.abs(ps_products)[:,2:-2], rtol=0.4).all())
```
nkern, line 1595: this looks like a factor of 2 difference--any idea why this might be the case?

hera_pspec/pspecdata.py
JianrongTan: Sure. Nice suggestions!

hera_pspec/pspecdata.py
JianrongTan: I keep the filter extension here to make sure the shape of the input covariance array for the spectral window is the same as ones obtained in other methods, such as when model is chosen as `empirical' or 'dset', we have the processed data self.x(key, include_extension=True), self.w(key, include_extension=True), which are already clipped by filter extension. And yes the default filter_extension is set to [0, 0].

hera_pspec/pspecdata.py
JianrongTan: Sure.

hera_pspec/pspecdata.py
nkern: Okay, let's use the same structure as `PSpecData.x()`. Can we include a `include_extensions=False` kwarg in the function call and make this
```
if include_extensions:
    spw = slice(...
else:
    spw = slice(...
```

JianrongTan: Sure.

hera_pspec/pspecdata.py
```diff
@@ -3357,6 +3435,13 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
             given an input visibility noise model, and store the output
             in the UVPSpec object.
 
+        store_QE_error : bool, optional
```
nkern, line 3438: can we change this to `store_cov_diag`, which I think makes clearer what is actually happening. The doc string is fine, I think

hera_pspec/pspecdata.py
```diff
@@ -3756,7 +3851,7 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
                             pv[:,t] = pv[:,t] * sd
 
                     #Generate the covariance matrix if error bars provided
-                    if store_cov:
+                    if store_cov or store_QE_error:
```
nkern, line 3854: Is this needed? if store_QE_error is True, then certainly store_cov is true, right?

hera_pspec/pspecdata.py
```diff
@@ -3956,7 +4064,9 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
             uvp.cov_array_real = cov_array_real
             uvp.cov_array_imag = cov_array_imag
             uvp.cov_model = cov_model
-
+        if store_QE_error:
+            uvp.stats_array = odict()
+            uvp.stats_array["QE_"+cov_model] = stats_array_cov_model
```
nkern, line 4069: likewise can we rename this to `cov_model` + "_diag"

nkern: Also, in this indented if statement, we should also remove the `cov_array_real`, as the point of storing its diagonal is so we don't have to store the full array

hera_pspec/pspecdata.py
JianrongTan: Sure.

hera_pspec/pspecdata.py
JianrongTan: store_cov and store_cov_diag would be independent choices here, while the former one determines if we store the whole cov_array, and the latter one determines if we store only its diagonal.

hera_pspec/pspecdata.py
JianrongTan: Since choices of store_cov and store_cov_diag are independent to each other, we can just set store_cov=False and store_cov_diag=True to not store the full cov_array.

hera_pspec/pspecdata.py
nkern: right, but I think we should do one OR the other, not both. That would be storing redundant information

hera_pspec/pspecdata.py
nkern: I think my comment above applies here: I think we should store either the full cov_array as cov_array_real etc. or just its diagonal in stats_array.

I recall us having this discussion a few years ago (when you initially made the cov_array_real and cov_array_imag distinction), and we decided to not store the diagonal b/c it can be accessed by just manually taking the diagonal of cov_array. The same logic applies here I think--if someone is explicitly asking for `store_cov_diagonal`, then we should remove the full array, as the only reason we would just store the diagonal is to save memory / disk space.

JianrongTan: Maybe we just set store_cov=False when store_cov_diag=True?

nkern: Ah, yes I think that would work!

nkern:
@JianrongTan this looks close to being done--lmk if you've resolved the factor of 2 problem in the unittest

nkern:
@JianrongTan added some functionality to reproduce my ensemble signal + noise simulations for validating the P_N errorbar. I've updated the unit-test in question and made the analogous validation test for the autocorrelation-derived P_N floor, which passes.

When trying this for the foreground_dependent errorbar a few things fail. 1. I get a factor of 2 off inside the wedge, and 2. I am getting negative values for the diagonal of the covariance, which I don't think should happen. Lmk if you have any insight into this. perhaps we can set up a time to chat and I can walk you through what I've added.

nkern:
@JianrongTan there seem to be a lot of instances in pspecdata where we do `self.spw_range[0]-self.filter_extension[0], self.spw_range[1]+self.filter_extension[1]`

I think it would be more efficient and safer if we made a new method 
```
def get_spw(self, include_extension=False):
    """
    Get self.spw_range with or without spw extension (self.filter_extension)

    Parameters
    -----------
    include_extension : str
        If True, include self.filter_extension in spw_range
   
    Returns
    --------
    tuple
        spectral window in (start_chan, end_chan)
    """
    if include_extension:
        return self.spw_range[0] - self.filter_extension[0], self.spw_range[1] + self.filter_extension[1]
    else:
        return self.spw_range
```

JianrongTan:
> @JianrongTan there seem to be a lot of instances in pspecdata where we do `self.spw_range[0]-self.filter_extension[0], self.spw_range[1]+self.filter_extension[1]`
> 
> I think it would be more efficient and safer if we made a new method and called that when we need it
> 
> ```
> def get_spw(self, include_extension=False):
>     """
>     Get self.spw_range with or without spw extension (self.filter_extension)
> 
>     Parameters
>     -----------
>     include_extension : str
>         If True, include self.filter_extension in spw_range
>    
>     Returns
>     --------
>     tuple
>         spectral window in (start_chan, end_chan)
>     """
>     if include_extension:
>         return self.spw_range[0] - self.filter_extension[0], self.spw_range[1] + self.filter_extension[1]
>     else:
>         return self.spw_range
> ```

Sure.

hera_pspec/pspecdata.py
```diff
@@ -1797,95 +1854,134 @@ def get_analytic_covariance(self, key1, key2, M=None, exact_norm=False, pol=Fals
...
+                q_q += np.einsum('bij, cji->bc', E12C21, E12C21, optimize=einstein_path)
+
+            q_qdagger = np.einsum('bij, cji->bc', E12C22, E21C11, optimize=einstein_path) 
+            if np.isclose(E12P21, 0).all() or np.isclose(E12starS21,0).all():
+                q_qdagger += 0.+1.j*0
+            else:
+                q_qdagger += np.einsum('bij, cji->bc', E12P21, E12starS21, optimize=einstein_path)
+            if model == 'foreground_dependent':
+                q_qdagger -= 2*np.einsum('bij, cji->bc', np.multiply(E_matrices, np.diag(C22_autos[time_index])), np.multiply(np.transpose(E_matrices.conj(), (0,2,1)), np.diag(C11_autos[time_index])), optimize=einstein_path)
+                q_qdagger -= np.einsum('bij, cji->bc', np.matmul(E_matrices, C22_outer_product[time_index]), np.matmul(np.transpose(E_matrices.conj(), (0,2,1)), C11_outer_product[time_index]), optimize=einstein_path)
```
acliu, line 1918: This doesn't appear to take advantage of the fact that `C11_outer_product` and `C22_outer_product` can be written as the outer product of two vectors. I'm curious---is this because it was actually quicker this way, or is it just that it's more cumbersome to code it in a way that explicitly takes advantage of the outer product? I don't mind either way as long as it's quick enough for our purposes.

hera_pspec/pspecdata.py
```diff
@@ -1797,95 +1854,134 @@ def get_analytic_covariance(self, key1, key2, M=None, exact_norm=False, pol=Fals
...
+            else:
+                q_q += np.einsum('bij, cji->bc', E12C21, E12C21, optimize=einstein_path)
+
+            q_qdagger = np.einsum('bij, cji->bc', E12C22, E21C11, optimize=einstein_path) 
+            if np.isclose(E12P21, 0).all() or np.isclose(E12starS21,0).all():
+                q_qdagger += 0.+1.j*0
+            else:
+                q_qdagger += np.einsum('bij, cji->bc', E12P21, E12starS21, optimize=einstein_path)
+            if model == 'foreground_dependent':
+                q_qdagger -= 2*np.einsum('bij, cji->bc', np.multiply(E_matrices, np.diag(C22_autos[time_index])), np.multiply(np.transpose(E_matrices.conj(), (0,2,1)), np.diag(C11_autos[time_index])), optimize=einstein_path)
```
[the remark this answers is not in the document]
acliu: Maybe this is the best way to do it. I'm not sure.

hera_pspec/pspecdata.py
```diff
@@ -642,27 +649,36 @@ def C_model(self, key, model='empirical', time_index=None, known_cov=None):
                 # add model to key
                 Ckey = ((dset, dset), (bl,bl), ) + (model, time_index, False, True,)
 
+        # Update self._C with known_cov
+        if known_cov is not None:
+            spw = slice(self.get_spw(include_extension=include_extension)[0], self.get_spw(include_extension=include_extension)[1])
```
nkern, line 654: just FYI, this can be done more elegantly as `spw = slice(*self.get_spw(include_extension=include_extension))`. I'm not insistent that you change it, but just wanted to let you know :)

hera_pspec/pspecdata.py
```diff
@@ -1797,95 +1854,134 @@ def get_analytic_covariance(self, key1, key2, M=None, exact_norm=False, pol=Fals
...
+            else:
+                q_qdagger += np.einsum('bij, cji->bc', E12P21, E12starS21, optimize=einstein_path)
+            if model == 'foreground_dependent':
+                q_qdagger -= 2*np.einsum('bij, cji->bc', np.multiply(E_matrices, np.diag(C22_autos[time_index])), np.multiply(np.transpose(E_matrices.conj(), (0,2,1)), np.diag(C11_autos[time_index])), optimize=einstein_path)
+                q_qdagger -= np.einsum('bij, cji->bc', np.matmul(E_matrices, C22_outer_product[time_index]), np.matmul(np.transpose(E_matrices.conj(), (0,2,1)), C11_outer_product[time_index]), optimize=einstein_path)
+
+            if np.isclose(E21C12, 0).all(): 
+                qdagger_qdagger = 0.+1.j*0
+            else:
+                qdagger_qdagger = np.einsum('bij, cji->bc', E21C12, E21C12, optimize=einstein_path)
```
acliu, line 1923: What I'm seeing here is that the computational speedups have only been applied to the `q_qdagger` term, but not the `q_q` or the `qdagger_qdagger` term. In principle, we should be able to squeeze some more performance out of this by optimizing the last two as well, no? Again, if this is good enough we can leave it as-is, but this might be good to remember for future reference.

nkern:
Thanks for adding that method @JianrongTan . I do have one more comment though, and that is in constructing the covariance you hardcode `include_extension=True` in many areas. That may be fine, as `filter_extension=[0, 0]` by defaultso it won't make a difference; however, I'm thinking about trying to make our philosophy consistent here. By creating a `include_extension` kwarg we are saying that this can be controlled via this kwarg. If we hardcode this parameter in certain areas, then probably we shouldn't be using it.

Let's either:
1. not hardcode the `include_extension` kwarg and pass it up to the user wherever it appears, or
2. get rid of it entirely and make the user control the filter extension by modifying the `self.filter_extension` attribute manually.

Let me know what you think is most appropriate, I'm happy to go with whatever you think is best

hera_pspec/pspecdata.py
JianrongTan: Thanks! I will modify it.

nkern:
@JianrongTan can we move the pdf to `examples/internal_memos/` and also include the .tex file? thanks

nkern:
@JianrongTan looks great, thanks. I'll approve shortly after a final look. One thing, though, you seem to have comitted a few `.DS_Store` files, which are not relevant to the code and are just OSX specific files. These are in the `internal_memos` and `errorbar_choices` directories. If you could `git rm` those files that would be great, thanks.

JianrongTan:
> @JianrongTan looks great, thanks. I'll approve shortly after a final look. One thing, though, you seem to have comitted a few `.DS_Store` files, which are not relevant to the code and are just OSX specific files. These are in the `internal_memos` and `errorbar_choices` directories. If you could `git rm` those files that would be great, thanks.

Have removed them. Thanks!

JianrongTan:
@nkern Thanks! Could you help merge it since I am not authorized?

PR #268, opened 2020-06-27
Fix the inconsistency between fold_spectra and average_spectra

JianrongTan:
Current method like get_data(), get_window_function(), ... in uvpspec.py will lead to bugs on the shape of outputs when folding the averaged_spectra, compared to fold spectra first and then do the average.


nkern:
@JianrongTan if its not too much trouble, it might be worth closing this PR, and opening a new branch off of #267 and adding this new code to it, rather than having to do the full re-base as you did for the previous PR. I can take a stab at it if you'd like, just lmk.

PR #279, opened 2020-08-11
Fix the inconsistency on exchanging the order of folding and averaging power spectra

JianrongTan:
This fixes an issue with power spectra folding. This is a short-term fix: there is an issue out to deprecate the "folding" behavior entirely, as it is quite opaque and can be more easily achieved with recent code added for spherical binning.

### DIFF
--- a/hera_pspec/tests/test_uvpspec.py
+++ b/hera_pspec/tests/test_uvpspec.py
@@ -514,6 +514,20 @@ class Test_UVPSpec(unittest.TestCase):
         cov_folded = uvp1.get_cov((0, ((37, 38), (38, 39)), ('xx','xx')))
         data_folded = uvp1.get_data((0, ((37,38), (38, 39)), ('xx','xx')))
 
+        # Test fold_spectra method is consistent with average_spectra()
+        uvp = testing.uvpspec_from_data(uvd, bls, data_std=uvd_std,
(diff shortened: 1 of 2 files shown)


hera_pspec/tests/test_uvpspec.py
```diff
@@ -514,6 +514,20 @@ def test_fold_spectra(self):
...
+        uvp_avg = uvp.average_spectra(time_avg=True, inplace=False)
+        # Fold averaged spectra
+        uvp_avg_folded = copy.deepcopy(uvp_avg)
+        uvp_avg_folded.fold_spectra()
+        # Fold then average
+        uvp_folded = copy.deepcopy(uvp)
+        uvp_folded.fold_spectra()
+        # Average folded spectra
+        uvp_folded_avg = uvp_folded.average_spectra(time_avg=True, inplace=False)
+        nt.assert_equal(uvp_avg_folded.get_data((0, ((37, 38), (38, 39)), 'xx')).shape, uvp_folded_avg.get_data((0, ((37, 38), (38, 39)), 'xx')).shape)
```
nkern, line 530: lets make this assert_almost_equal with a very small tolerance

JianrongTan: Sure.

PR #283, opened 2020-08-24
added min_EW_cut option to get_reds

nkern:
This adds some tools for getting down selecting on projected east-west baseline length in the `utils.get_reds()` method, which operates on UVData files. 

I also added the method `uvp.get_blpair_blvecs()`, which returns the baseline vector of the baseline-pair array (matching the order of `uvp.blpair_array`).
Although a single function for down selecting on a uvp object is not provided, this method makes it fairly straightforward to do so as
```
blp_vecs = uvp.get_blpair_blvecs()
uvp.select(blpairs=[blp for i, blp in enumerate(uvp.get_blpairs()) if abs(blp_vecs[i][0]) > my_blcut])
```


philbull (approved):
Looks good!

PR #284, opened 2020-08-24
added autos load in pspec_run for fg_dep cov_model

nkern:
Added `foreground_dependent` cov_model to list of cov_models that will automatically load autos if fed filepaths in `pspec_run`

### DIFF
--- a/hera_pspec/tests/test_pspecdata.py
+++ b/hera_pspec/tests/test_pspecdata.py
@@ -2081,6 +2081,15 @@ def test_pspec_run():
     if os.path.exists("./out.h5"):
         os.remove("./out.h5")
 
+    # test with cov_model that requires autos w/ fname as filepath 
+    fnames = glob.glob(os.path.join(DATA_PATH, "zen.even.xx.LST.1.28828.uvOCRSA"))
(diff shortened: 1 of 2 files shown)


hera_pspec/tests/test_pspecdata.py
```diff
@@ -2081,6 +2081,15 @@ def test_pspec_run():
     if os.path.exists("./out.h5"):
         os.remove("./out.h5")
 
+    # test with cov_model that requires autos w/ fname as filepath
```
JianrongTan, line 2084: Can you explain what the test is for? Sorry I don't quite get it.

nkern: No worries: if pspec_run is fed a string, then it will only load the required baselines from the data (not all the data), so in this case it will only load (37, 38). However, for cov_model = `autos` or `fg_dep`, the autos are required. so this is ensuring that if you feed `fnames` as a string and `blpairs` with cross correlations but cov_model=`fg_dep`, that the autos are also loaded. If they weren't, the function would error. but I am just checking here that `cov_array_real` is created as a sanity check.

JianrongTan: I see. Thanks!

PR #286, opened 2020-09-03
Fix omitted bl_tol argument in call to utils.calc_blpair_reds

philbull:
The `pspec_run` function wasn't passing the `bl_error_tol` argument to `utils.calc_blpair_reds`. One-line fix.


acliu (approved):
Looks great!

PR #287, opened 2020-09-24
metadata handling fixes in spherical_avg and combine_uvpspec

nkern:
This fixes some minor bugs in how metadata is propagated in our `spherical_average` and `combine_uvpspec` methods. 

- First, some of the `Nblpairts` arrays were not being down selected after spherical averaging (from Nblpairts -> Ntimes).

- Second, the when combining uvpspec objects the dly_array was being overwritten with a `fftfreq` call, which is only true if the pspectra have not been spherically averaged. This changes how the dly_array and freq_array are treated in the `combine_uvpspec` to simply propagate what's already in the input objects, which in hindsight is how we should have done this in the first place.

Tests are added to catch these bugs going forward...

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -2180,18 +2180,18 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
         if store_stats:
             for stat in stored_stats:
                 u.stats_array[stat][i] = np.empty((Nblpairts, spw[3], Npols), np.complex128)
-        # Set frequencies and delays: if concat_ax == 'spw' this is changed below
-        if i == 0:
-            # assumes spw metadata are the same for all uvps
-            u.spw_array = uvps[0].spw_array.copy()
-            u.spw_freq_array = uvps[0].spw_freq_array.copy()
-            u.spw_dly_array = uvps[0].spw_dly_array.copy()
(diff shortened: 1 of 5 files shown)


hera_pspec/uvpspec.py
```diff
@@ -2180,35 +2180,26 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
...
-        u.freq_array.extend(spw_freqs)
-        u.dly_array.extend(spw_dlys)
-
-    # Convert to numpy arrays
-    u.spw_array = np.array(u.spw_array)
-    u.spw_freq_array = np.array(u.spw_freq_array)
-    u.spw_dly_array = np.array(u.spw_dly_array)
-    u.freq_array = np.array(u.freq_array)
-    u.dly_array = np.array(u.dly_array)
+        # Set frequencies and delays: if concat_ax == 'spw' this is changed below
```
JianrongTan, line 2183: Seems we could move this block outside the loop?

nkern: ah yes, sure thing

nkern:
@JianrongTan thanks for the comment--lmk if anything else comes up

JianrongTan (approved):
Looks good to me. Thanks @nkern!

PR #301, opened 2021-01-07
Moved CI to GitHub actions

nkern:
In line with hera_cal and other repos, this moves our CI to GitHub actions from travis.ci


.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Tests
+    env:
+      ENV_NAME: hera_pspec_tests
+      PYTHON: ${{ matrix.python-version }}
+      OS: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: [3.6, 3.7]
```
philbull, line 20: Do we want to explicitly support Python 3.6? The final supported release was in August 2020, and it's security bugfixes only from now on. I would be fine with moving to Python 3.7 only...

philbull (approved):
Looks good to me!

.github/workflows/ci.yml
nkern: Hmm interesting point. Perhaps its good to include it here until it is officially deprecated? It seems to me a lot of people use Python 3.6, so maybe its good to just keep it in our tests until we have reason to suspect people will be forced to move on?

nkern: I'm going to merge the PR, but if you think we should address this please raise an issue, thanks!

PR #303, opened 2021-02-24
spherical averaging acceleration.

### LINKED ISSUE #313: Spherical averaging can be sped up by Nblpairs and have its memory footprint shrunk by Nblpairs for per-baseline power-spectra.
Adressed in #303 

Weighting of each per-baseline power spectrum in spherical binning is assumed to be performed by a matrix that is Ndlyblpairs x Ndlyblpairs which we might want to do if we wanted to pre-weight each per-baseline power-spectrum by  an inverse covariance matrix before averaging. If the covariances between baselines are ignored (which they currently are anyways), then this matrix is block diagonal and can be represented by a matrix that is Ndlyblpairs x Ndly. 

Doing this shrinks the memory footprint by a factor of nblpair by representing the Ndlyblpair x Ndlyblpair block-diagonal dimensions of the E matrix as an Ndlyblpair x Ndly matrix instead.

This also speeds up the matrix multiplication of E @ A by a factor of Nblpair (which dominates the matrix multiplication runtime for A^T @ E @ A.

In light of this, I think that we should at least offer an option to enable this speedup and decide what we want to use by default.

aewallwi:
Closes #313 

Reduces memory footprint of spherical_average by a factor of nblpair by representing the Ndlyblpair x Ndlyblpair block-diagonal dimensions of the E matrix as an Ndlyblpair x Ndly matrix instead.

This also speeds up the matrix multiplication of E @ A by a factor of Nblpair (which dominates the matrix multiplication runtime for A^T @ E @ A.

### DIFF
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -799,7 +799,8 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
         for t in range(uvp.Ntimes):
             for p in range(uvp.Npols):
                 for b in range(uvp.Nblpairs):
-                    EmAm[p, t, Ndlys * b: Ndlys * (b+1)] = Em[p, t, Ndlys * b: Ndlys * (b+1)].squeeze() @ Am[p, t, Ndlys * b: Ndlys * (b + 1)].squeeze()
+                    blpslice = slice(Ndlys * b, Ndlys * (b + 1))
+                    EmAm[p, t, blpslice] = Em[p, t, blpslice] @ Am[p, t, blpslice]
         invAEA = np.linalg.pinv(Am.transpose(0, 1, 3, 2) @ EmAm)
         H = EmAm @ invAEA
         Ht = H.transpose(0, 1, 3, 2)
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -98,7 +98,7 @@ def sample_baselines(bls, seed=None):
     return [random.choice(bls) for i in range(len(bls))]
 
 
-def average_spectra(uvp_in, blpair_groups=None, time_avg=False, 
+def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     blpair_weights=None, error_field=None,
                     error_weights=None,
                     normalize_weights=True, inplace=True,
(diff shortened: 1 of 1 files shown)


hera_pspec/grouping.py
```diff
@@ -787,12 +788,20 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
...
-        invAEA = np.linalg.pinv(Am.transpose(0, 1, 3, 2) @ Em @ Am)
-        H = Em @ Am @ invAEA
+        # Multiply block diagoinal Em @ Am
+        # by applying each baseline block in Em
+        # to each Ndly x Nk baseline-horizontal block in Am
+        EmAm = np.zeros_like(Am)
+        for t in range(uvp.Ntimes):
+            for p in range(uvp.Npols):
+                for b in range(uvp.Nblpairs):
+                    EmAm[p, t, Ndlys * b: Ndlys * (b+1)] = Em[p, t, Ndlys * b: Ndlys * (b+1)].squeeze() @ Am[p, t, Ndlys * b: Ndlys * (b + 1)].squeeze()
```
nkern, line 802: why is squeeze necessary?

hera_pspec/grouping.py
```diff
@@ -787,12 +788,20 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
...
         Em = np.moveaxis(E, -1, 0)
-        invAEA = np.linalg.pinv(Am.transpose(0, 1, 3, 2) @ Em @ Am)
-        H = Em @ Am @ invAEA
+        # Multiply block diagoinal Em @ Am
+        # by applying each baseline block in Em
+        # to each Ndly x Nk baseline-horizontal block in Am
+        EmAm = np.zeros_like(Am)
+        for t in range(uvp.Ntimes):
+            for p in range(uvp.Npols):
+                for b in range(uvp.Nblpairs):
```
nkern, line 801: let's make a variable
blpslice = slice(Ndlys * b, Ndlys * (b+1))
and then use that to index the arrays in the line below

anguta: good idea. I've added this.

hera_pspec/grouping.py
anguta: I've removed the squeeze. I sometimes I do it reflexively even if it isn't necessary.

nkern (approved):
I think this looks good!

PR #304, opened 2021-02-24
Save memory by only initializing the window function and covariance matrices if store_cov is True and store_window is True

aewallwi:
With an eye towards nightly power spectra for the RTP that dont necessarily need to produce detailed window functions and covariance matrices, this PR only stores the large cov and window function arrays when store_cov is True and store_window is True.


philbull (approved):
This looks like a sensible change to me. The tests are also passing, so I approve.

PR #305, opened 2021-03-15
Updated notebooks to match changes in functionality

hughbg:
Tested all notebooks in examples/, and four failed to run. The issues and fixes are listed by notebook.

## Bootstrap example

The import of uvutils fails. uvutils doesn’t exist now, and is not used in the notebook.

Fix: remove it from the import.

## Error estimation

*1.* 

Calls to ds.pspec fail. The cov_models parameter has changed to cov_model, and only accepts one string as a value. One of these was  “time_average” should be “empirical”. ds.spec can only return one object, but two are expected.

Fix: Create two calls to ds.pspec, one with cov_model=”empirical” and one with cov_model=”autos”. 

*2.*

Plotting fails because get_cov() does not take a cov_model parameter now. 

Fix: Remove that parameter and plot the appropriate data by using the two objects from the two separate calls to ds.pspec().

## Forming pseudo stokes, PS estimation

Same problem in both. dset1 and dset2 are set to strings (file names).  Attempts to use them fail because they should be UVData() objects

Fix: Create them as UVdata objects and load data from the files.


philbull (approved):
These changes look good to me. It's unfortunate that the hera_sim-related warnings are included in the notebooks, but this should resolve when hera_sim gets its next version bump.

PR #314, opened 2021-03-22
Bash script interfaces for generate_pstokes and noise from autocorrelations

aewallwi:
Expanded bash support <del>(with slight change to `grouping.bootstrap_average_blpairs`)</del> to enable running essential pspec steps in makeflow.

* `auto_noise_run.py` is a bash interface for computing thermal-noise error bars from autocorrelations.
* `generate_pstokes_run.py` is a bash script interface for generating psuedo-stokes visibilities from linear pol visibilities as preprocessing step.
* Added additional options for string formatting for polpairs and spw_ranges in `pspec_run` to make it easier to pass multiple polpairs and spw_ranges from bash (Closes https://github.com/HERA-Team/hera_pspec/issues/318)

### DIFF
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -1172,13 +1172,13 @@ def bootstrap_average_blpairs(uvp_list, blpair_groups, time_avg=False,
             j += n_blps
 
     # Loop over UVPSpec objects and calculate averages in each blpair group,
-    # using the bootstrap-sampled blpair weights
-    if hasattr(uvp, 'stats_array'):
-        error_fields = list(uvp.stats_array.keys())
-    else:
-        error_fields = None
     uvp_avg = []
     for i, uvp in enumerate(uvp_list):
+        # using the bootstrap-sampled blpair weights
+        if hasattr(uvp, 'stats_array'):
+            error_fields = list(uvp.stats_array.keys())
+        else:
+            error_fields = None
         _uvp = average_spectra(uvp, blpair_groups=blpair_grps_list[i],
                                blpair_weights=blpair_wgts_list[i],
                                time_avg=time_avg, inplace=False, error_field=error_fields)
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -688,9 +688,7 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
         A[spw] = np.zeros((uvp.Ntimes, Ndlyblps, Nk, uvp.Npols), dtype=np.float64)
 
         # setup weighting matrix: block diagonal for each Ndly x Ndly
-        # we can represent the Ndlyblps x Ndlyblps block diagonal matrix as Ndlyblps x Ndlys
-        E = np.zeros((uvp.Ntimes, Ndlyblps, Ndlys, uvp.Npols), dtype=np.float64)
-
+        E = np.zeros((uvp.Ntimes, Ndlyblps, Ndlyblps, uvp.Npols), dtype=np.float64)
 
         # get kperps for this spw: shape (Nblpairts,)
         kperps = uvp.get_kperps(spw, little_h=True)
@@ -746,21 +744,22 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
                 # weight by inverse (real) covariance
                 for p in range(uvp.Npols):
                     # the covariance is block diagonal assuming no correlations between baseline-pairs
-                    E[:, dslice, :, p] = np.linalg.pinv(cov_array_real[spw][:, dslice, dslice, p].real)
+                    E[:, dslice, dslice, p] = np.linalg.pinv(cov_array_real[spw][:, dslice, dslice, p].real)
 
             elif error_weights is not None:
                 # fill diagonal with by 1/stats_array^2 as weight
                 stat_weight = stats_array[error_weights][spw][:, dslice].real.copy()
                 np.square(stat_weight, out=stat_weight, where=np.isfinite(stat_weight))
-                E[:, range(dstart, dstop), range(0, Ndlys)] = 1 / stat_weight.clip(1e-40, np.inf)
+                E[:, range(dstart, dstop), range(dstart, dstop)] = 1 / stat_weight.clip(1e-40, np.inf)
 
             else:
-                E[:, range(dstart, dstop), range(0, Ndlys)] = 1.0
+                # uniform weighting along diagonal, except for flagged data
+                E[:, range(dstart, dstop), range(dstart, dstop)] = 1.0
                 f = np.isclose(uvp.integration_array[spw][blpt_inds] * uvp.nsample_array[spw][blpt_inds], 0)
-                E[:, range(dstart, dstop), range(0, Ndlys)] *= (~f[:, None, :])
+                E[:, range(dstart, dstop), range(dstart, dstop)] *= (~f[:, None, :])
 
             # append to non-dly arrays
-            Emean = np.trace(E[:, dslice, :], axis1=1, axis2=2)  # use sum of E across delay as weight
+            Emean = np.trace(E[:, dslice, dslice], axis1=1, axis2=2)  # use sum of E across delay as weight
             wgt_array[spw] += wgts * Emean[:, None, None, :]
             integration_array[spw] += ints * Emean
             nsample_array[spw] += nsmp
@@ -788,21 +787,12 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
         # first compute: H = E A [A.T E A]^-1
         # move axes to enable matmul and inv over Ndlyblps and Nk axes
         # Am shape (Npols, Ntimes, Ndlyblps, Nk)
-        # Em shape (Npols, Ntimes, Ndlyblps, Ndlys)
+        # Em shape (Npols, Ntimes, Ndlyblps, Ndlyblps)
         # Ht shape (Npols, Ntimes, Nk, Ndlyblps)
         Am = np.moveaxis(A[spw], -1, 0)
         Em = np.moveaxis(E, -1, 0)
-        # Multiply block diagoinal Em @ Am
-        # by applying each baseline block in Em
-        # to each Ndly x Nk baseline-horizontal block in Am
-        EmAm = np.zeros_like(Am)
-        for t in range(uvp.Ntimes):
-            for p in range(uvp.Npols):
-                for b in range(uvp.Nblpairs):
-                    blpslice = slice(Ndlys * b, Ndlys * (b + 1))
-                    EmAm[p, t, blpslice] = Em[p, t, blpslice] @ Am[p, t, blpslice]
-        invAEA = np.linalg.pinv(Am.transpose(0, 1, 3, 2) @ EmAm)
-        H = EmAm @ invAEA
+        invAEA = np.linalg.pinv(Am.transpose(0, 1, 3, 2) @ Em @ Am)
+        H = Em @ Am @ invAEA
         Ht = H.transpose(0, 1, 3, 2)
 
         # bin data: p_sph = H.T p_cyl
@@ -1172,16 +1162,12 @@ def bootstrap_average_blpairs(uvp_list, blpair_groups, time_avg=False,
             j += n_blps
 
     # Loop over UVPSpec objects and calculate averages in each blpair group,
+    # using the bootstrap-sampled blpair weights
     uvp_avg = []
     for i, uvp in enumerate(uvp_list):
-        # using the bootstrap-sampled blpair weights
-        if hasattr(uvp, 'stats_array'):
-            error_fields = list(uvp.stats_array.keys())
-        else:
-            error_fields = None
         _uvp = average_spectra(uvp, blpair_groups=blpair_grps_list[i],
                                blpair_weights=blpair_wgts_list[i],
-                               time_avg=time_avg, inplace=False, error_field=error_fields)
+                               time_avg=time_avg, inplace=False)
         uvp_avg.append(_uvp)
 
     # Return list of averaged spectra for now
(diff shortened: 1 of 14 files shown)


hera_pspec/grouping.py
```diff
@@ -1103,10 +1103,10 @@ def bootstrap_average_blpairs(uvp_list, blpair_groups, time_avg=False,
                         for blpg in blpair_groups]
         blpair_groups = new_blp_grps
 
-    # Homogenise input UVPSpec objects in terms of available polarizations 
+    # Homogenise input UVPSpec objects in terms of available polarizations
     # and spectral windows
     if len(uvp_list) > 1:
-        uvp_list = uvpspec_utils.select_common(uvp_list, spws=True, 
+        uvp_list = uvpspec_utils.select_common(uvp_list, spws=True,
```
aewallwi, line 1109: This miss was generated by white space change.

hera_pspec/grouping.py
```diff
@@ -1161,10 +1161,10 @@ def bootstrap_average_blpairs(uvp_list, blpair_groups, time_avg=False,
             blpair_wgts_list[i].append( list(_wgts) )
             j += n_blps
 
-    # Loop over UVPSpec objects and calculate averages in each blpair group, 
-    # using the bootstrap-sampled blpair weights
```
nkern, line 1165: looks like this comment got rm'd

hera_pspec/pspecbeam.py
```diff
@@ -751,14 +751,14 @@ def power_beam_sq_int(self, pol='pI'):
...
+
         if pol in self.OmegaPP.keys():
             return self.OmegaPP[pol]
         else:
-            raise KeyError("OmegaPP not specified for polarization '%s'. " 
+            raise KeyError("OmegaPP not specified for polarization '%s'. "
                            "Available polarizations are: %s" \
                            % (pol, self.available_pols))
-    
+
```
nkern, line 761: I can't tell--did you actually change any code in pspecbeam.py?

hera_pspec/pspecdata.py
```diff
@@ -3999,7 +4012,11 @@ def pspec_run(dsets, filename, dsets_std=None, cals=None, cal_flag=True,
     # Construct dataset pairs to operate on
     Ndsets = len(dsets)
     if dset_pairs is None:
-        dset_pairs = list(itertools.combinations(range(Ndsets), 2))
+        # this fails silently if we only provided a single dset.
```
nkern, line 4015: this comment is out of place, b/c it is referring to the old behavior, which is now fixed, so it won't match what will be in the codebase once this is merged. perhaps something like "# define dset_pairs if not fed"

hera_pspec/pspecdata.py
```diff
@@ -4218,28 +4235,28 @@ def get_pspec_run_argparser():
     a = argparse.ArgumentParser(description="argument parser for pspecdata.pspec_run()")
 
     def list_of_int_tuples(v):
-        v = [tuple([int(_x) for _x in x.split()]) for x in v.split(",")]
+        v = [tuple([int(_x) for _x in x.split('~')]) for x in v.split(",")]
```
nkern, line 4238: this should definitely maintain backwards compatibility! this should be fairly easy by looking for the unique `~` character, which if it exists uses the new version, if not uses the old version. and a brief docstring on this function explaining the two expected formats would be nice.

hera_pspec/pspecdata.py
```diff
@@ -4218,28 +4235,28 @@ def get_pspec_run_argparser():
...
         return v
 
     def list_of_str_tuples(v):
-        v = [tuple([str(_x) for _x in x.split()]) for x in v.split(",")]
+        v = [tuple([str(_x) for _x in x.split('~')]) for x in v.split(",")]
```
nkern, line 4242: same here

hera_pspec/pspecdata.py
```diff
@@ -4218,28 +4235,28 @@ def get_pspec_run_argparser():
...
         return v
 
     def list_of_tuple_tuples(v):
-        v = [tuple([int(_x) for _x in x.split()]) for x in v.split(",")]
+        v = [tuple([int(_x) for _x in x.split('~')]) for x in v.split(",")]
```
nkern, line 4246: and here

hera_pspec/tests/test_pspecdata.py
```diff
@@ -2188,8 +2188,8 @@ def test_window_funcs():
 
 def test_get_argparser():
     args = pspecdata.get_pspec_run_argparser()
-    a = args.parse_args([['foo'], 'bar', '--dset_pairs', '0 0, 1 1', '--pol_pairs', 'xx xx, yy yy',
-                         '--spw_ranges', '300 400, 600 800', '--blpairs', '24 25 24 25, 37 38 37 38'])
+    a = args.parse_args([['foo'], 'bar', '--dset_pairs', '0~0,1~1', '--pol_pairs', 'xx~xx,yy~yy',
```
nkern, line 2191: Let's keep the old test with the spaces to ensure backwards compatibility, as well as the new test with the updated delimiter character

hera_pspec/tests/test_pstokes.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 import os, sys
 from hera_pspec.data import DATA_PATH
-from .. import pstokes 
+from .. import pstokes
```
nkern, line 5: My personal opinion is to not use relative imports in the testing suite, so I think from hera_pspec import pstokes is better here (I know that wasn't what was added in this PR but might as well address it while we are here)

hera_pspec/uvpspec.py
```diff
@@ -1967,18 +1967,18 @@ def average_spectra(self, blpair_groups=None, time_avg=False,
...
-            If `error_weight` is set to `None`, which means we just use the 
+
+            If `error_weight` is set to `None`, which means we just use the
             integration time as weights. If `error_weights` is specified,
             then it also gets appended to `error_field` as a list.
             Default: None.
 
         inplace : bool, optional
-            If True, edit data in self, else make a copy and return. 
+            If True, edit data in self, else make a copy and return.
```
nkern, line 1981: same question here: is anything actually changed here?

scripts/generate_pstokes_run.py
```diff
@@ -0,0 +1,21 @@
...
+ap = pstokes.generate_pstokes_argparser()
+args = ap.parse_args()
+uvd = UVData()
+uvd.read(args.inputdata)
+if args.outputdata is None:
+    args.outputdata = args.inputdata
+# generate pstokes and append to file.
+for p in args.pstokes:
+    if pyuvdata.utils.polstr2num(p) not in uvd.polarization_array:
+        uvd += pstokes.construct_pstokes(uvd, uvd, pstokes=p)
```
nkern, line 19: I'm fairly certain that saving data files with both linear and stokes pol can be quite messy for some data types (e.g. fits) that require an even stride between the polarization integers (e.g. -5, -6, -7, -8).

also not clear that it makes sense to write another data file to disk with copies of the vis pol data, just unnecessary data size bloat, no?

hera_pspec/pspecbeam.py
aewallwi: I didn't really change any code.

hera_pspec/uvpspec.py
aewallwi: Nothing actually changed (my editor removed a space).

scripts/generate_pstokes_run.py
aewallwi: Agreed. Storing lin-pols in the same file works fine in uvh5 but it causes problems with uvfits. I've added a new `--inplace` arg that is default `False`.

hera_pspec/tests/test_pstokes.py
aewallwi: I was having trouble getting the correct coverage registered without relative imports (when I don't use relative imports, for some reason running something like `pytest --cov-report html` results in no coverage being registered. Maybe something else is configured incorrectly that is causing this. I had the same problem migrating to pytest in `uvtools` recently.

hera_pspec/tests/test_pspecdata.py
aewallwi: Good idea.

hera_pspec/pspecdata.py
aewallwi: Agreed. I'll add an option for the old system if no "~" are registered.

hera_pspec/pspecdata.py
aewallwi: will do.

hera_pspec/pspecdata.py
aewallwi: will do.

hera_pspec/pspecdata.py
```diff
@@ -3244,7 +3248,7 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
 
                     # Calculate unnormalized bandpowers
                     if verbose: print("  Building q_hat...")
-                    qv = self.q_hat(key1, key2, exact_norm=exact_norm, pol=pol)
+                    qv = self.q_hat(key1, key2, exact_norm=exact_norm, pol=pol, allow_fft=allow_fft)
```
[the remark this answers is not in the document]
aewallwi: `get_Q_alt` returns the same matrix regardless of whether `allow_fft` is set though I agree we should propagate it there as well to avoid confusion.

hera_pspec/grouping.py
aewallwi: Oops. Will restore it.

hera_pspec/pspecdata.py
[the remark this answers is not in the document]
aewallwi: Looking some more `get_Q_alt` does seem to have internal checks that avoid ever using the fft unless `spw_Ndlys == nfreq`. I think this should be fine even if we have `sampling=False` since `Q_alt` is defined to be with sampling (infinitesimal fourier bins).

hera_pspec/pspecdata.py
aewallwi: Will fix this.

hera_pspec/tests/test_pstokes.py
nkern: I see, yeah that's weird--i get coverage locally on my machine.

I think this may not be the place to institute this change, but I do think non-relative imports seems to be the standard for test files (see pyuvdata and astropy, for example). let's come back to this later, I'll make an issue

scripts/generate_pstokes_run.py
```diff
@@ -0,0 +1,27 @@
...
+    # if inplace, append new pstokes onto existing file.
+    uvd_output = copy.deepcopy(uvd)
+else:
+    # otherwise, output uvd does not contain original polarizations.
+    uvd_output = pstokes.construct_pstokes(uvd, uvd, args.pstokes[0])
+for p in args.pstokes:
+    if pyuvdata.utils.polstr2num(p) not in uvd_output.polarization_array:
+        uvd_output += pstokes.construct_pstokes(uvd, uvd, pstokes=p)
+
+# overwrite file.
```
nkern, line 26: the read and write have different arguments, so perhaps this comment should just be: write the file.

also I don't think inplace is really the correct kwarg here, b/c in the normal use-case, the outputdata is not the inputdata, so you are writing a new file, not overwriting the original file (the latter of which we never do in our pipeline anyways by policy, I think).

b/c it has a specific use-case of either keeping the vispols or not, perhaps its better to use something like `keep_vispols`?

aewallwi: Good point. I will change `inplace` to `keep_vispols`.

nkern:
@aewallwi can you confirm that your edits to grouping.py are correct? If I'm not mistaken, this looks like reverting your previous PR on spherical binning. Seems like you  rebased improperly? Please confirm

nkern:
@aewallwi also it looks like there is no significant change to `bootstrap_average_blpairs`, can you explain what changes you made that you mention in the PR description?

aewallwi:
> @aewallwi also it looks like there is no significant change to `bootstrap_average_blpairs`, can you explain what changes you made that you mention in the PR description?

There should be no changes in this PR. I had made some originally then decided against it. I've made sure to remove errant comments and line breaks.

aewallwi:
> @aewallwi can you confirm that your edits to grouping.py are correct? If I'm not mistaken, this looks like reverting your previous PR on spherical binning. Seems like you rebased improperly? Please confirm

Thanks for catching this! There must have been a rebase bug. I've restored the speed up code.

hera_pspec/tests/test_pstokes.py
aewallwi: sounds good!

scripts/generate_pstokes_run.py
```diff
@@ -0,0 +1,27 @@
...
+import pyuvdata
+
+ap = pstokes.generate_pstokes_argparser()
+args = ap.parse_args()
+uvd = UVData()
+uvd.read(args.inputdata)
+if args.outputdata is None:
+    args.outputdata = args.inputdata
+if args.keep_vispols:
+    # if inplace, append new pstokes onto existing file.
```
nkern, line 17: change comment to `if keep vispol`

scripts/generate_pstokes_run.py
```diff
@@ -0,0 +1,27 @@
...
+    uvd_output = copy.deepcopy(uvd)
+else:
+    # otherwise, output uvd does not contain original polarizations.
+    uvd_output = pstokes.construct_pstokes(uvd, uvd, args.pstokes[0])
+for p in args.pstokes:
+    if pyuvdata.utils.polstr2num(p) not in uvd_output.polarization_array:
+        uvd_output += pstokes.construct_pstokes(uvd, uvd, pstokes=p)
+
+# overwrite file.
```
nkern, line 26: change to `write file`

hera_pspec/pspecdata.py
```diff
@@ -4218,28 +4234,45 @@ def get_pspec_run_argparser():
     a = argparse.ArgumentParser(description="argument parser for pspecdata.pspec_run()")
 
     def list_of_int_tuples(v):
-        v = [tuple([int(_x) for _x in x.split()]) for x in v.split(",")]
+        if '~' in v:
```
nkern, line 4237: do you mind just putting the examples listed in the argparser here as well, e.g. `'0~0,1~1' --> [(0, 0), (1, 1), ...]` etc. and same for the other two functions below. thanks

nkern (approved):
Looks good thanks for the edits. a few more trivial fixes noted above, then feel free to merge, thanks

scripts/generate_pstokes_run.py
aewallwi: Good idea. I've done this.

hera_pspec/pspecdata.py
aewallwi: Sounds good! i've done this.

PR #331, opened 2021-04-05
Avoid recomputing noise scalars in `utils.generate_noise_errors`

### LINKED ISSUE #330: running `compute_scalar` on every baseline-pair in  `uvp_noise_error` significantly bloats runtime.
It takes roughly 20k seconds to run `uvp_noise_error` a file with roughly 2000 baseline-pairs, 3 pols, 170 delays, and a single time integration. Since this code is nominally computing thermal noise using the radiometer equation, I was pretty surprised it was taking so long to run. When we crack things open with a line-profiler, we see that the runtime is dominated by computing the power-spectrum scalar. 

image

image

Computing the scalar alone should not take 20k seconds either. However, the current implementation calculates and identical scalar for each baseline-pair individually. Apparently running quadrature integration for each baseline pair does add up to almost the entire run time.

The problem is that `generate_noise_spectra` (from which `compute_scalar` is called) is run for each baseline pair individually.

https://github.com/HERA-Team/hera_pspec/blob/10c290fd33e3bfdcf13bc3de80f3d9199d57c9ff/hera_pspec/utils.py#L1390

Two potential fixes are:

* Create a chaching system for the scalar
*  Only run generate_noise_spectra on all of the baseline pairs at once.

aewallwi:
Closes #330 

`utils.uvp_noise_errors` now precomputes the normalization scalar before looping through blpairs. This avoids recomputing the noise scalar many times which significantly increases runtime.

loss in coverage is due to white space removal from an already uncovered line (so diff should actually be 100% covered).

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -368,7 +368,7 @@ class UVPSpec(object):
         one entry per baseline.
 
         In a `UVPSpec` object, the dictionary is compressed so that a single
-        `r_param` entry correspondsto multiple baselines and is stored as a
+        `r_param` entry corresponds to multiple baselines and is stored as a
         JSON format string.
 
         This function reads the compressed string and returns the dictionary
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -1762,7 +1762,7 @@ class UVPSpec(object):
 
     def generate_noise_spectra(self, spw, polpair, Tsys, blpairs=None,
                                little_h=True, form='Pk', num_steps=2000,
-                               component='real'):
+                               component='real', scalar=None):
         """
         Generate the expected RMS noise power spectrum given a selection of
         spectral window, system temp. [K], and polarization. This estimate is
@@ -1826,6 +1826,11 @@ class UVPSpec(object):
             Options=['real', 'imag', 'abs'].
             If component is real or imag, divide by an extra factor of sqrt(2).
 
+        scalar : float
+            Optional noise scalar used to convert from Tsys to cosmological units
+            output from pspecbeam.compute_pspec_scalar(...,noise_scalar=True)
+            Default is None. If None provided, will calculate scalar in function.
+
         Returns
         -------
         P_N : dict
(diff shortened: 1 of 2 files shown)


aewallwi:
> addresses #330
> 
> Add a new optional array `uvpspec.noise_scalar_array` to store the noise power-spectrum scalar. This avoids recomputing the noise scalar many times in the different `utils` noise calculators that often compute the noise scalar on a per-baseline-pair cadence.

With these updates, the runtime for my test case is reduced from 1.94e4 seconds to 140 seconds

image

(compare to the 1.94e4 seconds in #330)

aewallwi:
I also double check that the noise arrays are approximately equal by comparing two files using the new and old code.

```

In [3]: psc0 = container.PSpecContainer('zen.grp1.of2.LST.0.93515.sum.low-band-a
   ...: llbls.xtalk_filtered.tavg.pspec.old.h5')

In [4]: uvp0 = psc0.get_pspec('dset0', 'dset0_x_dset0')

In [5]: psc1 = container.PSpecContainer('zen.grp1.of2.LST.0.93515.sum.low-band-a
   ...: llbls.xtalk_filtered.tavg.pspec.h5')

In [6]: uvp1 = psc1.get_pspec('dset0', 'dset0_x_dset0')

In [22]: for spw in uvp0.spw_array:
    ...:     for stat in ['P_N', 'P_SN']:
    ...:         assert np.all(np.isclose(uvp0.stats_array[stat][spw][np.isfinite(uvp0.stats_array[stat][spw])], uvp1.stats_array[stat][spw][np.is
    ...: finite(uvp1.stats_array[stat][spw])]))
```

hera_pspec/uvpspec.py
```diff
@@ -363,17 +363,17 @@ def get_integrations(self, key, omit_flags=False):
...
-        In a `PSpecData` object, the `r_params` are stored as a dictionary with 
+
+        In a `PSpecData` object, the `r_params` are stored as a dictionary with
         one entry per baseline.
-        
-        In a `UVPSpec` object, the dictionary is compressed so that a single 
-        `r_param` entry correspondsto multiple baselines and is stored as a 
+
+        In a `UVPSpec` object, the dictionary is compressed so that a single
+        `r_param` entry correspondsto multiple baselines and is stored as a
```
JianrongTan, line 371: corresponds to

aewallwi: fixed

JianrongTan (approved):
Thanks @aewallwi! All look good to me.

PR #332, opened 2021-04-09
cov_model and known_cov now parsed

### LINKED ISSUE #327: User-input covariance not properly parsed to other functions
There is an option for a user-input covariance in pspec through the arguments `cov_model` and `known_cov` (the dictionary giving the covariance matrix for each key). However these arguments are not parsed to subsequent functions such as R and so there is no effect on the output power spectrum.

adeliegorce:
Resolves #327

- Added cos_model and known_cov arguments to functions related to data covariance in order to properly parsed the potential user-input covariance matrix/dictionary. These include: iC, cov_q_hat, q_hat, get_G, get_H, get_unnormed_V, get_analytic_covariance, scalar_delay_adujstement. 
- Added these parameters to each call on the above-listed functions. It clutters the code a little bit (also the docstrings), maybe we could consider making the cov_model and the known_cov dictionary attributes?
- Added checks at the start of pspec function: asserts if cov_model is a string and if known_cov is given, in case the str isn't among the existing models (['empirical', 'dsets', 'autos', 'foreground_dependent']); if known_cov is given, asserts if the input_data_weight is set to something else than 'identity' - if it is set to identity, then raise warning saying that the input covariance will not be used in the R matrix.

Tested the code with the identity matrix as data covariance.

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -637,14 +637,15 @@ class PSpecData(object):
             # add model to key
             Ckey = ((dset, dset), (bl,bl), ) + (model, None, False, True,)
         else:
-            assert isinstance(time_index, int), "time_index must be integer if cov-model=={}".format(model)
+            if (known_cov is None): 
+                assert isinstance(time_index, int), "time_index must be integer if cov-model=={}".format(model)
             # add model to key
             Ckey = ((dset, dset), (bl,bl), ) + (model, time_index, False, True,)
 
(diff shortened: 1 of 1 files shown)


adeliegorce:
Running the tests highlighted issues in parsing the covariance information (even for already existing options such as 'dsets' and 'autos'): if input_data_weighting is set to 'iC', for any input cov_model, it is always the empirical covariance that will be computed, therefore the R matrix does not change with cov_model. In particular, it needs to depend on the time_index and so the format of Rkeys needs to be changed.

PR #333, opened 2021-04-14
added utils to init

nkern:
adds `utils` to `__init__.py`


acliu (approved):
Uncontroversial and looks good. Thanks for taking care of this!

PR #336, opened 2021-04-28
Codecov

aewallwi:
Upload coverage.xml to codecov to enable coverage comparisons and checks.


philbull (approved):
This looks fine, but it's hard to tell if it's working correctly without fully enabling it I guess...

PR #337, opened 2021-04-29
Fix issue where coverage isn't actually being recorded in coverage.xml (resulting in 0% coverage recorded by codecov).

aewallwi:
For reasons that are somewhat opaque to me, the existing pytest setup was not actually recording coverage (coverage files were produced, but all the lines were recorded as missed). By adjusting imports in the test files and removing a problematic arg, I think I've fixed this.


philbull (approved):
Looks good to me. Changes to relative imports, which is fine.

PR #342, opened 2021-06-09
Updated docstrings for get_H and scalar_delay_adjustment to reflect h…

acliu:
…ow we actually account for tapering in our code

This PR is in response to a scary moment where we thought that there was a bug in the pspec normalization. The code actually is correct, but some docstrings were misleading. I've updated the docstrings to (accurately) state that:

- Aside from the factors of the taper in `R1` and `R2`, the `get_H` function does not make any extra adjustments in normalization to account for the taper.
- There is, however, an extra factor of the taper that enters into `scalar_delay_adjustment`. This is needed for a correct power spectrum normalization, and we had implemented that in the function. However, we neglected to say that this was being done in the docstring.

@r-pascua reran the HERA validation test 0.1.0 with v0.3.0 with `hera_pspec` to show that the power spectrum is correctly normalized whether we use a taper or not. The result is good to the percent level vs Adam Lanman's analytic prediction for his simulated data. See attached file.

[new_hera_pspec_validation_test_result (1).pdf](https://github.com/HERA-Team/hera_pspec/files/6626449/new_hera_pspec_validation_test_result.1.pdf)

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -1439,15 +1439,7 @@ class PSpecData(object):
             H_ab = (1/2) Tr[R_1 Q_a^alt R_2 Q_b]
 
         (See HERA memo #44). As currently implemented, this approximates the
-        primary beam as frequency independent. Under this approximation, the
-        our H_ab is defined using the equation above *except* we have
-        Q^tapered rather than Q_b, where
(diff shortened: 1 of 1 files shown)


aewallwi (approved):
Looks good to me.

PR #343, opened 2021-06-17
Notebook and data demonstrating use cases for code review

hughbg:
This PR adds a notebook "Overview_Core_Pspec_functions" which gives examples of how to run hera_spec and use basic functions to compute, average, and plot a power spectrum. It implements the use cases listed in the [code design/review document](https://docs.google.com/document/d/1wE4U-jnMjbRjlooXg7MKMvsoAzaFp-PLdTESnYA7tQw/edit). Everything is in a single notebook for ease of review but the notebook can be split later for the final version. 

There are two main issues that were encountered when developing the code:

**Data**

It's necessary to have data that is appropriate to demonstrate different operations. One data file has been used (obtained from simulations), but different data files could be used for different operations, and we could fiddle endlessly trying different data. To avoid that we settle on this one data file.

**Alternative methods**

There is more than one way to do things, and the notebook doesn’t contain all the variations we tried (although it does show an alternative for incoherent/coherent power spectra). What's in the notebook may not necessarily be the "right" way, if there is such a thing.  

At a hera_pspec meeting it was decided not to do a few of the use cases because there is other work in those areas of the hera_pspec code. These are highlighted in the code design/review document.

*Adélie and Hugh*


philbull (changes requested):
I left some review comments on the notebook in ReviewNB: https://app.reviewnb.com/HERA-Team/hera_pspec/blob/use_cases/examples%2FOverview_Core_Pspec_functions.ipynb%2F/discussion

I have two other general comments:

1. The explanatory text in the notebook is quite sparse. Depending on how we are intending this notebook to be used, it might be worth expanding the text a little, e.g. if we want this to also function as a tutorial, it would be good to explain more fully what is going on in each step.
2. It does feel a little long. Splitting it up would be one option, but I think it's better to have it as a single self-contained notebook. In this case, a bit more help with navigation (e.g. a table of contents with hyperlinks: https://sebastianraschka.com/Articles/2014_ipython_internal_links.html) and some trimming-down of the outputs might do the trick.

adeliegorce:
Hi! I've made the changes requested by @philbull but there was a version conflict on `pyuvdata` because of `hera_sim` so I had to change `ci/hera_pspec_tests.yml`  and `setup.py` to go back to an 'old' commit on `hera_sim` (`git+https://github.com/HERA-Team/hera_sim.git@2005a45af50049c78e1f30b4bc0401b5ab3bf044` instead  of `git+https://github.com/HERA-Team/hera_sim`). This is temporary of course, until the `hera_sim` crew fixes the compatibility issue. If you confirm the commit, I will ping the `hera_pspec` team so that people are aware that this change should be eventually revoked.

philbull (changes requested):
Thanks, looks good! Just a couple of very minor comments added.

adeliegorce (approved):

Issue #344, opened 2021-07-02
hera_sim and pyuvdata API changes have broken things.


aewallwi:
API changes to `hera_sim` and `pyuvdata` are breaking `hera_pspec` in various places. A few lines that are using outdated API calls are 

https://github.com/HERA-Team/hera_pspec/blob/13e89d6c82a101da8e0e4973a7bef8386b50e6da/hera_pspec/testing.py#L488

https://github.com/HERA-Team/hera_pspec/blob/13e89d6c82a101da8e0e4973a7bef8386b50e6da/hera_pspec/testing.py#L508

and

https://github.com/HERA-Team/hera_pspec/blob/13e89d6c82a101da8e0e4973a7bef8386b50e6da/hera_pspec/pspecdata.py#L213
https://github.com/HERA-Team/hera_pspec/blob/13e89d6c82a101da8e0e4973a7bef8386b50e6da/hera_pspec/pspecdata.py#L216

There may be more.

steven-murray:
@aewallwi these look like easy enough fixes. I'll get to it sometime this week.

adeliegorce:
Hi, do you have any update on this? 
I managed for tests to run without errors yesterday by requesting `hera_pspec` to use a given `hera_sim` commit (https://github.com/HERA-Team/hera_sim/commit/2005a45af50049c78e1f30b4bc0401b5ab3bf044) as suggested by @piyanatk but 24hrs later they are now failing because of the function changes mentioned above by @aewallwi. Any idea of what changed? @steven-murray have you been able to fix the `pyuvdata` version clash?
Thanks!

aewallwi:
@adeliegorce, are the problems in `pyuvdata` or are the `hera_sim` calls still causing problems?

adeliegorce:
No, since I call a previous version of `hera_sim` which did not require `pyuvadata<2.2.0`, this part is fine, but now I get the same errors as you did in your first post, for example `hs.noise.jy2T does not exist`. The weird part is that all tests were running fine yesterday afternoon Eastern time

steven-murray:
@adeliegorce if you're using an old version of hera_sim, it should have the `jy2T` function. It sounds like you're maybe using the wrong env, or have inadvertantly updated hera_sim since yesterday?

BTW I think current hera_sim does not required pyuvdata<2.2.0?

steven-murray:
The `hera_sim` part of this issue is resolved, right? Is this issue still open because of the other calls?

[...]

PR #345, opened 2021-07-07
Attemp to update hera_sim cals for v1+

steven-murray:
Addresses #344 (partly).

### DIFF
--- a/hera_pspec/testing.py
+++ b/hera_pspec/testing.py
@@ -485,7 +485,7 @@ def sky_noise_sim(data, beam, cov_amp=1000, cov_length_scale=10, constant_per_bl
     autos = {}
     for key in bls:
         if key[0] == key[1] and key[2].upper() in AUTOVISPOLS:
-            J2K = hs.utils.jansky_to_kelvin(freqs/1e9, OmegaP[key[2]]) / 1e3
+            J2K = hs.utils.jansky_to_kelvin(freqs/1e9, OmegaP[key[2]])
             # handle stokespols
             if key[2].upper() in ['PQ', 'PU', 'PV']:
                 autos[key] = uvd.get_data(key[:2] + ('pI',)).real * J2K
--- a/hera_pspec/testing.py
+++ b/hera_pspec/testing.py
@@ -485,12 +485,11 @@ def sky_noise_sim(data, beam, cov_amp=1000, cov_length_scale=10, constant_per_bl
     autos = {}
     for key in bls:
         if key[0] == key[1] and key[2].upper() in AUTOVISPOLS:
-            J2K = hs.utils.jansky_to_kelvin(freqs/1e9, OmegaP[key[2]])
             # handle stokespols
             if key[2].upper() in ['PQ', 'PU', 'PV']:
-                autos[key] = uvd.get_data(key[:2] + ('pI',)).real * J2K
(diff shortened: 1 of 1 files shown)


steven-murray:
@aewallwi I think tests involving hera_sim are fixed here. I don't know how to do the uvdata stuff.

aewallwi (approved):
Looks good to me!

aewallwi (changes requested):
For some reason, get_analytic_covariance is now failing.

It looks like this might just be an isclose to zero issue not quite being met (maybe for numerical reasons), unrelated to the `hera_sim` API issue so i think I'll still go ahead and approve this.

hera_pspec/testing.py
```diff
@@ -485,7 +485,7 @@ def sky_noise_sim(data, beam, cov_amp=1000, cov_length_scale=10, constant_per_bl
     autos = {}
     for key in bls:
         if key[0] == key[1] and key[2].upper() in AUTOVISPOLS:
-            J2K = hs.noise.jy2T(freqs/1e9, OmegaP[key[2]]) / 1e3
+            J2K = hs.utils.jansky_to_kelvin(freqs/1e9, OmegaP[key[2]]) / 1e3
```
r-pascua, line 488: The divide by 1000 shouldn't be here anymore. In the past, this function returned Jy -> mK conversion, but now it does Jy -> K.

aewallwi (approved):

r-pascua (commented):
`sky_noise_jy` call has been updated correctly, but the `J2K` definition needs to be updated.

hera_pspec/testing.py
steven-murray: ah.

steven-murray: Done.

aewallwi (changes requested):
Now it does look like the RMS is low by a factor of 1000.

hera_pspec/testing.py
```diff
@@ -505,7 +505,7 @@ def sky_noise_sim(data, beam, cov_amp=1000, cov_length_scale=10, constant_per_bl
         Tsys = np.sqrt(autos[(bl[0], bl[0], bl[2])] * autos[(bl[1], bl[1], bl[2])])
 
         # get raw thermal noise
-        n = hs.noise.sky_noise_jy(Tsys, freqs/1e9, lsts, OmegaP[bl[2]], inttime=int_time)
+        n = hs.noise.sky_noise_jy(lsts, freqs/1e9,  autovis=Tsys, omega_p=OmegaP[bl[2]], integration_time=int_time)
```
r-pascua, line 508: Ah, so there's a problem here: `sky_noise_jy` needs `autovis` to be in Jy. Since this is being passed in units of K, there's an extra conversion that's happening under the hood. So I think you want to do `autovis=Tsys/J2K` instead. (See line 118 in `hera_sim/noise.py`.) Sorry I didn't catch this before--I had to look at more of the `sky_noise_sim` function to figure this out.

steven-murray: Ah, even easier... just never convert to K in the first place :-)

r-pascua: Tada! Much easier...

aewallwi (approved):
Looks like the `hera_sim` issues are now resolved.

PR #346, opened 2021-07-28
Reflect recent API changes in pyuvdata and hera_sim

philbull:
This PR fixes broken code and tests in light of recent API changes in `hera_sim` and `pyuvdata`. (Note that some other fixes have been applied previously.)

A simplified noise realisation function has been added to the `testing` module, which allows us to remove our dependency on `hera_sim` entirely. The dependency has been removed in the setup and CI configuration.

The example `calfits` files in the `data/` directory have been fixed to replace old/improper antenna names, and a deprecated kwarg has been removed from some calls to `uvcalibrate()`.


philbull:
(Note that `black` has been applied to the `testing.py` file to improve the formatting.)

philbull:
Closes #344

hera_pspec/testing.py
```diff
@@ -401,22 +527,87 @@ def gauss_cov_fg(cov_amp, cov_length_scale, freqs, Ntimes=100, constant_in_time=
...
+
+    Returns
+    -------
+    noise : ndarray
+        2D array of white noise in LST/freq.
+    """
+    # Beam solid angle
+    if callable(omega_p):
+        omega_p = omega_p(freqs)
+
```
adeliegorce, line 579: Add check that if not callable, then array of length same as freqs. Maybe assert freqs is array too

hera_pspec/testing.py
```diff
@@ -401,22 +527,87 @@ def gauss_cov_fg(cov_amp, cov_length_scale, freqs, Ntimes=100, constant_in_time=
...
+            2, Ntimes, Nfreqs
+        )
         s = s[0] + 1j * s[1]
 
     return s
 
 
-def sky_noise_sim(data, beam, cov_amp=1000, cov_length_scale=10, constant_per_bl=True,
-                  constant_in_time=True, bl_loop_seed=None, divide_by_nsamp=False):
+def sky_noise_jy_autos(lsts, freqs, autovis, omega_p, integration_time, channel_width=None, Trx=0.0):
```
adeliegorce, line 548: Could be worth adding to noise.py or utils.py, as it could be useful outside of testing. See `variance_from_auto_correlations` in utils.py

adeliegorce (commented):
Hi, I've done the review. I mainly just have a general comment about the new function - it could have its place in another module than testing. I, personally, would use such a function. Thanks!

hera_pspec/testing.py
philbull: Done!

hera_pspec/testing.py
philbull: I think this is OK here for now as it's legacy code supporting some tests; we can eventually replace it with an alternative in hera_sim or similar.

adeliegorce (approved):

PR #347, opened 2021-07-29
Allow P_N calculation to LSTs that wrap through 0 radians

jsdillon:
Right now, the calculation of P_N uses np.unique to figure out the LSTs, which are then used in the Tsys interpolator. However, if the input LST order wraps through 0 radians, this will produce a mismatch in the data and the associated LSTs, since the order will get changed in `lsts`. This PR fixes that by preserving the order of the input data, allowing for the extraction of unique LSTs without undoing the phase wrap.


nkern:
@jsdillon lmk when tests pass and I'll have another look, but so far looks fine. thanks. I am wondering if this will break other things downstream, but we'll probably just have to see if that's the case.

jsdillon:
Tests are passing locally, they are failing here only because of the hera_sim pyuvdata version conflict issue.

jsdillon:
image

This seemed to fix my issue. @nkern can we merge it in even though tests fail (they pass locally, it's a version installation issue that's fixed in #346 that causes them to fail here, but I'd prefer not to wait for that review)?

jsdillon:
@nkern ping?

jsdillon:
@nkern this passes tests now.

nkern (approved):
looks good thanks

PR #350, opened 2021-12-09
Exact window function computation

adeliegorce:
This pull request aims at introducing a new class in the `hera_pspec` module, dedicated to computing accurate window functions. These window functions include a better mapping from measurement space (b,tau) to cosmological space (k_perp, k_parallel). The main equations, and their counterparts in the code, are detailed in this [file](https://github.com/HERA-Team/hera_pspec/files/7686840/Window_functions.pdf), which should eventually appear as a HERA memo.

The following files/modules have been added/changed:
1.  **hera_pspec/uvwindow.py** introduces the new class, called `UVWindow`. It requires the path to a file containing the Fourier transform of the beam in the sky plane, for all frequencies along the HERA bandwidth. The main functions of the module are:
- `get_spherical_wf` which gives the window functions for a set of spherical k-bins, baselines and a spectral window.
- `get_cylindrical_wf` which gives the cylindrical window functions for a given baseline and spectral window.
- `get_kperp_bins` which gives the k_perpendicular bins that have been used in the above, for a given baseline length.
- `get_kpara_bins` which does the same for k_parallel.
2. **hera_pspec/tests/test_uvwindow.py** introduces tests related to the different functions in the new module.
3. **examples/Overview_Core_Pspec_functions.ipynb** has been modified to give an example of how to use the new functions (final section).
4. **hera_pspec/__init__.py**

This is a preliminary PR, but more changes are to come. Namely:
- For now, the window functions are only weighted by redundancy. The code will be modified to include error weights.
- Getting the cylindrical window functions for one baseline length currently takes about 8 secs. For all the redundant groups in HERA, it takes over an hour. The code will be improved to speed up calculations.
- `uvpspec.py`, `pspecdata.py` and `grouping.py` will be modified to directly call `UVWindow` and include the  computation of the exact window functions as an option.
- Validation with the validation simulations is ongoing.

This is a non-exhaustive list and I would be happy to hear any other improvement suggestion.


steven-murray (commented):
Hi @adeliegorce, thanks for this huge and important PR. I'm still in the process of review, but thought I'd post my comments thus far so I don't lose them. They're all kinda just nit-picky code stuff so far, the product of me just trying to absorb the code.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,946 @@
...
+        Get spherical window functions for a set of baselines, polarisation,
+        along a given spectral range, and for a set of kbins used for averaging.
+
+        Parameters
+        ----------
+        spw_range : tuple of ints
+            In (start_chan, end_chan). Must be between 0 and 1024 (HERA bandwidth).
+        pol : str
+            Can be pseudo-Stokes or power: 'pI', 'pQ', 'pV', 'pU', 'xx', 'yy', 'xy', 'yx'
+        kbins : array_like
```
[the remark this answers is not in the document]
adeliegorce: I'm not sure what would be the astropy.Quantity here?

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,946 @@
...
+        filename = '%s_%s.hdf5' %(file,self.pol)
+        assert os.path.isfile(filename), "Cannot find FT beam file: %s." %filename
+        f = h5py.File(filename, "r") 
+        self.mapsize = f['mapsize'][0] 
+        FT_beam = f['FT_beam'][self.spw_range[0]:self.spw_range[1],:,:]
+        HERA_bw = f['freq'][...]
+        f.close()
+
+        # Set spw parameters such as frequency range and average redshift.
+        self.set_spw_parameters(HERA_bw)
```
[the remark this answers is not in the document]
adeliegorce: The idea was to call the function again if the used changed the ft_file. I added an if statement to check this.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,946 @@
...
+        
+        """
+
+        # Summary attributes
+
+        # initialises other attributes
+        if cosmo is None: cosmo = conversions.Cosmo_Conversions()
+        self.cosmo = cosmo
+
+        assert isinstance(little_h, bool), "little_h must be boolean"
```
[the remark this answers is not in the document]
adeliegorce: The bool() functions does not seem to raise TypeError very often so instead I added a try statement with a ValueError exception (if the input is a ndarray)

aewallwi (commented):
Viewed the additions to the Core PSpec functions notebook which illustrates the new functionality pretty clearly. Wedge window functions are implemented using a new UVWindow class with a driver function that takes in list of baseline positions, spectral band, and beam information, along with a set of kbins, and produces spherical window functions for the kbins given this information; performing the appropriate sums and integrals.

aewallwi (changes requested):
Thanks for all the great work @adeliegorce . I have a few design-level requests to be considered and after that I'm happy to approve this.

* I don't see any functions to compute the FT_beam object. These should probably be included and if not, we should give detailed information on how to compute one of these in the docstrings.


* I think we should have some high level functions that automatically calculate the appropriate true window functions for a `UVPSpec` object with multiple spectral windows  and retrieving the information on these windows from the object. This might be done by having a `UVWindow` object intialize off of a non-baseline-averaged `UVPSpec` object so that it knows about all of the spectral windows and baseline-pairs. 
* There doesn't appear to be any incorporationg uniform per-baseline weights in the spherical averaging. These could be retrieved again by passing a `UVPSPec` object and specifying the weights to use.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,946 @@
...
+from pyuvdata import UVBeam, UVData 
+
+from . import conversions, noise, version, pspecbeam, grouping, utils, uvpspec_utils as uvputils
+
+class UVWindow(object):
+    """
+    An object for storing window functions copmuted without the delay approximation.
+    """
+
+    def __init__(self, ftbeam='', uvdata='',taper='none',
```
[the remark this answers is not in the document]
adeliegorce: Yes I'm currently working on this (future PR!). For now I have opted for adding a function to `UVPSpec` which computes the exact window functions of the object (something along the lines of `uvp.get_exact_window_functions`) and includes the `error_weights` if given. Since the `uvp.stats_array` retains the baseline-pair information even after averaging, we avoid the caveat of missing information. And a user warning could be sufficient to discourage someone from computing the exact window functions of an averaged `UVPSpec` object without using weights. 
I'd be happy to talk more about how to do all of this properly!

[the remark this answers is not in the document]
adeliegorce: I think I'm going to remove the option to initialise the `UVWindow` object with a `UVData` object because I am not sure how useful it is in the end. What is really useful is the interface with `UVPSpec` indeed

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,946 @@
...
+            (in case they were computed with another taper)
+        """
+        self.taper = taper
+        if (len(self.cyl_wf)>0):
+            raise_warning('New taper set but window functions have not been updated.')
+            if clear_cache:
+                self.clear_cache(clear_cyl_bins=True)
+
+
+    def set_spw_range(self,spw_range):
```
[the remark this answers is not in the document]
adeliegorce: I've modified the class so that `spw_range` is now an attribute of the `UVWindow` object, that needs to be defined when initialising the object as @steven-murray was pointing out that redefining  the option allowing the user to change the spectral window of the object mid-course might be confusing to the user. The point of specifying the spectral window is that you do not have to load the array containing the FT of the beam on the whole bandwidth to do your computations (which is pretty heavy), but just a portion of it.
From what I've seen most of the `hera_pspec` code works by looping over `spw`, then `pol`, then time,... so I don't think we can avoid re-defining a new `UVWindow` object at each iteration of these loops as the FT_beam file loaded depends on the polarisation. In the new version of the code, it would be only one line of code. Does that seem reasonable to you?

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+            assert pols[0] and pols[1] in ['pI', 'pQ', 'pV', 'pU', 'xx', 'yy', 'xy', 'yx'],\
+                "Wrong polarisation"
+        # if tuple of polarisation integers
+        elif isinstance(pols[0], int):
+            assert pols[0] and pols[1] in [1, 2, 4, 3, -5, -6, -7, -8],\
+                "Wrong polarisation"
+            pols = (uvutils.polnum2str(pols[0], x_orientation=x_orientation),
+                    uvutils.polnum2str(pols[0], x_orientation=x_orientation))
+        else:
+            raise TypeError("Must feed pols as (tuple of) str or int.")
```
steven-murray, line 336: feels like a LOT of code repetition here... maybe just convert to tuple first, then iteratively test each element.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        if ftbeam_obj is None:
+            # create ftbeam_obj for first pol in the pair
+            self.ftbeam_obj_pol.append(FTBeam(self.pols[0], ftfile=ftfile, spw_range=spw_range,
+                                              verbose=verbose))
+            # copy object if polarisations in pair are identical
+            if self.pols[0] == self.pols[1]:
+                self.ftbeam_obj_pol.append(copy.deepcopy(self.ftbeam_obj_pol[0]))
+            # else, create new ftbeam object
+            else:
+                self.ftbeam_obj_pol.append(FTBeam(self.pols[1], ftfile=ftfile, spw_range=spw_range,
```
steven-murray, line 360: I feel like it would be a LOT simpler to just accept a tuple of `FTBeam` objects into the constructor here. Then you let the `FTBeam` object deal with validating its own input. It's not any harder to call, really. If you really really want, you can have a constructor on `UVWindow` like the following:

```python
class UVWindow:
    @classmethod
    def from_ftfile(cls, ftfile, pols, **kwargs):
        ft = [FTBeam(ftfile, pol=pol, **kwargs) for pol in pols]
       return cls(ft_beam=ft)
```

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+                                                  verbose=verbose))
+        # if fed, read inputs
+        else:
+            ftbeam_obj = ftbeam_obj if isinstance(ftbeam_obj, (list, tuple, np.ndarray)) else [ftbeam_obj, ftbeam_obj]
+            # if input is only one object
+            if (ftbeam_obj[0] == ftbeam_obj[1]):
+                # check if polarisations in pair are identical
+                assert ftbeam_obj[0].pol == self.pols[0] and ftbeam_obj[1].pol == self.pols[1],\
+                    'If feeding only one ftbeam_obj, polarisations in pair must be identical'
```
steven-murray, line 369: using the above method of ONLY accepting the ftbeam_obj tuple would mean you don't need this check -- don't accept pols, just get them from the ftbeam_obj directly.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+            # raise warning if ftfile was fed as it will be ignored
+            if ftfile is not None:
+                warnings.warn('Fed ftfile value will be overriden by input ftbeam_obj')
+                assert ftfile == ftbeam_obj[0].ft_file and ftfile == ftbeam_obj[1].ft_file,\
+                    "FT file of ftbeam_obj does not match ftfile of UVWindow"
```
steven-murray, line 374: also wouldn't need this check

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+            # read object to fill self.ftbeam_obj_pol list
+            for ib, ftbeam in enumerate(ftbeam_obj):
+                ftbeam = copy.deepcopy(ftbeam)
+                # assert properties of ftbeam object fed are consistent with other inputs
+                assert ftbeam.check(), "There is an issue with the ft_beam object."
+                assert ftbeam.pol == self.pols[ib],\
+                    "Polarisation of ftbeam_obj does not match polarisation of UVWindow"
```
steven-murray, line 381: or this check

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+                if spw_range is not None and ftbeam.spw_range != tuple(spw_range):
+                    ftbeam.update_spw(spw_range)
```
steven-murray, line 383: I'd just leave it up to the user to set the spw_range they want on the FTBeam object. Don't let them pass in a potentially different one here.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+                self.ftbeam_obj_pol.append(ftbeam)
```
steven-murray, line 384: why are you deepcopying this? Why not just use the ones passed in?

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        # define spectral window-related attributes
+        self.freq_array = np.copy(self.ftbeam_obj_pol[0].freq_array)
+        self.Nfreqs = len(self.freq_array)
+        self.dly_array = utils.get_delays(self.freq_array,
+                                          n_dlys=len(self.freq_array))
+        self.avg_nu = np.mean(self.freq_array)
+        self.avg_z = self.cosmo.f2z(self.avg_nu)
+
+    @classmethod
+    def __from_uvpspec__(cls, uvp, ftfile=None, ipol=0, spw=0,
```
steven-murray, line 400: make this just `from_uvspec`

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        cyl_wf = np.divide(cyl_wf, sum_per_bin, where=sum_per_bin != 0)
+
+        if (return_bins == 'unweighted') or return_bins:
+            return kperp_bins, kpara_bins, cyl_wf
+        elif (return_bins == 'weighted'):
+            return kperp, kpara, cyl_wf
+        else:
+            return cyl_wf
+
+    def cylindrical2spherical(self, cyl_wf, kbins, ktot, bl_lens,
```
steven-murray, line 941: I would call this `cylindrical_to_spherical` but that's personal taste

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+                                                            bl_lens, bl_weights)
+
+        if return_weights:
+            return wf_spherical, kweights
+        else:
+            return wf_spherical
+
+    def run_and_write(self, filepath, bl_lens, bl_weights=None,
+                      kperp_bins=None, kpara_bins=None,
+                      overwrite=False):
```
steven-murray, line 1206: HERA code typically seems to use the argument `clobber` instead of `overwrite`

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        # as a function of (kperp, kpara)
+        # the kperp and kpara bins are given as global parameters
+        cyl_wf = np.zeros((nbls, self.Nfreqs, nbins_kperp, nbins_kpara))
+        for ib in range(nbls):
+            cyl_wf[ib, :, :, :] = self.get_cylindrical_wf(bl_lens[ib],
+                                                          kperp_bins*self.kunits,
+                                                          kpara_bins*self.kunits)*bl_weights[ib]
+
+        # Write file
+        with h5py.File(filepath, 'w') as f:
```
steven-murray, line 1283: things like ints and float are typically better written into the `.attrs` group a dataset.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        except AttributeError:
+            raise AttributeError('Feed k array with units (astropy.units).')
+        assert self.kunits.is_equivalent(karray.unit),\
+            "k array units not consistent with little_h"
+
+
+def raise_warning(warning, verbose=True):
+    """Warning function."""
+    if verbose:
+        print('Warning: {}'.format(warning))
```
steven-murray, line 1320: i don't understand the point of this function... why not use the `logging` module? Or just do a `from warnings import warn`?

steven-murray (commented):
Thanks @adeliegorce, this is starting to really take shape. 

I have a few suggestions, most of them very simple. Let me know if you have any questions about any of them.

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+                to use, without the polarisation
+                Ex : ft_beam_HERA_dipole (+ path)
+                - None (default). Computation from beam simulations (slow).
+                Not yet implemented.
+        verbose : bool, optional
+            If True, print progress, warnings and debugging info to stdout.
+        x_orientation: str, optional
+            Orientation in cardinal direction east or north of X dipole.
+            Default keeps polarization in X and Y basis.
+            Used to convert polstr to polnum and convertly.
```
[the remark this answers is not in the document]
adeliegorce: I meant "conversely" actually! Made the change, thanks for spotting the typo

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        """
+        # verbose
+        try:
+            bool(verbose)
+        except ValueError:
+            raise ValueError("verbose must be boolean")
```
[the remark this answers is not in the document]
adeliegorce: Actually I ended up choosing this rather convoluted `try` statement because `bool(verbose)` very rarely raises an error. It basically interpret anything (let's say any reasonable input such as an array, a string, etc.) that is not zero as `True`... Do you think that is sufficient?

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+            raise ValueError('Wrong ftfile input. See docstring.')
+        else:
+            self.ft_file = ftfile
+            # obtain bandwidth in file to define spectral window
+            bandwidth = self.get_bandwidth()
+
+            # spectral window
+            if spw_range is not None:
+                # check format
+                spw_range = np.array(spw_range, dtype=int)
```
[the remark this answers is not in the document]
adeliegorce: Good point, done

hera_pspec/uvwindow.py
adeliegorce: This was to have the `verbose` option, that is if you run the `UVWindow` module with `verbose=False`, the messages are not printed to screen. But maybe it is not that useful - I switched to `warnings.warn`

hera_pspec/uvwindow.py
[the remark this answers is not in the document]
steven-murray: Yes, most things are convertible to bools in python, which is fine. But if `bool(verbose)` rarely raises an error, you're not achieving anything more by putting it in a try/except -- it still won't raise the error! You *could* do something like `if not isinstance(verbose, bool)` but I would *definitely* not recommend this. Just document that it should be a bool and convert it to such.

hera_pspec/uvwindow.py
steven-murray: yeah, if you really want to have the explicit "verbose" option, I'd just do:

```python
if verbose:
    warnings.warn("message")
```
in your main code, rather than calling another function to do it. But really, if you use `warnings.warn` (or even better, `logger.warn`) then the user can switch on/off the warnings (or other output) without passing that parameter around at all.

adeliegorce: Okay, that makes sense, thanks. None of the other `hera_pspec` modules use a logger so I guess I'll keep to `warnings.warn`

hera_pspec/uvwindow.py
```diff
@@ -0,0 +1,1320 @@
...
+        """
+        for p in ['pol', 'freq_array', 'mapsize', 'ft_beam']:
+            if not hasattr(self, p):
+                return False
+        if len(self.freq_array) <= 1:
+            return False
+        if len(self.ft_beam.shape) != 3:
+            return False
+
+        return True
```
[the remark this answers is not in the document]
adeliegorce: The idea was to check the correct format of any FTBeam object that is fed to UVWindow for example. Something that would work like `isinstance(input, FTBeam)`. Is there a clean way of doing this?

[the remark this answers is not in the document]
steven-murray: Yeah, so if your `FTBeam.__init__` has checks for these things, then you know that any `FTBeam` object you come across is "valid". Then in `UWindow` you just need to do `isinstance(ftbeam, FTBeam)` and you'll know it has the right structure.

hera_pspec/uvwindow.py
adeliegorce: I am not sure I understand. The idea here is to be able to parse the FTBeam directly to UVWindow so that, if you've already computed it, you don't have to do it again. Since I've now changed FTBeam to have two class methods `from_file` and `from beam` I think it's made what you say impossible because I need to know before initialising the `FTBeam` object if it going to be initialized with a ft_file or not

hera_pspec/uvwindow.py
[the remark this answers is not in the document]
adeliegorce: Okay thanks, wasn't sure `isinstance(ftbeam, FTBeam)` was doing this.

hera_pspec/uvwindow.py
adeliegorce: Okay, I think I got your point in the end. I've removed the option for `UVWindow` to be initialized with a ftfile object. It will always have to be fed one (or a tuple of) FTBeam object(s), it simplifies the code **a lot**

steven-murray: I'm not sure I understand the difficulty here (probably my own thickness...). I envisage a workflow like the following:

```python
ftbeam_x = FTBeam.from_file(ftfile, pol='x')
ftbeam_y = FTBeam.from_file(ftfile, pol='x')

uvw = UVWindow(ftbeams=(ftbeam_x, ftbeam_y))
```

Since the `ftbeam_x` object has the `pol`, `spw` etc. defined on it already, you don't then need to pass that to `UVWindow`, you just need to verify that the two beams passed in are compatible. Your `from_uvpspec` function I presume would essentially just implement the above workflow but getting the beams from the UVPSpec object.

Am I missing something important here?


```

Comment (steven-murray) [reply hera_pspec/uvwindow.py:360]: Oops sorry didn't see your last response :-)

Comment (adeliegorce) [reply hera_pspec/uvwindow.py:360]: No, I think I'm just trying to make the code more complicated than in needs to be (e.g. too many options for the user). Initializing the `UVWindow` attributes from the `FTBeam` ones seems like a good option

File: hera_pspec/uvwindow.py [hunk 22]
```diff
@@ -75,152 +78,153 @@ def __init__(self, pol, spw_range=None, ftfile=None,
...
+        x_orientation: str, optional
+            Orientation in cardinal direction east or north of X dipole.
+            Default keeps polarization in X and Y basis.
+            Used to convert polstr to polnum and conversely.
+        """
+        # check file path input
+        if not isinstance(ftfile, str):
+            raise TypeError('Wrong ftfile input. See docstring.')
+        if not os.path.isfile(ftfile):
+             raise ValueError("Cannot find FT beam file: {}.".format(ftfile))
```
steven-murray, line 145: We should probably support the input `ftfile` being a `pathlib.Path` instance. This would allow you to change this code to

```python
from pathlib import Path
...
def from_file(cls, ftfile, ...):
    ftfile = Path(ftfile)  # if it's a string or Path this will work.
    if not (ftfile.exists() and ftfile.is_file()):
        raise ValueError(...)

File: hera_pspec/uvwindow.py [hunk 23]
```diff
@@ -75,152 +78,153 @@ def __init__(self, pol, spw_range=None, ftfile=None,
...
+            ft_beam = ft_beam[spw_range[0]:spw_range[1], :, :]
+        else:
+            # if spectral range is not specified, use whole bandwidth
+            spw_range = (0, bandwidth.size)
+            freq_array = bandwidth
+
+        return cls(data=ft_beam, pol=pol, freq_array=freq_array,
+                   mapsize=mapsize, verbose=verbose, x_orientation=x_orientation)
+ 
+    def get_bandwidth(self, ftfile):
```
steven-murray, line 171: this can also be a classmethod (you never use "self").

hera_pspec/uvwindow.py
```diff
@@ -75,152 +78,153 @@ def __init__(self, pol, spw_range=None, ftfile=None,
...
-                                    "frequency indices."
-        assert spw_range[1]-spw_range[0] > 0, \
-            "Require non-zero spectral range."
-        assert min(spw_range) >= 0 and max(spw_range) < self.freq_array.size,\
-            "spw_range must be integers within the given bandwith."
-        assert self.ft_beam is not None,\
-            "The FT of the beam needs to have already been read or computed"
+        assert check_spw_range(spw_range, self.freq_array), \
+            "Wrong spw range format, see dosctring."
+        assert self.ft_beam is not None, "No ft_beam attribute."
```
steven-murray, line 215: I think this could never fail, right? Since ft_beam is always part of the instance now.

hera_pspec/uvwindow.py
```diff
@@ -305,90 +279,19 @@ def __init__(self, pols, spw_range=None, ftfile=None,
...
-            # assert different elements in lists have same properties
-            assert np.unique([ft.mapsize for ft in self.ftbeam_obj_pol]).size == 1,\
-                'FTBeam objects fed have different properties'
-            assert np.all(self.ftbeam_obj_pol[0].freq_array == self.ftbeam_obj_pol[1].freq_array),\
-                'FTBeam objects fed have different bandwidths'
-
-        # define spectral window-related attributes
+        self.ftbeam_obj_pol = list(ftbeam_obj) if np.size(ftbeam_obj) > 1 else [ftbeam_obj, ftbeam_obj]
+        assert isinstance(self.ftbeam_obj_pol[0], FTBeam) and isinstance(self.ftbeam_obj_pol[1], FTBeam), \
+            "Wronge input given in ftbeam_obj: must be (a list of) FTBeam object(s)"
```
steven-murray, line 286: typo in "wrong"

hera_pspec/uvwindow.py
```diff
@@ -305,90 +279,19 @@ def __init__(self, pols, spw_range=None, ftfile=None,
...
+        # check if elements in list have same properties
+        assert np.all(self.ftbeam_obj_pol[0].freq_array == self.ftbeam_obj_pol[1].freq_array), \
+            'Spectral ranges of the two FTBeam objects fed to not match'
```
steven-murray, line 289: "fed to" -> "do"

hera_pspec/uvwindow.py
```diff
@@ -305,90 +279,19 @@ def __init__(self, pols, spw_range=None, ftfile=None,
...
+        assert self.ftbeam_obj_pol[0].mapsize == self.ftbeam_obj_pol[1].mapsize, \
+            'Physical properties of the two FTBeam objects fed to not match'
```
steven-murray, line 291: "fed to" -> "do"

hera_pspec/uvwindow.py
```diff
@@ -452,26 +350,33 @@ def __from_uvpspec__(cls, uvp, ftfile=None, ipol=0, spw=0,
...
+                                                       x_orientation=x_orientation))
+            else:
+                ftbeam_obj_pol.append(FTBeam.from_file('{}_{}.hdf5'.format(ftfile, pol),
+                                                       spw_range=None,
+                                                       verbose=verbose,
+                                                       x_orientation=x_orientation))
+            if pol[1] == pol[0]:
+                # do not recompute if two polarisations are identical
+                ftbeam_obj_pol.append(copy.deepcopy(ftbeam_obj_pol[0]))
+                continue
```
steven-murray, line 368: I think this will add a third `ftbeam` to `ftbeam_obj_pol` in the case that `pol[1] == pol[0]`. You probably need something like 

```python
for ip, pol in enumerate(polpair):
    if ip > 0 and pol[ip] == pol[0]:
        ftbeam_obj_pol.append(ftbeam_obj_pol[0])
    else:
        if ftfile is None:
            ....

Comment (steven-murray) [reply hera_pspec/uvwindow.py:368]: Furthermore, not sure why you're deepcopying the ftbeam here -- it could be a waste of memory, unless you plan on users (or yourself) modifying the second beam in-place independently of the first beam. I would say this shouldn't be done, but there are cases in which it would be reasonable.

File: hera_pspec/uvwindow.py [hunk 29]
```diff
@@ -1278,24 +1163,24 @@ def run_and_write(self, filepath, bl_lens, bl_weights=None,
             cyl_wf[ib, :, :, :] = self.get_cylindrical_wf(bl_lens[ib],
                                                           kperp_bins*self.kunits,
                                                           kpara_bins*self.kunits)*bl_weights[ib]
+        # cannot write None to file
+        if self.taper is None:
+            self.taper = 'none'
```
steven-murray, line 1168: I would not set this here, it would be surprising to the user. If you want to ensure the taper is always a string "none" instead of None, then you should validate that in the `__init__`. Otherwise when you *use* the taper you could just send in `self.taper or 'none'` instead of simply `self.taper`.

steven-murray (commented):
So close! This is looking really really nice. I think the last few things will be easy :-)

hera_pspec/uvwindow.py
adeliegorce: But if `Path` only takes a string or a path as input, won't I need a `isinstance(ftfile, str)` anyway to avoid that `Path(ftfile)` raises an error?

hera_pspec/uvwindow.py
adeliegorce: I think it could. Imagine you've initialized a `FTBeam` object on a reduced bandwidth and try to reduce it even more.

I know you don't like the `update_something` functions but this one is necessary inside `from_uvpspec` because of the limited spectral info available inside the `UVPSpec` objects

hera_pspec/uvwindow.py
adeliegorce: The `deepcopy` is here because when I later run `update_spw` on the first `ftbeam_obj_pol`, without the `deepcopy`, it also updates the frequency array of the second `ftbeam_obj_pol`. Which is what we want to achieve in the end but I was worried the fact it is only a pointer might mess things up further along the way

hera_pspec/uvwindow.py
adeliegorce: Went for the second option!

hera_pspec/uvwindow.py
steven-murray: Firstly, to the point of using it: it's becoming the standard way to deal with file paths in python, so it's not unlikely that someone will want to pass in a `Path` instance as the `ftfile`. They can always pass in `str(ftfile)` to convert it to string, but I think that covering both cases (since it's easy enough) is more future-proof.

Secondly, no you don't need to do the check. Doing `Path(ftfile)` will raise an error if `ftfile` is not str or Path, but it will be a nice understandable error that the user will be then able to fix. I.e. calling `Path(ftfile)` essentially *does* the type-checking for you.

hera_pspec/uvwindow.py
steven-murray: You're right to be cautious here (and things like this are why in general I like to have classes that are immutable, i.e. no `update_xxx`, wherever possible). I'll leave it up to your judgment, but I think that given there's just the one `update` function, and that you can dissuade users from updating any other attributes, that this can be a pointer.

hera_pspec/uvwindow.py
adeliegorce: Oh I misunderstood your comment and thought you were referring to `assert check_spw_range(spw_range, self.freq_array)`. You are right, there is no reason for `self.ft_beam` to be None in the new version of the `FTBeam` class.

hera_pspec/uvwindow.py
adeliegorce: You convinced me!

aewallwi (approved):
It looks like all of my concerns have been addressed. I'm happy to approve this.

adeliegorce:
Hi!

Here is the second major window function-related PR. It introduces functions that link between the `UVWindow` and the `UVPSpec` classes. 

The new additions in the `UVPSpec` class are:
- A new boolean `exact_windows` attribute for `UVPSpec` objects. If not set because an older version of `UVPSpec` was used to create the object, it is automatically set to `False`.
- A new `uvp.get_exact_window_functions` method which reads the information stored in the `uvp` object, as well as a `ftbeam_file` given as input to compute exact window functions. The result is stored in three attribute dictionaries: `uvp.window_function_array`, `uvp.window_function_kperp_bins` and `uvp.window_function_kpara_bins` -- the latter two being necessary to keep track of the $(k_\perp, k_\parallel)$ grid the window functions were computed over.
- The `uvp.get_window_function` method has been updated to be compatible with exact window functions. No weights are applied here.
The new additions in the `grouping` module are:
- The `grouping.average_spectra` and `grouping.spherical_average` methods have been updated to accurately deal with window functions if `uvp.exact_windows == True`.
- A new method, `grouping.spherical_wf_from_uvp` has been added to obtain the spherical average of the exact window functions stored in a `UVPSpec` object for a given set of k-bins.
All above methods handle `error_weights` and `stats_array` properly.

For more details, you can refer to `examples/UVWindow_tutorial.ipynb`.
I will be working on a few things in parallel to your review, that is:
- Add an option for exact window functions computations within the `pspecdata` module and the `run_pspec` function.
- Update docstring.
- Appropriate handling of exact window functions within `uvp.write_hdf5`, `uvp.read_hdf5` etc.
Happy to hear any other suggestion!

Looking forward to your review.

steven-murray (commented):
@adeliegorce  -- I haven't looked at everything, but my initial look seems to suggest the call in `UVPSpec` is not correct.

hera_pspec/uvpspec.py
```diff
@@ -1616,6 +1630,293 @@ def set_cosmology(self, new_cosmo, overwrite=False, new_beam=None,
...
+        # Copy these, so we don't modify the input lists
+        blpair_groups = copy.deepcopy(blpair_groups)
+        blpair_weights = copy.deepcopy(blpair_weights)
+
+        # sets attribute exact_windows to False if not defined
+        # (UVPspec object created with older versions of hera_pspec)
+        try:
+            self.exact_windows
+        except AttributeError:
+            self.exact_windows = False
```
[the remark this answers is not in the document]
adeliegorce: That is to deal with the issue of a `UVPSpec` object which would have been created with an earlier version of pspec (saved to file, then loaded) and which wouldn't have the exact_windows attribute. It'll probably get sorted when I have a look into the save and load functions in the `UVPSpec` module

hera_pspec/uvpspec.py
```diff
@@ -1616,6 +1630,293 @@ def set_cosmology(self, new_cosmo, overwrite=False, new_beam=None,
...
+                                ' all baseline pairs in object.')
+            # use spw given
+            spws = np.array([this_spw])
+
+        # Create new window function array
+        window_function_array = odict()
+        window_function_kperp_bins, window_function_kpara_bins = odict(), odict()
+
+        # initialise UVWindow object
+        uvw = UVWindow(ftbeam=ftbeam_file, taper = self.taper,
```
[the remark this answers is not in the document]
adeliegorce: This version is not the latest version on the branch.. if you go to the "Files changed" tab above, it is a different version. I'm sorry, I should have done a whole other PR really...

hera_pspec/uvpspec.py
```diff
@@ -1616,6 +1630,142 @@ def set_cosmology(self, new_cosmo, overwrite=False, new_beam=None,
...
+            Spectral window index. If None, the window functions are computed on 
+            all the uvp.spw_ranges, successively. Default: None.
+
+        verbose : bool, optional
+            If True, print progress, warnings and debugging info to stdout.
+
+        inplace : bool, optional
+            If True (default value), the UVPspec attribute window_function_array is filled with the
+            values computed in the function, and window_function_kperp_bins and
+            window_function_kpara_bins array are added as attributed.
```
steven-murray, line 1664: typo in attributes

hera_pspec/uvpspec.py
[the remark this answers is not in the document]
steven-murray: Yeah, I think the proper place to do this is in the `load` function (or whatever it's called).

hera_pspec/uvpspec.py
```diff
@@ -1616,6 +1630,142 @@ def set_cosmology(self, new_cosmo, overwrite=False, new_beam=None,
...
+
+        blpair_groups, blpair_lens, _ = self.get_red_blpairs()
+
+        # check spw input and create array of spws to loop over
+        if this_spw is None:
+            # if no spw specified, use attribute
+            spws = np.arange(self.Nspws)
+        else:
+            # check if spw given is in uvp
+            assert this_spw in self.spw_array, "input spw is not in UVPSpec.spw_array."
```
steven-murray, line 1695: Shouldn't you allow to set multiple spws since you obviously handle the case of multiple spws?

steven-murray (commented):
Thanks @adeliegorce, this looks about right! In saying that, there's a lot in this PR and I wouldn't be surprised if I'm missing something. But the guts of it looks right to me. I made only a few very small comments that could be fixed up then I'll approve :-)

hera_pspec/uvpspec.py
adeliegorce: Done :)

adeliegorce:
Hi! 
I've made the changes @steven-murray requested. I've also modified some functions that I hadn't updated to handle exact window functions so far (`uvp.fold_spectra`, `combine_uvpspec`, `uvp.read_hdf5` and `uvp.write_hdf5` and added corresponding tests. I just want to point out that the output of the `uvp.get_exact_window_functions` method will always be normalized so that for each (b, tau) pair, sum_(kperp, kpara) W(kperp, kpara) = 1; to apply weights, one must use `uvp.average_spectra`. It might be important to keep in mind for the likelihood development.
Thank you for your reviews!

steven-murray (approved):
I think this looks all good, thanks @adeliegorce for the huge amount of effort here!

PR #353, opened 2022-03-18
Fix issue with git clone in setup.py

adeliegorce:
Replaced `git clone git://github.com/HERA-Team/uvtools` calls with `git clone https://github.com/HERA-Team/uvtools` to allow checks to run properly. Otherwise, throws git error.


adeliegorce:
@nkern I am not authorized to merge the PR, would you mind doing it? thanks!

PR #355, opened 2022-05-06
Handle single time noise files

aewallwi:
Small modification to `uvp_noise_errors` to allow for noise calculation in files with a single time.


philbull (approved):
Looks good to me!

PR #357, opened 2022-06-07
Update on the UVWindow and FTBeam modules

adeliegorce:
PR including 
- Various usage-induced fixes for the `UVWindow` and `FTBeam` modules, 
- An option to define the Fourier transform of the beam (`FTBeam` object) from a Gaussian beam with the `FTBeam.gaussian` class method.
- Related tests

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -1,6 +1,6 @@
 import numpy as np
 from collections import OrderedDict as odict
-import os, copy, shutil, operator, ast, fnmatch, sys
+import os, copy, shutil, operator, ast, fnmatch
 from pyuvdata import utils as uvutils
 import h5py
 import warnings
@@ -1726,7 +1726,7 @@ class UVPSpec(object):
                 # Iterate over baseline-pair groups
                 for j, blpg in enumerate(blpair_groups):
                     if verbose: 
-                        sys.stdout.write('\rComputing for bl group {} of {}...'.format(j+1, len(blpair_groups)))
+                        print('\rComputing for bl group {} of {}...'.format(j+1, len(blpair_groups)), end='')
 
                     # window functions identical for all times
                     window_function_blg = uvw.get_cylindrical_wf(blpair_lens[j],
@@ -1743,7 +1743,7 @@ class UVPSpec(object):
                             pol_window_function[iblts, :, :, :] = np.copy(window_function_blg)
 
                 if verbose: 
-                    sys.stdout.write('\rComputed wf for baseline-pair groups {} of {}.\n'.format(len(blpair_groups),len(blpair_groups)))
+                    print('\rComputed wf for baseline-pair groups {} of {}.'.format(len(blpair_groups),len(blpair_groups)))
 
                 # Append to lists (spectral window)
                 spw_window_function.append(pol_window_function)
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -256,16 +256,13 @@ class UVPSpec(object):
 
         spw, blpairts, polpair = self.key_to_indices(key, omit_flags=omit_flags)
 
-        if self.exact_windows:
-             return self.window_function_array[spw][blpairts, :, :, :, polpair]
-
         # Need to deal with folded data!
         # if data has been folded, return only positive delays
         if self.folded:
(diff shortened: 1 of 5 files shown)


hera_pspec/uvpspec.py
```diff
@@ -1721,17 +1718,15 @@ def get_exact_window_functions(self, ftbeam_file=None, spw_array=None,
...
                 # lengths given as input
                 kperp_bins = uvw.get_kperp_bins(blpair_lens)
                 kpara_bins = uvw.get_kpara_bins(uvw.freq_array)
                 pol_window_function = np.zeros((self.Nblpairts, self.get_dlys(spw).size, kperp_bins.size, kpara_bins.size))
-
                 # Iterate over baseline-pair groups
                 for j, blpg in enumerate(blpair_groups):
                     if verbose: 
-                        sys.stdout.write('\rComputing for bl group {} of {}...'.format(j+1,len(blpair_groups)))
+                        sys.stdout.write('\rComputing for bl group {} of {}...'.format(j+1, len(blpair_groups)))
```
steven-murray, line 1729: why is this not just a print statement, or `logging` statement?

hera_pspec/uvwindow.py
```diff
@@ -168,6 +168,78 @@ def from_file(cls, ftfile, spw_range=None, verbose=False, x_orientation=None):
...
+    def gaussian(cls, freq_array, widths, pol, 
+                 mapsize=1.0, npix=301,
+                 cosmo=conversions.Cosmo_Conversions(),
+                 verbose=False, x_orientation=None):
+        """
+        Read Fourier transform of beam in sky plane from file.
+
+        Initialise FTBeam object by reading file containing the 
+        Fourier transform of the  instrument beam in the sky plane 
+        for given frequencies.
```
steven-murray, line 181: looks like this wasn't updated?

hera_pspec/uvwindow.py
```diff
@@ -168,6 +168,78 @@ def from_file(cls, ftfile, spw_range=None, verbose=False, x_orientation=None):
         return cls(data=ft_beam, pol=pol, freq_array=freq_array,
                    mapsize=mapsize, verbose=verbose, x_orientation=x_orientation)
  
+    @classmethod
+    def gaussian(cls, freq_array, widths, pol, 
+                 mapsize=1.0, npix=301,
+                 cosmo=conversions.Cosmo_Conversions(),
+                 verbose=False, x_orientation=None):
```
steven-murray, line 175: For all the parameters that are literally just passed through to the main constructor, I would just set them as `**kwargs` and point in the docs to the main constructor's documentation.

hera_pspec/uvwindow.py
```diff
@@ -754,8 +827,8 @@ def get_cylindrical_wf(self, bl_len,
         else:
             self.check_kunits(kperp_bins)
         kperp_bins = np.array(kperp_bins.value)
-        if not np.isclose(np.diff(kperp_bins),np.diff(kperp_bins)[0]).all():
-            warnings.warn('kperp_bins must be linearly spaced.')
+        if not np.isclose(np.diff(kperp_bins), np.diff(kperp_bins)[0]).all():
+            warnings.warn('get_cylindrical_wf: kperp_bins must be linearly spaced.')
```
steven-murray, line 831: why does this have to be true? And if so, why is it just a warning, not an error?

steven-murray (approved):
Thanks @adeliegorce, other than a few minor comments I think this is good to go.

hera_pspec/uvpspec.py
adeliegorce: Changed to print

hera_pspec/uvwindow.py
adeliegorce: That is because the edges of each bin are obtained assuming a linear spacing, e.g.:
`kperp_bin_edges = np.arange(kperp_bins.min()-dk_perp/2, kperp_bins.max() + dk_perp, step=dk_perp)`.
It can be modified in the future but I don't think handling log-spaced bins is necessary at the moment (the data is linearly spaced).

adeliegorce: Yes I changed for a `ValueError`! Sorry, forgot to mention

steven-murray: 👍🏻

PR #360, opened 2022-07-05
fix: circular import with __version__

steven-murray:
Fixes a circular-import issue when importing `hera_pspec`. The init was importing a submodule which was importing `__version__` from the init, but `__version__` hadn't yet been defined in the init...


Zhilei:
Thanks for the PR! Looks good to me.

nkern (approved):
looks good, thanks SM!

PR #361, opened 2022-07-07
Issue in little_h homogeneity when using grouping.spherical_average

### LINKED ISSUE #367: Ensure little_h consistency throughout
It would be good to add units to objects tracing Fourier modes, e.g. with `astropy.cosmology.units.littleh`. This would ensure little_h consistency throughout the code. Note that, right now, `grouping.spherical_average` assumes `little_h=True` and there is no way around it / no way to check consistency with the spherical k-bins given as input.

adeliegorce:
There was an issue in `grouping.spherical_average`: if keyword `little_h` is `False` in function, then the input k-bins are automatically adjusted to be in h Mpc-1 units:
```
# transform kgrid to little_h units
 if not little_h:
      kbins = kbins / uvp.cosmo.h
      bin_widths = bin_widths / uvp.cosmo.h
```
However this change occurred _after_ defining `kbin_left` and `kbin_right` that are later used to bin. I changed this and moved the above code block up.


hera_pspec/grouping.py
```diff
@@ -1024,6 +1025,21 @@ def spherical_wf_from_uvp(uvp_in, kbins, bin_widths,
 
     if isinstance(bin_widths, (float, int)):
         bin_widths = np.ones_like(kbins) * bin_widths
+        
+    # if window functions have been computed without little h
```
steven-murray: Hmmm, I'm not sure I'm totally comfortable with this logic haha.

I think what I'd do is not accept `little_h`, and instead check here if littleh is included in the UVP object. If it is, then kbins are interpreted as being in h/Mpc, and vice versa. Thus, it's up to the user to pass the correct kbins given they know what the units of the pspec are, rather than playing around with little h throughout the function.

In the end, we should just use astropy units everywhere and let it handle all the conversions properly.

adeliegorce:
This was implemented as part of #428

PR #362, opened 2022-07-11
Fix workflow to run checks at each PR

adeliegorce:
PR to make sure all checks are now running when someone starts a PR...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,13 @@ jobs:
       run: |
         export PATH="$HOME/miniconda/bin:$PATH"
         ./ci/install_conda.sh
+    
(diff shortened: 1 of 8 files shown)


adeliegorce:
Here are some details on the changes made:
- The checks were not running because `.github/workflows/ci.yml` referred to the `master` branch instead of `main`
- The `version` module has been largely modified by a previous PR and does not contain the attribute `git_hash` anymore. I have replaced the occurrences of `version.git_hash` by `__version__` while ensuring backward compatibility with a `try` statement (note: `git_hash` is called when saving history strings in `UVPSpec` objects)
- The required dependencies (`pyuvdata`, `hera_cal`, `uvtools`) are now install directly with `pip install hera_calibration` instead of calling their git repository.

PR #366, opened 2022-08-31
Fix numpy warnings

### LINKED ISSUE #348: numpy errors can unsettle the user
I've noticed hera_pspec and other packages generate some numpy errors that are probably not important but the user might think something has gone wrong. An example in hera_pspec is when running pspec() in PSpecData():

```
uvp = ds.pspec(baselines1, baselines2, dsets=(0, 1), pols=[(pol, pol)], spw_ranges=[spw_range], verbose=False,
              input_data_weight='identity')
```

gives
```
       Casting complex values to real discards the imaginary part
```
The output seems fine so should I worry about this or not? It's not clear because I know other cases where I can ignore numpy errors (found by checking code or asking someone). It would be better if the errors that can be ignored didn't appear.

steven-murray:
Removes `np.int`, `np.str`, `np.float` and `np.complex` and replaces them with their builtin counterparts.

Partially fixes #348 (but doesn't deal with the specific example raised there)


hera_pspec/uvpspec.py
```diff
@@ -2037,7 +2037,7 @@ def generate_noise_spectra(self, spw, polpair, Tsys, blpairs=None,
         for i, blp in enumerate(blpairs):
             # get indices
             inds = self.blpair_to_indices(blp)
-            assert isinstance(Tsys[blp], (float, np.float, int, np.int)) \
+            assert isinstance(Tsys[blp], (float, int, int)) \
```
adeliegorce, line 2040: I think you mean `(float, int)`

hera_pspec/tests/test_uvpspec_utils.py
```diff
@@ -152,7 +152,7 @@ def test_get_red_blpairs():
     assert len(blps) == len(angs) # Ditto, for angles
 
     # Check output type
-    assert isinstance(blps[0][0], (np.int, int))
+    assert isinstance(blps[0][0], (int, int))
```
adeliegorce, line 155: I think you mean `isinstance(blps[0][0], int)`

adeliegorce (commented):
Hey, thank you for this, it is useful. I spotted two typos but apart from that, if the checks run, happy to merge!

steven-murray:
Thanks @adeliegorce, I fixed those :-)

adeliegorce (approved):

PR #368, opened 2022-12-16
Enhancement of the `UVWindow` module

adeliegorce:
This PR inludes several small changes to the `UVWindow` module, which should all be backward-compatible. Namely:
- Applied `nan_to_num` function to weights in `grouping.spherical_average` to ensure there are no NaN weights (the real parts are never NaN but the imaginary part is sometimes)
- The way to feed a beam model to the `UVWindow.from_uvpspec` method is now through the `ftbeam` key (previously `ftfile`). It can the path to a beam simulation (`Path` or `str`) or a `FTBeam` object directly. This facilitates using Gaussian beam models.
- Added related tests.


adeliegorce:
NB: This update is required for the PR #49 "Cylindrical window functions implementation" of `pspec_likelihood` (see [https://github.com/HERA-Team/pspec_likelihood/pull/49](link)) to be merged (checks fail without).

steven-murray (approved):
These changes all look good to me, thanks @adeliegorce !

PR #370, opened 2022-12-29
Doc files for PyPi installation

adeliegorce:
This PR includes new doc files to account for the new possibility to install `hera_pspec` directly from PyPi:
- The main `README` has been updated from Markdown to ReStructuredText (RST) which is more easily read by PyPi
- The doc files for readthedocs have been slightly changed, and a new `Installation` section has been added.


README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+New versions are frequently released on PyPi.
+
+For developers
+--------------
+
+If you are developping and/or want to be use the latest working version
+of ``hera_pspec``, you can directly install from the GitHub repository.
+
+Preferred method of installation for users is simply ``pip install .``
```
steven-murray, line 39: presume this line should be "developers" instead of "users"

docs/index.rst
```diff
@@ -1,19 +1,41 @@
+**************
```
steven-murray, line 1: since this is essentially the README, you should just `.. include::` it (you can even set marks where you want the include to start and end)

docs/installation.rst
```diff
@@ -0,0 +1,65 @@
+Installation
```
steven-murray, line 1: same here

steven-murray (commented):
Thanks @adeliegorce, this is great. I made a couple of suggestions, but otherwise should be good to go.

adeliegorce:
@steven-murray it seems like the tests are failing because of the latest version of `hera_cal` (I tried with the PyPi distribution and the `main` branch of the repo)... Any idea what changed?
If you want to have a look, the error is triggered when running `pytest test_grouping.py` in `hera_pspec` and comes from a `ValueError` in `hera_cal/utils.py`

steven-murray:
@adeliegorce hmm I see. I made https://github.com/HERA-Team/hera_cal/pull/864 to try and fix this.

steven-murray:
@adeliegorce can you try running these tests on the new branch of hera_cal -- that'll help us know if that branch actually fixes the problem.

adeliegorce:
There is another `ValueError`. I am having some trouble tracing it back but it seems like `hera_cal.io.load_vis` outputs arrays of a different shape than it used to. Does that make sense considering you've modified `hera_cal.utils.lst_rephase`?

adeliegorce:
Okay it has to be related because the method failing is `pspecdata.rephase_to_dset`

steven-murray (approved):

PR #371, opened 2023-01-19
Check spws up front

steven-murray:
Just checks that input spw's to the pspec_run function are valid, before reading too much stuff in, and print out helpful info if it fails.


adeliegorce (commented):
This is useful but it makes more sense to me to have these checks in `pspec` than `pspec_run` as there are already some checks for `spw_ranges` there.

steven-murray:
@adeliegorce that was my first thought as well, and maybe that's the right thing to do (I don't know hera_pspec well). However, I changed my mind and put it in `pspec_run` because if it's in `pspec` it's already done a bunch of stuff and taken a while before you know that things are wrong. It turns out to be easy to get the spws wrong from makeflow because you previously select other spws  that can conflict with the pspec ones.

adeliegorce:
Okay, makes sense!

adeliegorce (approved):

PR #372, opened 2023-01-19
ci: faster env setup

steven-murray:
Makes the Python setup much faster, and moreover tests what an actual user will actually install if they do `pip install hera_pspec`.


adeliegorce (approved):
Seems good to me. Let's maybe keep the PR open after merging to remember to change back `git+https://github.com/hera-team/hera_cal@fix-rephase` to `hera_cal`.

setup.cfg
```diff
@@ -46,7 +46,7 @@ install_requires =
     pyyaml
     h5py
     uvtools
-    hera-calibration
+    hera-calibration>=3.2.2
```
adeliegorce, line 49: Can you explain your thought process here quickly? Newer versions of `hera_cal` have backwards compatibility for the array shapes, but shouldn't it be the other way around?

steven-murray: hera_cal took out support for old array shapes semi-recently. The newest version adds support for old array shapes back in (to some parts of the code, relevant for pspec)

PR #374, opened 2023-01-20
Changed test beamfile to have only float data

adeliegorce:
Had to modify a beam file used in tests because the `data_array` was filled with complex numbers, which `pyuvdata` does not support anymore (release 2.2.12). 
If tests are still failing, it is because of the future array shapes in `pyuvdata`. With this PR, the tests triggering the error 
`ValueError: UVParameter _data_array is not the appropriate type. Is:  <class 'complex'>. Should be: <class 'float'>.`
are fixed but the `ValueError: non-broadcastable output operand` remain.


steven-murray:
@adeliegorce I merged the fix for the hera_cal stuff, but this branch looks like it's still failing the pyuvdata stuff?

adeliegorce:
It is failing because there is a requirement `pyuvdata<=2.2.8`. At this time `pyuvdata` wanted complex data arrays but newer versions require float data arrays. Runs are passing locally without this PR and with the updated `hera_cal`. Not sure why it was failing before?

steven-murray:
@adeliegorce ah I see. You probably have a newer version of pyuvdata locally, which works with this. Let's relieve the version of pyuvdata on the branch and see...

steven-murray (approved):
Looks good!

PR #375, opened 2023-01-24
Automate release to PyPi

adeliegorce:
This PR adds a file to the workflow to automate the release of versions to PyPi every time a new tag is defined.


steven-murray (approved):
Looks good!

PR #377, opened 2023-01-26
Fix bug in log messages.

aewallwi:
Errors are being caused by log message line trying to index with upper end of spw_range. The spw-ranges are meant to be inclusive for the base index and exclusive for the end-index (hence being allowed to range up to Nfreqs). Thus we should actually print spw[1] - 1 in the log-error message. If spw[1] = Nfreqs then an error will occur.


steven-murray (approved):
Oh, I should have thought about this!

PR #378, opened 2023-01-29
Now use time_1 and time_2 arrays to determine time overlap.

aewallwi:
By using time_1_array and time_2_array to define a unique blpairtime, this PR allows us to store different interleaved time sets within a single uvpspec object.


aewallwi:
This has been superceded by a different PR #379

PR #379, opened 2023-02-01
Unique time-keys  in `UVPSpec` objects based on `time_1_array` and `time_2_array` instead of the current `time_avg_array`

aewallwi:
This PR addresses a current limitation in UVPspec class in that it addresses unique data samples with baseline pair and the average time of the two samples being crossed for a power spectrum sample. A common use case that comes up is when we cross two different times for the same baseline pair. For a concrete example, consider the baseline pair `(bl_0, bl_1)` where each baseline is sampled at two close-together times `t1, t2`. It is common practice to form interleaves between the times so that we form the baseline-time pairs, `(bl_0, t1) x conj((bl_1, t2))` and `(bl_0, t2) x conj((bl_1, t1))` Both of these power spectra have the same average-time (`t1/2 + t2/2`) and as a result, they cannot actually be stored in the same `UVPSpec` object. Currently `pspec_run` is forced to compute these interleaves from separate files and store them as different `UVPSpec` objects in a `PSPecContainer`. Unfortunately, none of the data-averaging tools in `hera_pspec` that rigorously track error statistics, etc, can operate on multiple `UVPSpec` objects.

I saw two potential solutions to this problem.  

(1) Change the definition of `UVPSpec` to allow for  time-permuted blt-pairs to be stored in the same object.
(2) Change the averaging functions to operate on multiple UVPSpec objects. 

This PR addresses the problem through the first option. I decided to take the first approach because 

1) **Option(1) will result in a simpler interfaces than option (2).** interleaving in time is done routinely in power spectrum estimation and I think that supporting it at the level of the `UVPSPec` class will simplify our interfaces for this common task by allowing us to calculate arbitrary time-interleaves on arbitrary numbers of input data files, avoiding additional up-stream pipeline tailoring. This PR does not add such functionality to the `pspecdata.pspec_run` routines changing the `UVPSPec` definition is necessary to support this. 

2) **Option (1) requires less additional code then option (2).** My initial hunch is that option 2 would result in more _additional_ code since it would mean writing entire new versions of all of our averaging functions to operate over multiple objects (rather then axes in the data arrays of one of these objects). Option 1 (chosen here) focuses more on making point changes to _existing_ code which in most cases involve swapping variable names. 

* We want to key our datapoints with baseline and time pairs instead of baseline pairs and a single average time. Thus, this PR changes the parameter `Nblpairts` to `Nbltpairs`.
* We introduce a new variable, `Ntpairs` to represent the total number of unique time_1, time_2 pairs.
* We change the meaning of `Ntimes` in UVPspec objects to reference the total number of unique times in `time_1_array` union `time_2_array`. Before, it only referenced the total number of unique times in `lst_average_array`.  

With these changes, we can apply all of the averaging infrastructure and error bar propagation that exists for individual UVPspec objects to arbitrary time interleaves.

This PR also manages backwards compatibility with `UVPSPec` files created with previous versions of `UVPSpec` by detecting the deprecated `Nblpairts` variable when reading files and computing the new required variables, even if they are not present in these files from outdated versions.

### DIFF
--- a/hera_pspec/plot.py
+++ b/hera_pspec/plot.py
@@ -501,7 +501,7 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='abs-real',
         fig, axes = plt.subplots(Nside, Nside, figsize=figsize)
     
     # Ensure axes is an ndarray
-    if isinstance(axes, matplotlib.axes._subplots.Axes):
+    if isinstance(axes, matplotlib.axes.Axes):
         axes = np.array([[axes]])
     if isinstance(axes, list):
         axes = np.array(axes)
--- a/hera_pspec/plot.py
+++ b/hera_pspec/plot.py
@@ -384,6 +384,7 @@ def delay_waterfall(uvp, blpairs, spw, pol, component='abs-real',
     """
     import matplotlib
     import matplotlib.pyplot as plt
+    import matplotlib.axes
 
     # assert component
     assert component in ['real', 'abs', 'imag', 'abs-real', 'abs-imag'], "Can't parse specified component {}".format(component)
--- a/hera_pspec/utils.py
+++ b/hera_pspec/utils.py
@@ -1333,7 +1333,7 @@ def uvd_to_Tsys(uvd, beam, Tsys_outfile=None):
 
     return uvd
 
-def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_correction=True,  num_steps_scalar=2000, little_h=True):
+def uvp_noise_error(uvp, auto_Tsys=None, auto_Tsys_int=None, err_type='P_N', precomp_P_N=None, P_SN_correction=True,  num_steps_scalar=2000, little_h=True):
     """
     Calculate analytic thermal noise error for a UVPSpec object.
     Adds to uvp.stats_array inplace.
@@ -1345,10 +1345,12 @@ def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_
         If err_type == 'P_SN', uvp should not have any
         incoherent averaging applied.
 
-    auto_Tsys : UVData object, optional
+    auto_Tsys : UVData object.
         Holds autocorrelation Tsys estimates in Kelvin (see uvd_to_Tsys)
         for all antennas and polarizations involved in uvp power spectra.
         Needed for P_N computation, not needed if feeding precomp_P_N.
+        If power spectra were computed from interleaved times then the supplied auto_Tsys should include all of the times that are pesent in all of the pairs.
+
 
     err_type : str or list of str, options = ['P_N', 'P_SN']
         Type of thermal noise error to compute. P_N is the standard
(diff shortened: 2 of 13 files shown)


hera_pspec/utils.py
```diff
@@ -1333,7 +1333,7 @@ def uvd_to_Tsys(uvd, beam, Tsys_outfile=None):
 
     return uvd
 
-def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_correction=True,  num_steps_scalar=2000, little_h=True):
+def uvp_noise_error(uvp, auto_Tsys=None, auto_Tsys_int=None, err_type='P_N', precomp_P_N=None, P_SN_correction=True,  num_steps_scalar=2000, little_h=True):
```
philbull, line 1336: Why the new kwarg? Needs a docstring too.

hera_pspec/utils.py
```diff
@@ -1345,10 +1345,12 @@ def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_
         If err_type == 'P_SN', uvp should not have any
         incoherent averaging applied.
 
-    auto_Tsys : UVData object, optional
+    auto_Tsys : UVData object.
         Holds autocorrelation Tsys estimates in Kelvin (see uvd_to_Tsys)
         for all antennas and polarizations involved in uvp power spectra.
         Needed for P_N computation, not needed if feeding precomp_P_N.
+        If power spectra were computed from interleaved times then the supplied auto_Tsys should include all of the times that are pesent in all of the pairs.
```
philbull, line 1352: Typo: pesent

philbull (changes requested):
Looks good to me. Only minor changes requested. Because we're changing the file format spec, we should do a new minor version release when this is merged, so please add the relevant lines to the changelog, version number bump etc.

hera_pspec/utils.py
aewallwi: Fixed

hera_pspec/utils.py
aewallwi: I think this is an artifact of an idea that I never ended up following through on. I don't remember what it is and it doesn't look like this kwarg is called anywhere else in hera_pspec so I've removed it.

aewallwi:
Thanks for the comments @philbull. I think I've addressed everything so this is ready for another look :)

hera_pspec/plot.py
```diff
@@ -1076,4 +1077,4 @@ def _round_sigfig(x, up=True):
     if up:
         return np.ceil(10**sigfigs * x) / 10**sigfigs
     else:
-        return np.floor(10**sigfigs * x) / 10**sigfigs
\ No newline at end of file
+        return np.floor(10**sigfigs * x) / 10**sigfigs
```
philbull, line 1080: Line not tested according to codecov

hera_pspec/uvpspec.py
```diff
@@ -1382,6 +1390,33 @@ def read_from_group(self, grp, just_meta=False, spws=None, bls=None,
...
+            self.Ntpairs = len(set([(t1, t2) for t1, t2 in zip(self.time_1_array, self.time_2_array)]))
+
+        # Clear deprecated attributes.
+        for dattr in self._meta_deprecated:
+            if hasattr(self, dattr):
+                to_delete = getattr(self, dattr)
+                del to_delete
+
+        for dattr in self._meta_dsets_deprecated:
+            if hasattr(self, dattr):
```
philbull, line 1416: Not covered by tests according to codecov

hera_pspec/uvpspec.py
```diff
@@ -1341,14 +1349,14 @@ def read_from_group(self, grp, just_meta=False, spws=None, bls=None,
...
+            if k in self._meta_attrs or k in self._meta_deprecated:
                 val = grp.attrs[k]
                 if isinstance(val, bytes): val = val.decode() # bytes -> str
                 setattr(self, k, val)
         for k in grp:
-            if k in self._meta_dsets:
+            if k in self._meta_dsets or k in self._meta_dsets_deprecated:
                 setattr(self, k, grp[k][:])
 
         # Backwards compatibility: pol_array exists (not polpair_array)
```
philbull, line 1362: Not covered by tests according to codecov

philbull (changes requested):
Looks good. I managed to fix the tests; some API churn in matplotlib seems to be the cause of the issue.
Now the tests are passing, we have a regression in unit test coverage. I've added comments in 3 places where test coverage is lacking.

hera_pspec/uvpspec.py
aewallwi: The reason this line is never covered is because there are not yet any items in `_meta_dsets_deprecated`. I created this as a helpful placeholder to deal with any future changes to `uvpspec.meta_dsets` that might occur. As of now, there is no concrete scenario where this line of code will actually get called. If you think this is unecessary we can drop it.

hera_pspec/uvpspec.py
aewallwi: See above.

hera_pspec/plot.py
aewallwi: I don't think this file was covered before?

aewallwi:
Hi @philbull . I believe @JianrongTan will be taking over this PR. Hopefully we can get it out soon.

hera_pspec/uvpspec.py
JianrongTan: del to_delete won't successfully remove the attribute from the object. I change it to delattr.

JianrongTan:
Hi @philbull and @aewallwi! I have made some minor changes:

1. Delete some redundant lines.
2. Add more tests on backward-compatibility.  
3. Fix the conflicts with new versions of pyuvdata. 

I think the PR now is fine to be merged into the main channel.

philbull (approved):
Thanks both, happy for this to be merged.

PR #380, opened 2023-03-30
feat: add profiling/logging arguments to scripts

steven-murray:
This simply uses the new cli-tools in hera_cal to add a bit of functionality to hera_pspec scripts.

PR #381, opened 2023-04-06
Speedup pspecrun

steven-murray:
The main thing this PR does is cache the `get_Q_alt` method, because it was taking up ~95% of the computation time for H6C runs (with many baseline pairs). For a small frequency and time window (and only baselines <60m), it was taking ~15hrs to compute the time-average power spectra. With these updates it takes ~4 hrs. 

For completeness, I'm including the line-profile output of the run with this update:
[lineprof.txt](https://github.com/HERA-Team/hera_pspec/files/11167901/lineprof.txt)


steven-murray:
@philbull this depends on #380 so that should be merged first.

hera_pspec/pspecdata.py
```diff
@@ -83,12 +87,17 @@ def __init__(self, dsets=[], wgts=None, dsets_std=None, labels=None,
             dsets_std = None
 
         # Store the input UVData objects if specified
-        if len(dsets) > 0:
+        if dsets is not None and len(dsets) > 0:
```
adeliegorce, line 90: There is no way `dsets` can be `None` here since you have the condition `dsets = [] if dsets is None else dsets` above

hera_pspec/tests/test_pspecdata.py
```diff
@@ -909,6 +909,9 @@ def test_q_hat(self):
         #Test if error is raised when one tried FFT approach on exact_norm
         pytest.raises(NotImplementedError, self.ds.q_hat, key1, key2, exact_norm=True, allow_fft = True)
 
+        print(self.ds._get_qalt_cached.cache_info())
```
adeliegorce, line 912: Are these here on purpose?

scripts/auto_noise_run.py
```diff
@@ -6,31 +6,31 @@
...
-parser = utils.uvp_noise_error_parser()
-args = parser.parse_args()
-
-# compute Tsys from autocorr uvd
-uvd = UVData()
-uvd.read(args.auto_file)
-auto_Tsys = utils.uvd_to_Tsys(uvd, beam=args.beam)
-# load in pspec container.
-psc = PSpecContainer(args.pspec_container, keep_open=False, mode='rw', swmr=False)
+def main(args):
```
adeliegorce, line 11: That's the previous PR, right? it's probably worth merging `main` to have a clean new PR

adeliegorce (changes requested):
Hi! Thanks for the PR, impressive speed-up. 
If I understand correctly, there is now a `get_Q_alt` method both in `pspecdata` and in `utils`. Isn't this confusing? Couldn't we use the one from `utils` in `pspecdata`? 
If possible, could you merge `main` into this branch before merging the PR? To keep things separate.
I have a couple other minor comments before approving.

hera_pspec/tests/test_pspecdata.py
steven-murray: Good catch, will remove.

hera_pspec/pspecdata.py
steven-murray: Good point!

steven-murray:
@adeliegorce I've merged in `main` -- sorry, should have done that previously. I addressed your comments to specific bits of code also. 

I'm not quite sure what you mean about the doubling-up of `get_Q_alt`? The plain (non-cached) function for `get_Q_alt` is in `utils`, and I wrapped that in a cache _inside_ the `PSpecData` object (as `self._get_qalt_cached`), and that function gets called by `self.get_Q_alt()`, which does a bit of setup first. I think this is the best way to organize it -- the cool thing about caching the function inside the class is that the maximum size of the cache can be set more intelligently.

adeliegorce:
Cool, thanks for the changes. I did spot the Q_alt being read from cache in the new version - I must have been looking at the wrong files last week. Could I just ask you to add the two missing tests? (see `codecov` report)

steven-murray:
@adeliegorce sorry for the delay on this, just got to it now. I've added a test of one of the lines, but the other has to do with a warning raised when a the `phase_type` of a constituent dataset is not `drift`. The only reason that line is in this PR is because I updated it from being a print statement to a warning (i.e. it obviously wasn't covered previously either). More concerning, the newer versions of pyuvdata don't support the `drift` phase_type (or even setting `phase_type` at all!). This should be fixed up properly in its own PR, I think.

adeliegorce (approved):

adeliegorce:
Ok I will merge and delete the branch! Thanks @steven-murray

PR #382, opened 2023-07-14
Implemented compatibility with UVData future_array_shapes

adeliegorce:
This PR makes `hera_pspec` compatible with the new version of `UVData` using `future_array_shapes` (i.e. `spw` is no longer an axis in data arrays) whilst remaining compatible with older `UVPspec` objects that could be read.
Main features are:
- All `UVData`, `UVCal` and `UVBeam` objects are now read in with the `future_array_shapes` option. When converted in a `PSpecBeam` or `PSpecData` object, the resulting object is (mostly) unchanged (see below).
- To reproduce the new `UVData` behaviour, now `PSpecData` objects have an array of `channel_width` instead of a single value. This will be (most of the time) `Nfreqs` times the same value.
All tests pass. Let me know if you want to discuss additional features or find bugs - this is very basic.


bhazelton:
This looks good and is very timely -- we're actively working on pyuvdata 3.0 where we will remove support for the old shapes. One thing I wanted to mention is that we now have quite good support for spectral windows. Each frequency is assigned to a spectral window using the `flex_spw_id_array` on UVData and UVCal objects. I know that pspec uses spectral windows, so I wanted to highlight this in case it was useful to you.

adeliegorce:
> One thing I wanted to mention is that we now have quite good support for spectral windows. Each frequency is assigned to a spectral window using the `flex_spw_id_array` on UVData and UVCal objects. I know that pspec uses spectral windows, so I wanted to highlight this in case it was useful to you.

Hi @bhazelton, thanks for the comments! I think this PR is a good occasion to discuss the structure of `UVPSpec` objects - but might be beyond its scope. We'll keep this feature in mind when doing the large-scale review of the code in the fall :)

hera_pspec/utils.py
```diff
@@ -116,7 +116,8 @@ def variance_from_auto_correlations(uvd, bl, spw_range, time_index):
     x_bl1 = uvd.get_data(bl1)[time_index, spw]
     x_bl2 = uvd.get_data(bl2)[time_index, spw]
     nsample_bl = uvd.get_nsamples(bl)[time_index, spw]
-    nsample_bl = np.where(nsample_bl>0, nsample_bl, np.median(uvd.nsample_array[:,:,spw,:]))
+    nsample_bl = np.where(nsample_bl>0, nsample_bl, np.median(uvd.nsample_array[:, spw, :]))
+    df = np.array(uvd.channel_width)[spw]
```
jsdillon, line 120: do you really want this to be an array? if not, `np.median(np.diff(uvd.freq_array))` might be what you want

jsdillon (approved):
Great work! A few minor comments, but I'm basically good to go.

hera_pspec/pspecdata.py
```diff
@@ -3513,7 +3516,7 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
         uvp.polpair_array = np.array(spw_polpair, int)
         uvp.Npols = len(spw_polpair)
         uvp.scalar_array = np.array(sclr_arr)
-        uvp.channel_width = dset1.channel_width  # all dsets validated to agree
+        uvp.channel_width = np.array(dset1.channel_width)  # all dsets validated to agree
```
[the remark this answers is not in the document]
adeliegorce: Just in case it is modified later on, I think it is safer to define it as a copy in the new object

hera_pspec/utils.py
adeliegorce: I chose to follow the new `UVData` syntax inside the `UVPSpec` objects and make `channel_width` an array (although, most of the time, it will be an array of identical values). I added broadcasting to the next equation to make array shapes clearer.

PR #386, opened 2023-08-10
Fix residual issues with future_array_shapes affecting noise calculation

jsdillon:
This PR replaces a couple of lingering `freq_array[0]` with np.squeeze() which works for both past and future array shapes.

I've added a unit test that fails on the current main branch and passes (I hope!) on this branch. 

Finally, while searching the repo for `freq_array[0]` I found a couple of calculations of delta nu that aren't quite right (they take the middle freq of the first and last bin and divide by the number of bins, when really want you want to do is a diff... that calculation will be wrong by a factor of 2 in the limit where there are only 2 frequency channels). So I fixed those too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ ex.hdf5
 out.h5
 .coverage
 coverage.xml
+.DS_Store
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -16,4 +16,5 @@ hera_pspec/_version.py
 ex.hdf5
 out.h5
 .coverage
-coverage.xml
\ No newline at end of file
+coverage.xml
+.DS_Store
```
adeliegorce, line 20: Is there a reason this changed? it seems unrelated to the PR

jsdillon: Just added some stuff to avoid Mac's accidentally committing files we don't want.

adeliegorce (approved):
All good! All tests pass. Thanks!

PR #388, opened 2023-09-06
Added requirement files for readthedocs

adeliegorce:
This PR deals with a new requirement from Readthedocs (see [link](https://blog.readthedocs.com/migrate-configuration-v2/)). It adds a `.readthedocs.yaml` file (version 2) listing the build requirements for the documentation as well as a `docs/requirements.txt` which lists the version of `sphinx` etc. required to build the documentation.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,33 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for 
+details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.9"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+# You can configure Sphinx to use a different builder, for instance use the dirhtml builder for simpler URLs
+# builder: "dirhtml"
+# Fail on all warnings to avoid broken references
+# fail_on_warning: true
+
+# Optionally build your docs in additional formats such as PDF and ePub
+# formats:
+#    - pdf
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,33 @@
...
+# details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.9"
```
steven-murray, line 13: Why not 3.10?

.readthedocs.yaml
```diff
@@ -0,0 +1,33 @@
...
+# formats:
+#    - pdf
+#    - epub
+
+# Optional but recommended, declare the Python requirements required
+# to build your documentation
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+   install:
+   - requirements: docs/requirements.txt
```
steven-murray, line 33: are you sure you don't need the package itself to be installed as well?

steven-murray (commented):
Thanks @adeliegorce. It'd be great to have fantastic pspec docs!

.readthedocs.yaml
adeliegorce: I copied the version in the `setup.cfg` file but there is no reason it couldn't be 3.11

.readthedocs.yaml
adeliegorce: I think it is only the requirements to build the docs, like the version of `sphinx` etc...

adeliegorce:
Thanks for the comments @steven-murray. I changed the python version to 3.11. We're on our way to "fantastic pspec docs" 😄

steven-murray (approved):
Looks good, thanks @adeliegorce and sorry for the delay!

PR #389, opened 2023-09-26
Fix uvp_noise_error() for interleaved LSTs

jsdillon:
The documentation of `utils.uvp_noise_error()` suggest that if "power spectra were computed from interleaved times then the supplied `auto_Tsys` should include all of the times that are present in all of the pairs." For interleaved power spectra, that the number of times in `auto_Tsys` (a `UVData` object) might be double (or more, for more interleaves) the number of times in `uvp.lst_avg_array`.

This creates a problem in the line 

```
Tsys = interp1d(lsts[~Tflag], Tsys[~Tflag], kind='nearest', bounds_error=False, fill_value='extrapolate')(lst_avg)
``` 

because `lsts` is calculated as

```
lst_indices = np.unique(auto_Tsys.lst_array, return_index=True)[1]
lsts = auto_Tsys.lst_array[sorted(lst_indices)]
```

while `Tflag` has a shape determined by the number of times in `uvp.time_1_array` and `uvp.time_2_array`. When that number of times don't match, this causes an error in `lsts[~Tflag]` because `Tflag` is half the length of `lsts` (for example). 

This PR fixes that bug by ensuring that the range of lsts indexed into by `~Tflag` is the precise right shape.

@Kai-FengChen can you review this PR? Also, @aewallwi just wanted to give you a heads up since this might impinge on your work.

### DIFF
--- a/hera_pspec/utils.py
+++ b/hera_pspec/utils.py
@@ -1406,7 +1406,6 @@ def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_
         for blp in uvp.get_blpairs():
             blp_int = uvp.antnums_to_blpair(blp)
             blp_inds = uvp.blpair_to_indices(blp)
-            lst_avg = uvp.lst_avg_array[blp_inds]
 
             time_1_selection = uvp.time_1_array[blp_inds]
             time_2_selection = uvp.time_2_array[blp_inds]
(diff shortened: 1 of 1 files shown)


hera_pspec/utils.py
```diff
@@ -1451,7 +1452,7 @@ def uvp_noise_error(uvp, auto_Tsys=None, err_type='P_N', precomp_P_N=None, P_SN_
                         Tflag = np.all(Tflag, axis=-1)
                         # interpolate to appropriate LST grid
                         if np.count_nonzero(~Tflag) > 1:
-                            Tsys = interp1d(lsts[~Tflag], Tsys[~Tflag], kind='nearest', bounds_error=False, fill_value='extrapolate')(lst_avg)
+                            Tsys = interp1d(auto_lst_avg[~Tflag], Tsys[~Tflag], kind='nearest', bounds_error=False, fill_value='extrapolate')(lst_avg)
```
JianrongTan, line 1455: Is it better to interpolate on tinds1[~Tflag] and tinds2[~flag] separately and then multiply?

jsdillon: That seems smart. I can do that.

jsdillon: Ok, done.

JianrongTan (approved):
Looks good to me. Thanks @jsdillon!

PR #392, opened 2024-02-05
Proper combining of nsamples for psuedo-Stokes to preserve proper noise variance

### LINKED ISSUE #391: Pseudo-Stokes Nsamples Accounting
https://github.com/HERA-Team/hera_pspec/blob/fc06cf7d1b8ba370e4194b02eb49a1b08b757152/hera_pspec/pstokes.py#L148

I'll repeat what I said in Slack:

In the case where ee- and nn-polarized visibilities have different nsamples (because of different antennas being flagged on different days, for example), it’s not clear to me that this is the proper nsamples for computing the thermal noise. Consider a case where `uvd1.nsample_array` is 10 and `uvd2.nsample_array `is 1. The variance will be dominated by `uvd2.nsample_array`. So if we want nsamples to properly reflect the variance, then I think we want to do something like
```uvdS.nsample_array = 4 * (uvd1.nsample_array**-1 + uvd2.nsample_array**-1)**-1```
where the 4 accounts for the fact that ```uvdS.data_array = uvd1.data_array / 2  + uvd2.data_array / 2```.
In other words, if the ee nsamples and the nn nsamples are equal, then the answer is just the sum of the nsamples. But if one of the two is zero, then the subsequent nsamples should also be 0 reflecting the fact that psuedo-I and psuedo-Q have infinite variance

jsdillon:
This PR resolved #391

### DIFF
--- a/hera_pspec/pstokes.py
+++ b/hera_pspec/pstokes.py
@@ -145,7 +145,8 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
     uvdS.data_array = stdata  # pseudo-stokes data
     uvdS.flag_array = flag  # flag array
     uvdS.polarization_array = np.array([pstokes], dtype=int) # polarization number
-    uvdS.nsample_array = uvd1.nsample_array + uvd2.nsample_array # nsamples
+    # nsamples combined to preserve proper variance, see hera_pspec issue #391
+    uvdS.nsample_array = 4 * (uvd1.nsample_array**-1 + uvd2.nsample_array**-1)**-1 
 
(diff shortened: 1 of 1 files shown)


mwilensky768 (approved):
The simulations show that it's clearly predicting the right variance, and moreover it's all mathematically sound. While I agree that it doesn't quite agree with the definition of nsamples, it's the right thing to do for variance prediction and that is all we use nsamples for in pspec as far as I know.

I've done a quick check to see that if one or both of these entries are 0.0 (that is, a numpy array with 0.0 as entries, not the builtin float 0.0), you get 0.0, which is as desired. Additionally nan's beget nans. 

I approve.

PR #394, opened 2024-02-09
Fix test errors: codecov secret and missing package for docs

adeliegorce:
This PR adds two elements that were missing for the tests to pass:
- Adds a secret for `codecov`. Although it is not required for public repos, uploads often fail when you don't have one. If you don't set one, it uses like a backup public token which has a rate limit (@steven-murray)
- Adds the `mock` package as a requirement for the documentation: building `readthedocs` is failing without (called in `conf.py`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,12 +41,13 @@ jobs:
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v4
+      env:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       with:
         files: ./coverage.xml
         directory: ./coverage/reports/
         flags: unittests
         env_vars: OS,PYTHON
         name: codecov-umbrella
-        token: ${{ secrets.CODECOV_TOKEN }}
         fail_ci_if_error: true
         verbose: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
       env:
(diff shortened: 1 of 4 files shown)


adeliegorce:
To whoever reviews this PR, a question: why is there a list of requirements for `docs` in `setup.cfg` _and_ a `docs/requirements.txt` file? it seems like the latter overruns the former? thanks

.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
```
steven-murray, line 43: You should update this to `v4` as well

steven-murray (commented):
Thanks @adeliegorce. To answer your question -- I agree it's redundant. I strongly prefer simply defining it in `setup.cfg` (or better, `pyproject.toml`!) so I would suggest removing the `docs/requirements.txt`, but other people have different opinions. I like defining all the requirements in one place.

.github/workflows/ci.yml
adeliegorce: I tried v4 but there is an issue that prevents codecov from finding the coverage.xml file. It is a regression issue currently being looked at by the team https://github.com/codecov/codecov-action/issues/1278

adeliegorce: So for now I stick with v3 :)

adeliegorce:
Hi @steven-murray, thanks for the review. I have made the requested changes apart from switching to `codecov-actions@v4` as there is a regression issue. Let's keep it in mind (or create an issue related to this) and do the update in a couple months once things are more stable.

steven-murray (approved):

Issue #396, opened 2024-03-05
Lack of backward compatibility for non-future_array_shape in utils.uvd_to_Tsys


JianrongTan:
In `utils.uvd_to_Tsys`, this line doesn't work if `uvd` is not using `future_array_shape`

`uvd.data_array[tinds, :, pind] *= J2K[pol]`

adeliegorce:
Thank you for pointing this out. This issue is addressed in [PR#397](https://github.com/HERA-Team/hera_pspec/pull/397).

adeliegorce:
@JianrongTan @lisaleemcb could you try to re-run your pipeline in the `future-array-shapes-bug` branch and let me know if this fixed your issue?

JianrongTan:
Thanks @adeliegorce! I have tested with the updated code and the issue is solved.

adeliegorce:
Great! Could you review the associated PR?

PR #397, opened 2024-03-05
Added tests re. future array shapes and ensured compatibility in utils

adeliegorce:
Following @JianrongTan's issue #396 , this PR

1. Fixes a backward compatibility issue in `utils.uvd_to_Tsys` which raised an error when the `UVData` file used has `future_array_shapes=False`
2. Adds tests to `test_utils` to ensure backwards compatibility of the code with and without future array shapes.


JianrongTan (approved):
Looks all good to me. Thanks @adeliegorce!

PR #398, opened 2024-03-05
Change python version in tests

adeliegorce:
This PR removes `python=3.8` from tests as there was a compatibility problem with `hera_cal` and `optax1` and added 3.11 and 3.12.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,8 @@ classifiers =
     Development Status :: 5 - Production/Stable
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
     Intended Audience :: Science/Research
     License :: OSI Approved
     Natural Language :: English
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -18,8 +18,8 @@ platforms = OSX,Linux
 # https://pypi.python.org/pypi?%3Aaction=list_classifiers
 classifiers =
     Development Status :: 5 - Production/Stable
-    Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
steven-murray, line 22: Why not also add 3.11 and 3.12?

steven-murray (approved):
Hey @adeliegorce, this is good -- just one minor comment and good to go.

setup.cfg
adeliegorce: Because I was not sure how useful it is to list versions here? but yes, now the tests passed with 3.11 and 3.12 you're right, I'll add them

adeliegorce:
Thank you for the review @steven-murray ! I will make the changes requested and merge.

PR #402, opened 2024-07-10
Fix api call to ENU_from_ECEF for pyuvdata 3.0

jsdillon:


steven-murray:
@jsdillon I fixed all the issues except three things:

1. There are a bunch of errors on the `setup_class()` method of `Test_Utils`. However, the line on which it errors *does not exist* on my local copy (nor if you look at the "Changes" in this PR). I cannot understand where this code is coming from that is causing the error. Is the github action checking out a different commit than the one in this PR??
2. Python 3.9 fails because there is no pyuvdata v3 for it. I could update the one breaking line of code in this PR so that it can work *also* with pre-pyuvdata 3, but in all other repos I've gone with "just require pyuvdata 3", so I'd argue for doing this, and support python 3.10+
3. There are two more errors occurring now in window function tests. I wonder if @adeliegorce could weigh in on these.

jsdillon:
1) No clue.

2) Agreed, drop python 3.9.

hera_pspec/pspecdata.py
adeliegorce, line 1: l. 338 is not tested

adeliegorce (commented):
Hi! 
I reviewed what @jsdillon and @steven-murray had done (thanks!) and added a few things myself that then I did not review. Note that we are dropping support for NOT `future_array_shapes` (as `pyuvdata` did): I removed all the tests re. the shape of the `data_arrays` since they were erroring with `pyuvdata` v3.0 (cannot create a `UVData` object with old array shapes anymore).
Everything should be working now bar the coverage tests: a test re. line 338 of `pspecdata` must be added - I did not do it as I am not familiar with the attributes in question.

jsdillon:
I removed some irrelevant test and added python 3.12 testing, but now one WF test is failing again if you want to look at it @adeliegorce.

I'm not so worried about the coverage issue, but maybe @steven-murray has an idea of why pyuvdata changes have changed the coverage here?

adeliegorce:
Hi @jsdillon, I fixed the issues in the `UVWindow` tests, we're just down to the coverage issue (it really is just one rather long `if` statement to cover)

steven-murray:
I totally don't think it's worth trying to cover that one corner case exception.

jsdillon:
Agreed. I'll merge.

adeliegorce:
Ah but it brought the coverage way down…

steven-murray:
Not according to codecov? It was just one line

adeliegorce:
I thought I saw somewhere coverage was down to 92% but you're right, it's actually still around 96%

PR #404, opened 2024-07-18
Implement `pol_convention` in `pstokes` module

### LINKED ISSUE #403: Implement psuedostokes that uses pol_convention appropriately
When https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1455 is merged in, UVData and UVCal objects will know about the `pol_convention` which is either `'sum'` or `'avg'`. hera_pspec currently assumes the average convention, but we're trying to move to sum as a collaboration.

https://github.com/HERA-Team/hera_pspec/blob/main/hera_pspec/pstokes.py needs to be modified to be responsive to the pol convention. I think it's fine to keep using the average convention when no pol_convention is provided, but there should definitely be a warning. If `uvd1` and `uvd2` disagree on the pol_convention, that should definitely be an error.

Also, when we do this, I'd like there to be a lower-level function (or set of functions) that is called by `_combine_pol()` and that operates on numpy arrays, rather than UVData objects. Right now, in https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_postprocessing_and_pspec.ipynb this is done with the following function:

```
def form_pseudostokes(data=None, flags=None, nsamples=None):
    '''This function creates pseudo-Stokes I and Q versions corresponding to the ee and nn entries
    in the input data containers, and stores them in the same data containers.'''
    # data are averaged (or diffed in the case of pQ)
    if data is not None:
        for antpair in data.antpairs():
            data[antpair + ('pI',)] = (data[antpair + ('ee',)] + data[antpair + ('nn',)]) / 2
            data[antpair + ('pQ',)] = (data[antpair + ('ee',)] - data[antpair + ('nn',)]) / 2

    # flags are ORed
    if flags is not None:
        for antpair in flags.antpairs():            
            flags[antpair + ('pI',)] = flags[antpair + ('ee',)] | flags[antpair + ('nn',)]
            flags[antpair + ('pQ',)] = flags[antpair + ('ee',)] | flags[antpair + ('nn',)]

    # nsamples are combined to produce the correct variance for pI and pQ. 
    # In the limit where the two pols have the same nsamples, it's just the sum.
    if nsamples is not None:
        for antpair in nsamples.antpairs():
            nsamples[antpair + ('pI',)] = 4 * (nsamples[antpair + ('ee',)]**-1 + nsamples[antpair + ('nn',)]**-1)**-1
            nsamples[antpair + ('pQ',)] = 4 * (nsamples[antpair + ('ee',)]**-1 + nsamples[antpair + ('nn',)]**-1)**-1
```

This needs to be updated to be `pol_covention`-aware too. It also needs to be able to do psuedo U and pseudo V as soon as cross-pols are available. But ideally the lower-level functions in hera_pspec would be callable here so that we can avoid code duplication.

adeliegorce:
This PR introduces some changes to the `pstokes` module for two purposes
1. Take into account the new `pol_convention` attribute of `UVData` objects (see this [PR](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1455)) when constructing pseudo-Stokes polarizations whilst ensuring backwards compatibility (if there is no `pol_convention`, then there is a warning and `avg` is assumed). This is commit [6ffd0dc](https://github.com/HERA-Team/hera_pspec/pull/404/commits/6ffd0dcc6143fe9ac12021dab97556aa5c284c05).
2. As per @jsdillon's request, create a new method called `_combine_pol_arrays` (final name still open to debate) which takes lists of data, flags, and nsamples arrays and combines them into pseudo-Stokes following the appropriate `pol_convention` (i.e. `avg` and `sum`). Although I see how useful this method can be, it does make the code slightly heavy so, reviewers, please let me know if you have any suggestions on how to make things cleaner. This is commit [2601670](https://github.com/HERA-Team/hera_pspec/pull/404/commits/2601670b3d8f17ea2d5c3e73408a683c16d69327).
I added new tests and modified existing ones accordingly.

### DIFF
--- a/hera_pspec/pstokes.py
+++ b/hera_pspec/pstokes.py
@@ -6,6 +6,7 @@ import pyuvdata
 import copy
 from collections import OrderedDict as odict
 import argparse
+import warnings
 
 from . import version, __version__
 
@@ -86,8 +87,10 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
         Polarization of the second UVData object to use in constructing
         pStokes visibility.
 
-    pstokes: Pseudo-stokes polarization to form, type: str
-        Pseudo stokes polarization to form, can be 'pI' or 'pQ' or 'pU' or 'pV'.
+    pstokes: Pseudo-stokes polarization to form, type: str or int
+        Pseudo stokes polarization to form, can be in a number
+        or a string, in which case pyuvdata is used to make the conversion
+        to integer.
         Default: pI
 
     x_orientation: str, optional
@@ -108,6 +111,18 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
         pol1 = pyuvdata.utils.polstr2num(pol1, x_orientation=x_orientation)
     if isinstance(pol2, str):
         pol2 = pyuvdata.utils.polstr2num(pol2, x_orientation=x_orientation)
+
+    for i, uvd in enumerate([uvd1, uvd2]):
+        try:
+            uvd.pol_convention
+        except AttributeError:
+            warnings.warn(
+                f'No polarization convention in uvd{i+1}. '
+                'Considering it to be "avg".')
+            setattr(uvd, 'pol_convention', 'avg')
+
+    if uvd1.pol_convention != uvd2.pol_convention:
+        raise ValueError('pol_convention of uvd1 and uvd2 are different.')
 
     # extracting data array from the UVData objects
     data1 = uvd1.data_array
@@ -138,7 +153,9 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
         "pol2 {} not used in constructing pstokes {}".format(pol2_str, pstokes_str)
 
     # constructing Stokes visibilities
-    stdata = 0.5 * (pol_weights[pstokes][pol1]*data1 + pol_weights[pstokes][pol2]*data2)
+    stdata = (pol_weights[pstokes][pol1]*data1 + pol_weights[pstokes][pol2]*data2)
+    if uvd1.pol_convention == 'avg':
+        stdata *= 0.5
 
     # assigning and writing data, flags and metadata to UVData object
     uvdS = copy.deepcopy(uvd1)
--- a/hera_pspec/pstokes.py
+++ b/hera_pspec/pstokes.py
@@ -106,12 +106,6 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
     assert isinstance(uvd2, pyuvdata.UVData), \
         "uvd2 must be a pyuvdata.UVData instance"
 
-    # convert pol1 and/or pol2 to integer if fed as a string
-    if isinstance(pol1, str):
-        pol1 = pyuvdata.utils.polstr2num(pol1, x_orientation=x_orientation)
-    if isinstance(pol2, str):
-        pol2 = pyuvdata.utils.polstr2num(pol2, x_orientation=x_orientation)
-
     for i, uvd in enumerate([uvd1, uvd2]):
         try:
             uvd.pol_convention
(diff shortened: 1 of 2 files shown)


hera_pspec/pstokes.py
```diff
@@ -86,8 +87,10 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
         Polarization of the second UVData object to use in constructing
         pStokes visibility.
 
-    pstokes: Pseudo-stokes polarization to form, type: str
-        Pseudo stokes polarization to form, can be 'pI' or 'pQ' or 'pU' or 'pV'.
+    pstokes: Pseudo-stokes polarization to form, type: str or int
```
jsdillon, line 90: Even though you now also take an integer, I think it's good to list the possible strings.

hera_pspec/pstokes.py
```diff
@@ -137,23 +226,43 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
+    # assert pol_convention makes sense
+    if pol_convention not in ['avg', 'sum']:
+        raise ValueError('pol_convention must be avg or sum.')
+
+    # checks on input lists
+    for a_list in [data_list, flags_list, nsamples_list]:
+        if a_list is None:
+            continue
+        assert len(a_list) == 2, \
+            "Cannot combine more than two arrays."
```
jsdillon, line 238: Or fewer

hera_pspec/pstokes.py
```diff
@@ -103,23 +106,109 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
+        List of data arrays to be combined to form their pseudo-Stokes equivalent.
+        Default is None.
+
+    flags_list : list of arrays
+        List of flag arrays to be combined to form their pseudo-Stokes equivalent.
+        Default is None.
+
+    nsamples_list : list of arrays
+        List of nsamples arrays to be combined to form their pseudo-Stokes equivalent.
+        Default is None.
```
jsdillon, line 188: I'd clarify that these could be any length-2 iterable of a numpy arrays.

hera_pspec/pstokes.py
```diff
@@ -103,23 +106,109 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
+
+    pol2 : Polarization, type: str
+        Polarization of the second UVData object to use in constructing
+        pStokes visibility.
+
+    pstokes: Pseudo-stokes polarization to form, type: str or int
+        Pseudo stokes polarization to form, can be in a number
+        or a string, in which case pyuvdata is used to make the conversion
+        to integer.
+        Default: pI
```
jsdillon, line 170: I'd mention the valid `str` inputs in the docstring for these parameters

hera_pspec/pstokes.py
```diff
@@ -137,23 +226,43 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
+    # constructing nsamples
     # nsamples combined to preserve proper variance, see hera_pspec issue #391
-    uvdS.nsample_array = 4 * (uvd1.nsample_array**-1 + uvd2.nsample_array**-1)**-1 
-
-    uvdS.history = "Merged into pseudo-stokes vis with hera_pspec version {}\n{}" \
-                    "{}{}{}{}\n".format(__version__, "-"*20+'\n',
-                    'dset1 history:\n', uvd1.history, '\n'+'-'*20+'\ndset2 history:\n',
-                    uvd2.history)
+    if nsamples_list is not None:
+        combined_nsamples = 4 * (nsamples_list[0]**-1 + nsamples_list[1]**-1)**-1
```
jsdillon, line 261: I believe this 4 is a 1 in the `sum` convention.

hera_pspec/pstokes.py
jsdillon: Also, clarify how flags and nsamples are combined in the docstring

hera_pspec/pstokes.py
```diff
@@ -103,23 +106,109 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
+        pstokes = pyuvdata.utils.polstr2num(pstokes, x_orientation=x_orientation)
+
+    # assigning and writing data, flags and metadata to UVData object
+    uvdS = copy.deepcopy(uvd1)
+    uvdS.data_array = combined_data # pseudo-stokes data
+    uvdS.flag_array = combined_flags # flag array
+    uvdS.polarization_array = np.array([pstokes], dtype=int) # polarization number
+    uvdS.nsample_array = combined_nsamples
+
+    uvdS.history = "Merged into pseudo-stokes vis with hera_pspec version {}\n{}" \
```
jsdillon, line 139: I find f-strings a lot cleaner and more modern.

hera_pspec/tests/test_pstokes.py
```diff
@@ -30,19 +30,87 @@ def runTest(self):
...
         # basic execution on pol strings
         out1 = pstokes._combine_pol(uvd1, uvd2, 'XX', 'YY')
         # again w/ pol ints
         out2 = pstokes._combine_pol(uvd1, uvd2, -5, -6)
         # assert equivalence
         assert out1 == out2
+        # basic execution with different polarization conventions
+        # out1 assumed avg by default
+        setattr(uvd1, 'pol_convention', 'sum')
+        setattr(uvd2, 'pol_convention', 'sum')
```
jsdillon, line 45: Let's also test that nsamples does the right thing when we're in `sum` mode.

jsdillon (changes requested):
Thanks @adeliegorce! A handful of comments, mostly related to documentation, as well as one substantial one about nsamples.

hera_pspec/pstokes.py
adeliegorce: Ok, I added the option to only feed a list of one element (e.g. `data_list=[uvd1.data_array]`), in which case it is duplicated and combined to itself to form the pseudo-Stokes pol

adeliegorce:
Hi @jsdillon, I made the requested changes. Let me know if you're happy with the new version.

hera_pspec/pstokes.py
```diff
@@ -103,23 +106,114 @@ def _combine_pol(uvd1, uvd2, pol1, pol2, pstokes='pI', x_orientation=None):
...
         "uvd2 must be a pyuvdata.UVData instance"
 
+    for i, uvd in enumerate([uvd1, uvd2]):
+        try:
+            uvd.pol_convention
+        except AttributeError:
+            warnings.warn(
+                f'No polarization convention in uvd{i+1}. '
+                'Considering it to be "avg".')
+            setattr(uvd, 'pol_convention', 'avg')
```
steven-murray, line 116: This is not quite correct -- the attribute can be defined, but set to `None`. So, I would do 

```python
pol_convention = getattr(uvd, 'pol_convention', None)
if pol_convention is None:
    warnings.warn(...)
```

steven-murray (commented):
Thanks @adeliegorce -- I htink is almost ready to go, but does need a simple fix

adeliegorce:
Done!

hera_pspec/pstokes.py
```diff
@@ -233,9 +237,14 @@ def _combine_pol_arrays(
...
+            a_list = None
             continue
-        assert len(a_list) == 2, \
-            "Cannot combine more than two arrays."
+        if not isinstance(a_list, list) and isinstance(a_list, np.ndarray):
+            a_list = [a_list]
+        if len(a_list) == 1:
+            a_list.append(a_list[0])
+        elif len(a_list) > 2:
+            raise ValueError("Cannot combine more than two arrays.")
```
jsdillon, line 247: I believe that if you use a tuple, this will break, since it will turn `(uvd1, uvd2)` into `[(uvd1, uvd2)]` which isn't what you want. Also tuples don't support appending.

To be more general, you can do something like

```from collections.abc import Iterable
is_iterable = isinstance(variable, Iterable)
```

Frankly, I think we should enforce that the length is 2. If the user wants to pass in (uvd1, uvd1) then they can do it themselves.

jsdillon (changes requested):
Just one suggestion that will hopefully avoid confusion in some weird edge cases

hera_pspec/pstokes.py
adeliegorce: Ah, my mistake, I thought from your previous comments that you wanted me to generalize to cases where len < 2.
I agree to keep it simple so I removed the option of only giving a length 1 list and changed the checks to `isinstance(a_list, Iterable)`.

jsdillon (approved):

jsdillon:
Looks good, thanks @adeliegorce!

PR #407, opened 2024-09-25
Use the same factor of 4 for "sum" and "avg" pol_convention

### LINKED ISSUE #406: Error bars off by 4 when using "sum" polarization convention
While converting the [single baseline pspec notebook](https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_postprocessing_and_pspec.ipynb) to use the new "sum" polarization convention, I've started using `hera_pspec.pstokes._combine_pol_arrays()`, as we intended in #404. 

I've confirmed that this yields psuedo-Stokes power spectra with the same normalization (despite the ee and nn visibilities being different by a factor of 2, as intended). 

image

However, the error bars are off, apparently by a factor of 4 in power spectrum units. Previously, the SNRs of power spectra above 1000 ns roughly followed the expected noise distribution:

image

Now, we're quite off:

image

For now, I can "fix" this by just increasing nSamples by a factor of 4 (which we do in the "avg" convention in `hera_pspec.pstokes._combine_pol_arrays()` but not in the "sum" convention). However, that feels very ad hoc and there's probably a more principled way to get this to work properly.

Here's the full function in the notebook, which hopefully will be pushed up soon:
```
def form_pseudostokes(data=None, flags=None, nsamples=None, pol_convention=hd.pol_convention, x_orientation=hd.x_orientation):
    '''This function uses hera_pspec.pstokes to supplement existing datacontainers with pstokes I, Q, U, and/or V (where possible).'''
    for ap in data.antpairs():
        # loop over pseudo-stokes parameters
        for pol1, pol2, pstokes in [('ee', 'nn', 'pI'), ('ee', 'nn', 'pQ'), ('en', 'ne', 'pU'), ('en', 'ne', 'pV')]:
            bl1 = ap + (pol1,)
            bl2 = ap + (pol2,)
            data_list = ([data[bl1], data[bl2]] if (data is not None) and (bl1 in data) and (bl2 in data) else None)
            flags_list = ([flags[bl1], flags[bl2]] if (flags is not None) and (bl1 in flags) and (bl2 in flags) else None)
            nsamples_list = ([nsamples[bl1], nsamples[bl2]] if (nsamples is not None) and (bl1 in nsamples) and (bl2 in nsamples) else None)

            # use hp.pstokes._combine_pol_arrays() to properly combine data/flags/nsamples in a pol_convention-aware way
            (combined_data, 
             combined_flags, 
             combined_nsamples) = hp.pstokes._combine_pol_arrays(pol1, pol2, pstokes, 
                                                                 pol_convention=pol_convention,
                                                                 data_list=data_list,
                                                                 flags_list=flags_list,
                                                                 nsamples_list=nsamples_list,
                                                                 x_orientation=x_orientation)
            # put results in original data containers
            if data is not None:
                data[ap + (pstokes,)] = combined_data
            if flags is not None:
                flags[ap + (pstokes,)] = combined_flags
            if nsamples is not None:
                nsamples[ap + (pstokes,)] = combined_nsamples
```

jsdillon:
This follows from @JianrongTan's work, as documented in #406.

This closes #406

### DIFF
--- a/hera_pspec/pstokes.py
+++ b/hera_pspec/pstokes.py
@@ -188,7 +188,7 @@ def _combine_pol_arrays(
 
     nsamples_list :  any length 2 iterable of numpy arrays
         Iterable of nsamples arrays to be combined to form their pseudo-Stokes equivalent.
-        nsamples are combined to preserve proper variance, see hera_pspec issue #391.
+        nsamples are combined to preserve proper variance, see hera_pspec issue #391 and #406
         If only one is given, it is duplicated.
         Default is None.
 
@@ -264,9 +264,9 @@ def _combine_pol_arrays(
 
     # constructing nsamples
     if nsamples_list is not None:
-        combined_nsamples = (nsamples_list[0]**-1 + nsamples_list[1]**-1)**-1
-        if pol_convention == 'avg':
-            combined_nsamples *= 4.
+        combined_nsamples = 4 * (nsamples_list[0]**-1 + nsamples_list[1]**-1)**-1
+        # for an explanation of the factor of 4, and why it doesn't depend on pol_convention,
+        # see https://github.com/HERA-Team/hera_pspec/issues/406
     else:
         combined_nsamples = None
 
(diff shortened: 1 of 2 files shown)


JianrongTan (approved):
Looks good to me. Thanks @jsdillon!

PR #408, opened 2024-10-24
Add recursive_combine_uvpspec()

jsdillon:
This PR graduates a function that currently lives in both https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/single_baseline_postprocessing_and_pspec.ipynb and https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/power_spectrum_summary.ipynb

It is a somewhat faster way to combine large numbers of `uvpspec` objects using recursion, allowing fewer total adds to be performed.

### DIFF
--- a/hera_pspec/uvpspec.py
+++ b/hera_pspec/uvpspec.py
@@ -2712,6 +2712,41 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
     return u
 
 
+def recursive_combine_uvpspec(uvps):
+    """
+    Method for faster combination of UVPSpec objects by combining them recursively.
+    This is faster than combine_uvpspec for long lists of files---e.g. if you have 
+    one uvpspec object for every unique baseline and hundreds of baselines. Note:
+    Histories are not merged, so this is the equivalent of running combine_uvpspec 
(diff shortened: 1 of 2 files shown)


hera_pspec/uvpspec.py
```diff
@@ -2712,6 +2712,41 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
...
+    uvps : list
+        A list of UVPSpec objects to combine.
+
+    Returns
+    -------
+    u : UVPSpec object
+        A UVPSpec object with the data of all the inputs combined.
+
+    """
+    if len(uvps) == 0:
```
adeliegorce, line 2734: Could you add the same tests to `combine_uvpspec` or move them there? only `len(uvps)==0` case is covered in `combine_uvpspec` right now (through `get_uvp_overlap`)

adeliegorce (approved):
Just a little comment but apart from this it's all good on my side! thanks!

hera_pspec/uvpspec.py
jsdillon: Sure

PR #412, opened 2025-03-25
Add `get_approximat_delay_delay_corr` function

steven-murray:
This adds a new simple function `noise.get_approximate_delay_delay_corr` which can compute a correlation matrix for power spectra estimated with the simplest power spectrum options, and ignoring complications such as inpainting. It also adds a method to the `UVPSpec` object to add the `cov_array_real` array to itself, using this approximate scheme.


Kai-FengChen:
This looks good, have now tested this code (black) against the full covariance treatment with (red) or without (blue) the effect of inpainting. The approximation agrees well with the full covariance for the neighbouring four bins which we use the covariance for.
Screenshot 2025-05-20 at 10 48 12

Kai-FengChen (approved):

PR #420, opened 2025-05-08
Get rid of future_array_shapes in PSpecBeam

mwilensky768:
Just getting rid of reference to future_array_shapes in PSpecBeam since this is no longer a thing in pyuvdata and it holds up some of the analysis notebooks. I did not check anywhere else for where this might cause us problems.

### DIFF
--- a/hera_pspec/pspecbeam.py
+++ b/hera_pspec/pspecbeam.py
@@ -415,10 +415,9 @@ class PSpecBeamUV(PSpecBeamBase):
         # setup uvbeam object
         if isinstance(uvbeam, str):
             uvb = UVBeam()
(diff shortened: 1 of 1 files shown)


hera_pspec/pspecbeam.py
```diff
@@ -415,10 +415,9 @@ def __init__(self, uvbeam, cosmo=None):
         # setup uvbeam object
         if isinstance(uvbeam, str):
             uvb = UVBeam()
-            uvb.read_beamfits(uvbeam, use_future_array_shapes=True)
```
mwilensky768, line 418: I wonder if this will cause things to break when reading in old files.

PR #421, opened 2025-05-20
maint: use pyuvdata 3.2

steven-murray:
Supercedes #420 and fixes all the API stuff for pyuvdata 3.2, making tests pass again.


mwilensky768 (commented):
These are all sensible changes. Have the notebooks run through OK? Or are you waiting because you have to merge several branches on some repos before you can test?

mwilensky768:
I just wanted to check that uncovered lines wouldn't come bite us later. I noticed this:

https://github.com/HERA-Team/hera_pspec/blob/6aa9f1304c6ce26789e73e114202561697c8dc73/hera_pspec/utils.py#L508

I think leaving this line will allow someone to use pyuvdata less than 3, but won't break if someone is using pyuvdata 3 so maybe it's fine to just leave this in perpetuity.

mwilensky768:
```
if ((notebooks_run_through) or (merging_multiple_branches)) and (line_508_in_utils_is_OK):
    mike_approves = True
...
```

mwilensky768 (approved):

PR #423, opened 2025-05-28
Use only other nsamples from the same basetime/time/spw to fill in zeros in variance_from_auto_correlations()

jsdillon:
This PR was made in response to @RajorshiSChandra's question in the May 28 pspec telecon.


mwilensky768 (approved):
This looks good. I think the inf should only happen when all the nsamples are 0, as any 0 nsamples would be replaced with nonzero ones if any nonzero ones exist initially. In that case, infinite variance makes sense in some contexts.

PR #424, opened 2025-06-12
Changed kpara binning to sample exact delays

kai-feng-chen:
@adeliegorce has made these changes so that the k_para binning for the exact window function aligns with the delay sampling. 

After this PR:
Screenshot 2025-06-12 at 18 03 48

Before:
Screenshot 2025-06-12 at 18 04 01


steven-murray (approved):
This does look correct

PR #425, opened 2025-06-17
fix: Ntimes is now the same as Ntpairs

steven-murray:
I updated the code so that `uvp.Ntimes` is no longer the number of unique times within both the `time_1_array` and `time_2_array`, but instead reflects what most people expect of it -- the number of unique *time-pairs* (i.e. the same as `Ntpairs`). 

I updated the docstring to reflect this, as well as updating a few of the existing functions that set `Ntimes` to be the set of unique times in the underlying visibilities:

* `combine_uvpspec`
* `read_hd5f_group`
* `average_spectra`

Note that in the code, several places already assumed that `Ntimes` referred to the number of time-pairs, not the number of unique times of the underlying visibilities, for example:

* `uvpspec_utils._select`
* Several of the tutorial notebooks
* Window functions

This standardizes the usage across the repo.

I also added a number of new test cases, especially within `test_uvpspec`, including test cases that would have failed before I made these updates (since the `build_vanilla_uvp` function always output pspec objects where `time_1_array` and `time_2_array` were equal, now I have made cases where they are not).

### DIFF
--- a/hera_pspec/grouping.py
+++ b/hera_pspec/grouping.py
@@ -179,6 +179,11 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
     add_to_history : str, optional
         Added text to add to file history.
 
+    time_tol : float, optional
+        The tolerance for checking if times are equivalent, in order to count
+        "unique" time-pairs in the average. The units are (julian) days. Setting too
+        low can result in numerical noise creating more time-pairs than expected.
+
     Notes
     -----
     Currently, every baseline-pair in a blpair group must have the same
@@ -591,7 +596,7 @@ def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
 
 
 def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=False, blpair_weights=None,
-                      weight_by_cov=False, error_weights=None,
+                      weight_by_cov=False, error_weights=None, time_tol: float = 1e-6,
                       add_to_history='', little_h=True, A={}, run_check=True):
     """
     Perform a spherical average of a UVPSpec, mapping k_perp & k_para onto a |k| grid.
@@ -641,6 +646,11 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
     run_check : bool, optional
         If True, run UVPSpec.check() on resultant object
 
+    time_tol : float, optional
+        The tolerance for checking if times are equivalent, in order to count
+        "unique" time-pairs in the average. The units are (julian) days. Setting too
+        low can result in numerical noise creating more time-pairs than expected.
+        
     Returns
     --------
     UVPSpec object
@@ -683,6 +693,7 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
     # perform time and cylindrical averaging upfront if requested
     if not uvp.exact_windows and (blpair_groups is not None or time_avg):
         uvp.average_spectra(blpair_groups=blpair_groups, time_avg=time_avg,
+                            time_tol=time_tol,
                             blpair_weights=blpair_weights, error_weights=error_weights,
                             inplace=True)
 
@@ -948,7 +959,7 @@ def spherical_average(uvp_in, kbins, bin_widths, blpair_groups=None, time_avg=Fa
 def spherical_wf_from_uvp(uvp_in, kbins, bin_widths,
                           blpair_groups=None, blpair_lens=None, blpair_weights=None,
                           error_weights=None, time_avg=False, spw_array=None,
-                          little_h=True, verbose=False):
+                          little_h=True, verbose=False, time_tol: float=1e-6):
     
     """
     Obtains exact spherical window functions from an UVPspec object,
@@ -995,6 +1006,11 @@ def spherical_wf_from_uvp(uvp_in, kbins, bin_widths,
         If True, print progress, warnings and debugging info to stdout.
         If None, value used is the class attribute.
 
+    time_tol : float, optional
+        The tolerance for checking if times are equivalent, in order to count
+        "unique" time-pairs in the average. The units are (julian) days. Setting too
+        low can result in numerical noise creating more time-pairs than expected.
+        
     Returns
     --------
     wf_spherical : array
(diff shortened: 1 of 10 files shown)


hera_pspec/testing.py
```diff
@@ -42,103 +46,108 @@ def build_vanilla_uvpspec(
...
-    norm_units = "Hz str"
-    weighting = "identity"
-    channel_width = np.ones(Nfreqs) * np.median(np.diff(freq_array))
-    history = "example"
-    taper = "none"
-    norm = "I"
-    git_hash = "random"
-    scalar_array = np.ones((Nspws, Npols), float)
+    #time_array = np.repeat(time_array, Nblpairs)
+    uvp.Ntpairs = len(set([(t1, t2) for t1, t2 in zip(uvp.time_1_array, uvp.time_2_array)]))
```
adeliegorce, line 82: This can simply be `uvp.Ntpairs = uvp.time_1_array.size`

hera_pspec/uvpspec.py
```diff
@@ -26,7 +26,14 @@ def __init__(self):
         by hera_pspec.
         """
         # Summary attributes
-        self._Ntimes = PSpecParam("Ntimes", description="Number of unique times.", expected_type=int)
+        # Note that in the past (pre-v0.5) Ntimes was the number of unique times in 
+        # either of the underlying visibilitydatasets (i.e. time_1_array and time_2_array),
```
adeliegorce, line 30: Missing space

adeliegorce: I'd rather say "the number of unique times in **a concatenation** of the underlying visibility datasets"

hera_pspec/uvpspec.py
```diff
@@ -26,7 +26,14 @@ def __init__(self):
...
+        # however throughout the code it was often assumed to mean the number of unique
+        # time-pairs (i.e. how many distinct average times there are in the data), which
+        # is really Ntpairs. In fact, there's never a situation in which you would need
+        # to know the number of unique underlying times in the context of power spectra,
+        # so now we make the two attributes identical.
```
adeliegorce, line 35: Should we just remove one? I guess `uvp.Ntimes`? Or we could keep this change for a future release

hera_pspec/uvpspec_utils.py
```diff
@@ -576,6 +577,8 @@ def _select(uvp, spws=None, bls=None, only_pairs_in_bls=False, blpairs=None,
         # if fed as list of tuples, convert to integers
         if isinstance(blpairs[0], tuple):
             blpairs = [uvp.antnums_to_blpair(blp) for blp in blpairs]
+        print("blps:", blpairs)
```
adeliegorce, line 580: Did you forget to remove those?

tests/test_grouping.py
adeliegorce, line 1: Took the opportunity to clean up some tests? :)

tests/conftest.py
adeliegorce, line 1: What is this new file for?

adeliegorce (commented):
Hi! Thank you for taking care of this. I made a few comments / asked a few questions. If everything runs on @Kai-FengChen's side, I'm all good with the PR.
I wonder if in the long run we should not just get rid of the `Ntimes` attribute?

hera_pspec/uvpspec.py
steven-murray: I've gone with "union" since we're talking about sets and uniqueness here :-)

hera_pspec/uvpspec.py
steven-murray: I definitely considered removing it, but I think maybe that would break more code than we intend to -- it might have quite significant downstream effects. I'm already scared about re-interpreting the meaning. I think I got all the usages inside the pspec repo sorted out, but I don't know about usages outside the repo, in pipeline code...

hera_pspec/uvpspec_utils.py
steven-murray: Yup! Thanks

hera_pspec/uvpspec.py
```diff
@@ -1411,9 +1418,9 @@ def read_from_group(self, grp, just_meta=False, spws=None, bls=None,
         if reading_old_version:
             # If we are reading a power spectrum object that was written in the pre-Nbltpairs days
             # then make sure to correctly set Ntimes and Ntpairs.
-            self.Ntimes = len(np.unique(np.hstack([self.time_1_array, self.time_2_array])))
```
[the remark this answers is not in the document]
steven-murray: Ah, you know I should have fixed this line -- even when reading old files we should make the object behave as a new object!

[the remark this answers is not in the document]
steven-murray: Oh, I realized I _did_ fix this line. I think we still need the `reading_old_version` code, at least for now. I'm not sure how "old" that version really is and what files we have lying around that would require that reader.

hera_pspec/grouping.py
```diff
@@ -103,7 +103,8 @@ def sample_baselines(bls, seed=None):
 def average_spectra(uvp_in, blpair_groups=None, time_avg=False,
                     blpair_weights=None, error_field=None,
                     error_weights=None, normalize_weights=True,
-                    inplace=True, add_to_history=''):
+                    inplace=True, add_to_history='',
+                    time_tol: float = 1e-6):
```
Kai-FengChen, line 107: This might not be a problem for our current pipeline, but there are a couple more averaging functionality that calls `average_spectra` like `spherical_average` and `spherical_wf_from_uvp`. Do we want to add `time_tol` for these functions as well so that this can be passed to the hidden `average_spectra` calls within these functions?

hera_pspec/testing.py
adeliegorce: Oh I’m just saying, if you test your code with random lists instead of the time arrays, the output is always the same whether you use your complicated line or just the `time_1_array.size`

adeliegorce: The repeated pairs get counted several times? Here if you have two arrays (a1,a2,a3) and (b1,b2,b3) your set is {(a1,b1), (a2,b2), (a3,b3)} because if the `zip` so the length is 3 - same as your input arrays. Or did I miss something?

steven-murray: Yes but in most cases the two arrays are like

```
time_1_array = (a1, a2, a3, a1, a2, a3)
time_2_array = (b1, b2, b3, b1, b2, b3)
```

where the second repetition corresponds to a different baseline pair (but the same set of times)

and in this case the length of the set will be 3 but the length of `time_1_array` is 6.

adeliegorce (approved):

PR #428, opened 2025-08-04
Handling non-square window functions

### LINKED ISSUE #415: More flexibility in Window Function (and covariances?) data shape
Currently window functions and covariances are always of size `(Nbltpairs, Ndly, Ndly, Npolpair)`. However, there are (approximate) cases in which the WFs and covariances are not dependent on either time or baseline, and so storing them in this configuration can be hugely inefficient. A couple of solutions might exist:

1. Allow them to be shape `(Nbltpairs or Ntimes or Nblpairs or 1, Ndly, Ndly, Npolpair)`, and then have an extra parameter on the `UVPSpec` object that tells it which of these forms it actually is. 
2. Have a separate `UVWindow` class (which actually already exists, but I'm proposing something a little different I think) that is specifically made to hold data about the window function, and can deal with these different shapes.

Similarly for the covariance.

### LINKED ISSUE #417: Arbitrary-sized window functions
Window functions should be able to have shape `(Nbltpairs, Ndly, Ndly2, Npolpairs)` -- i.e. the second delay-axis does not necessarily need to be the same as the first. Currently, there are at least some functions that assume that both delay axes are the same (e.g. the `fold_spectra` function). Of course, to do this, we'd need to also store the set of delays corresponding to the second delay-axis somewhere on the object. This could be implemented by creating a standalone `UVWindow` class or similar, as suggested in #415.

### LINKED ISSUE #426: Allow spherically-averaged exact WFs to not be square
While exact window functions don't need to be square (i.e. their shape `(Nbls, Ndelays, Nkperp, Nkpar, Npol)` does not have `Nkperp == Nbls` or `Ndelays==Nkpar`), the `spherical_wf_from_uvp` does in fact make the window function square (so it becomes shape `(1, Nkpar, Nkpar, 1)`.

We should allow it to be non-square so that the integral over "true" k's can be at a higher resolution.

adeliegorce:
This PR is to answer issue #426: I have modified `spherical_wf_from_uvp` to allow the output to be non-square, i.e. the spherical window functions have shape `(Nk, Nk_in)` where `Nk` is the number of "output" k-bins (average over the (kperp, kpar) corresponding to (bl_lens, delays) and `Nk_in` is the number of "input" k-bins (average over the (kperp, kpar) from the exact computation, derived from the beam simulation, etc.

To do this I have modified the input of `spherical_wf_from_uvp` such that:
- `kbin_widths` has been removed
- `kbins` has been replaced by `kbin_edges` to allow for non-uniform binning
- `kbin_edges_in` has been added, to describe the bin edges of the "input" k-bins
Some tests have been fixed, and others added, to account for this change.

### DIFF
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -1164,7 +1164,7 @@ class PSpecData:
 
         # Set the lru-cached get_Q_alt function, with a maxsize of Ndlys
         self._get_qalt_cached = lru_cache(maxsize=int(self.spw_Ndlys))(utils.get_Q_alt)
-        
+
 
     def cov_q_hat(self, key1, key2, model='empirical', exact_norm=False, pol=False,
                   time_indices=None):
@@ -3326,6 +3326,7 @@ class PSpecData:
                     else:
                         Mv, Wv = self.get_MW(Gv, Hv, mode=norm, exact_norm=exact_norm)
                     pv = self.p_hat(Mv, qv)
+                    # Gv.shape = Hv.shape = (nfreqs, nfreqs)
 
                     # Multiply by scalar
                     if self.primary_beam != None:
@@ -3380,7 +3381,12 @@ class PSpecData:
 
                     # store the window_function
                     if store_window:
+                        # Wv shape = (nfreqs, nfreqs) ie (64, 64)
+                        # qv shape = (nfreqs, ntimes) ie (64, 60)
                         pol_window_function.extend(np.repeat(Wv[np.newaxis,:,:], qv.shape[1], axis=0).astype(np.float64))
+                        # pol_wf shape = (ntimes, nfreqs, nfreqs) ie (60, 64, 64)
+                        # 4 blps so final wf_array shape for each spw is
+                        # (ntimes * nbls, nfreqs, nfreqs) = (4 * 60 = 240, 64, 64)
 
                     # Get baseline keys
                     if isinstance(blp, list):
@@ -3523,6 +3529,7 @@ class PSpecData:
         uvp.scalar_array = np.array(sclr_arr)
         uvp.channel_width = np.array(dset1.channel_width)  # all dsets validated to agree
         uvp.exact_windows = False
+        uvp.delay_avg = False
         uvp.weighting = input_data_weight
         uvp.vis_units, uvp.norm_units = self.units(little_h=little_h)
         # SGM: I've kept the same API in hera_pspec for now, but we should
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -3528,7 +3528,6 @@ class PSpecData:
         uvp.scalar_array = np.array(sclr_arr)
         uvp.channel_width = np.array(dset1.channel_width)  # all dsets validated to agree
         uvp.exact_windows = False
-        uvp.delays_are_binned = False
         uvp.weighting = input_data_weight
         uvp.vis_units, uvp.norm_units = self.units(little_h=little_h)
         # SGM: I've kept the same API in hera_pspec for now, but we should
--- a/hera_pspec/pspecdata.py
+++ b/hera_pspec/pspecdata.py
@@ -3326,7 +3326,6 @@ class PSpecData:
                     else:
                         Mv, Wv = self.get_MW(Gv, Hv, mode=norm, exact_norm=exact_norm)
                     pv = self.p_hat(Mv, qv)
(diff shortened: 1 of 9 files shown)


steven-murray (commented):
Thanks @adeliegorce, I think this looks correct. 

The default behaviour here is to produce square WFs when doing `spherical_average`. I wonder if there's a way to enable letting the output WFs be non-square when calling `spherical_average`? I'm actually not quite sure how this should work.

adeliegorce:
About the test on spherical average of window functions after delay binning:
I changed the spherical binning to make sure it agreed with the delay binning and things did not improve (see figures attached).
delay_avg_spherical_avg_binning
delay_avg_spherical_avg

adeliegorce:
I just committed changes which allow the spherical window functions to *not be square*, is the user can provide a `kbins_theory` to `grouping.spherical_average` which will be the spherical k-bins used by the theory. By default, they are taken identical to `kbins` and the wf is square. I still need to add tests for the functionality and make sure the delay averaging works as expected.

hera_pspec/testing.py
```diff
@@ -140,9 +141,17 @@ def build_vanilla_uvpspec(
...
+        Wv = np.diag(np.ones(uvp.Ndlys, dtype=np.float64))
+        window_function_array[s] = np.repeat(
+            # repeat along blpairtime axis
+            np.repeat(Wv[None, :], uvp.Nbltpairs, axis=0)[..., None],
+            uvp.Npols, axis=-1,  # repeat along polarization axis
         )
+        # else:
+        #     window_function_array[s] = np.ones(
+        #         (uvp.Nbltpairs, uvp.Ndlys, uvp.Ndlys, uvp.Npols), dtype=np.float64
+        #     )
```
steven-murray: can remove

hera_pspec/uvpspec.py
```diff
@@ -2459,6 +2472,7 @@ def combine_uvpspec(uvps, merge_history=True, verbose=True):
     store_window = np.all([hasattr(uvp, 'window_function_array') for uvp in uvps])
     exact_windows = np.all([uvp.exact_windows for uvp in uvps])
     store_stats = np.all([hasattr(uvp, 'stats_array') for uvp in uvps])
+    delay_avg = np.all([uvp.delay_avg for uvp in uvps])
```
steven-murray: Should we have a check that EITHER all uvps are delay-averaged or all are NOT delay-averaged?

steven-murray (commented):
Thanks @adeliegorce, this is great!

I have made a few comments on particular lines. Beyond these, I have the following comments:

1. The tests could use more comments/docstrings to make the *intent* more clear.  would like to be able to read through the test module in english and understand what is being tested and why.
2. Many of the tests seem to not actually have any assertions (they seem to be basically just be testing that the code runs without exceptions being raised). This is maybe OK, but it would be good to think of whether there are ways to make clear assertions about the results. If not, just comment that all you're doing is testing that it runs through.
3.

hera_pspec/pspecdata.py
```diff
@@ -3523,6 +3529,7 @@ def pspec(self, bls1, bls2, dsets, pols, n_dlys=None,
         uvp.scalar_array = np.array(sclr_arr)
         uvp.channel_width = np.array(dset1.channel_width)  # all dsets validated to agree
         uvp.exact_windows = False
+        uvp.delay_avg = False
```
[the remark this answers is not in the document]
adeliegorce: Ok, I will change it to a property. 
Regarding your second point, this is why so far I had the `delays_are_binned` attribute set as `False` by default and only modified when the `UVPSpec` object goes through the `grouping.average_in_delay_bins` method. Does that not make sense?

hera_pspec/uvpspec.py
adeliegorce: Agreed, changed to `np.any`

hera_pspec/pspecdata.py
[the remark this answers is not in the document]
steven-murray: Well, if we can't "detect" whether it's been delay-averaged after the fact, then it can't be a property. However, in that case I would ask why we need to know it? If its effect is invisible to the API, what do we need to store it for?

adeliegorce:
Hi @steven-murray, the PR is ready after implementing the changes we discussed last week, that is basically removing the `uvp.delays_are_binned` attribute and generalizing the functions it was used for (`grouping.fold_spectra`, `grouping.combine_uvpspec`) by defining a `Nkpar = uvp.window_function_array[spw].shape[2]`.

steven-murray (approved):

PR #429, opened 2025-10-15
Enable merging in batches, which may save memory

jsdillon:


steven-murray (approved):
This looks all good to me. It's still set as draft -- @jsdillon is it ready to go?

PR #430, opened 2025-10-17
Bug fix in the calculation of lst_avg_array

jsdillon:
I've encountered an unlikely scenario where the lst_1_array starts just before 0 and then wraps through it while lst_2_array starts just after 0. As a result, the lst_avg_array is off by pi!

image

This PR fixes that bug and adds a unit test to expose the issue.


steven-murray (commented):
This looks good to me, though I would _prefer_ if you extracted a `circular_average` function and used it in both of the cases you have here, and could then test that function directly without creating complicated PSpec objects. 

Happy to be talked out of it though.

jsdillon:
@steven-murray I've implemented your suggestion. Wanna take a look?

PR #433, opened 2026-04-06
deps: update code for x_orientation for new pyuvdata 3.2

steven-murray:
This updates access to the `.x_orientation` attribute in line with recent pyuvdata updates.


adeliegorce (commented):
Thank you for these changes that were long overdue @steven-murray. Will approve the PR once the tests pass (which is independent from the `x_orientation` changes I believe).

adeliegorce (approved):
The only tests failing are warning tests - these will need to be fixed eventually but are ok for now. PR approved

PR #442, opened 2026-04-20
Add label inventory and refresh uv development docs

kartikmandar:
This PR adds a declarative GitHub label inventory and updates the contributor docs to reflect the current uv-based development workflow.

It includes:
.github/labels.yml with the initial repo label set
README/docs install and development updates for Python >=3.11,<3.14
documented uv sync --all-extras --dev, uv run pytest, and uv run pytest -Werror
small Sphinx/docstring fixes needed to make docs validation pass cleanly

Validation:
python -m sphinx -W -b html docs docs/_build/html
python -m pytest


bhazelton:
I have learned the hard way taking over a repo that uses `uv` that it takes some work to be sure that you are doing all the CI that you probably want to be doing. If your CI just uses the uv lock file you won't be testing against new versions of packages as they come out, so you won't know if they are breaking things. Your users, on the other hand, are most likely using conda (at least last I checked most HERA folks use conda), so they may pick up those newer packages, especially since you haven't pinned upper limits on your dependencies in your pyproject.toml. So your users may find problems before your CI does.

I personally prefer for my CI to find these things before users do, which means that you need to have at least some CI jobs where you don't use the lock file. What I do is that in addition to my CI jobs that run against the lock file, I have one job that uses `uv sync --upgrade` and one that uses `uv sync --resolution lowest-direct` so that I make sure to test against the newest packages and I make sure that my minimum specified dependencies still work.

docs/index.rst
```diff
@@ -16,15 +16,14 @@ A set of `example Jupyter notebooks <https://github.com/HERA-Team/hera_pspec/tre
 Installation
 ------------
 
-The package is installable, along with its dependencies, with PyPi. We
-recommend using Anaconda and creating a new conda environment before
-installing the package:
+The package is installable from PyPI and supports Python 3.11, 3.12, and
```
steven-murray: I'd prefer not to encode python dependencies here. Let them just be in the pyproject.toml. Otherwise this will get stale.

docs/index.rst
```diff
@@ -16,15 +16,14 @@ A set of `example Jupyter notebooks <https://github.com/HERA-Team/hera_pspec/tre
...
+3.13.
 
 ::
 
-   $ conda create -n hera_pspec python=3
-   $ conda activate hera_pspec
-   $ python3 -m pip install hera_pspec
+   $ python -m venv hera_pspec
+   $ source hera_pspec/bin/activate
+   $ python -m pip install hera_pspec
```
steven-murray: This is not using uv. That's OK but a bit weird to me to change this to *neither* conda nor uv.

README.rst
```diff
@@ -21,18 +21,15 @@ Installation
...
 
 ::
 
-   $ conda create -n hera_pspec python=3
-   $ conda activate hera_pspec
-   $ conda install jax
-   $ python3 -m pip install hera_pspec
+   $ python -m venv hera_pspec
+   $ source hera_pspec/bin/activate
+   $ python -m pip install hera_pspec
```
steven-murray: Same comments here as for the index.rst. Actually, why is this information repeated in the index.rst instead of simply read from the README?

kartikmandar:
> I have learned the hard way taking over a repo that uses `uv` that it takes some work to be sure that you are doing all the CI that you probably want to be doing. If your CI just uses the uv lock file you won't be testing against new versions of packages as they come out, so you won't know if they are breaking things. Your users, on the other hand, are most likely using conda (at least last I checked most HERA folks use conda), so they may pick up those newer packages, especially since you haven't pinned upper limits on your dependencies in your pyproject.toml. So your users may find problems before your CI does.
> 
> I personally prefer for my CI to find these things before users do, which means that you need to have at least some CI jobs where you don't use the lock file. What I do is that in addition to my CI jobs that run against the lock file, I have one job that uses `uv sync --upgrade` and one that uses `uv sync --resolution lowest-direct` so that I make sure to test against the newest packages and I make sure that my minimum specified dependencies still work.

Should I open another PR to do  uv sync --upgrade and uv sync --resolution lowest-direct alongside the existing lockfile job ?

steven-murray:
@kartikmandar yes that sounds like a good idea to open a new PR with those updates to the CI.

In the meantime, this is looking very good now -- I just wanted to point you towards some ideas for labels that I use on some of my other repos: https://github.com/edges-collab/edges-analysis/blob/main/.github/labels.yml

You'll see I also use the `type: xxx` label content, as well as a `priority: ` context. I think these can be helpful for project/milestone planning. Might be worth taking a look and grabbing the ones you feel are worthwhile for pspec.

kartikmandar:
> @kartikmandar yes that sounds like a good idea to open a new PR with those updates to the CI.
> 
> In the meantime, this is looking very good now -- I just wanted to point you towards some ideas for labels that I use on some of my other repos: https://github.com/edges-collab/edges-analysis/blob/main/.github/labels.yml
> 
> You'll see I also use the `type: xxx` label content, as well as a `priority: ` context. I think these can be helpful for project/milestone planning. Might be worth taking a look and grabbing the ones you feel are worthwhile for pspec.

Sure, will check that out. Also, what should I do for these warning tests? These are mostly pyuvdata-related and can't really do much about them.

steven-murray:
Thanks @kartikmandar --- the warnings tests are known and its OK if they fail. I think there is another issue on the google doc to deal with that (separate PR)

.github/labels.yml
```diff
@@ -0,0 +1,92 @@
...
+  description: PSpecData, estimator setup, covariance models, and OQE execution.
+- name: "area: uvpspec"
+  color: "1d76db"
+  description: UVPSpec, HDF5 IO, selection, combination, and result metadata.
+- name: "area: grouping"
+  color: "1d76db"
+  description: Averaging, folding, bootstrapping, spherical binning, and delay-bin grouping.
+- name: "type: maintenance"
+  color: "0e8a16"
+  description: Tooling, cleanup, dependency updates, and non-science refactors.
```
steven-murray:
```suggestion

```

.github/labels.yml
```diff
@@ -0,0 +1,92 @@
...
+- name: "type: deprecation"
+  color: "0e8a16"
+  description: Planned removal or replacement of supported APIs or features.
+- name: "type: science"
+  color: "5319e7"
+  description: Scientific or statistical behavior, estimator math, normalization, and sensitivity logic.
```
steven-murray:
```suggestion
- name: "type: feature: science"
  color: "5319e7"
  description: Scientific or statistical behavior, estimator math, normalization, and sensitivity logic.
```

.github/labels.yml
```diff
@@ -0,0 +1,92 @@
...
+- name: "type: bug"
+  color: "d73a4a"
+  description: Something isn't working correctly.
+- name: "type: feature"
+  color: "0e8a16"
+  description: New analysis capability or user-facing functionality.
```
steven-murray:
```suggestion
- name: "type: feature: ux"
  color: "0e8a16"
  description: New user-facing functionality.
```

.github/labels.yml
```diff
@@ -0,0 +1,92 @@
...
+- name: "type: testing"
+  color: "b1fc6f"
+  description: Test additions, fixtures, and test-infrastructure changes.
+- name: "type: performance"
+  color: "016175"
+  description: Walltime, memory, or scaling improvements.
```
steven-murray:
```suggestion
- name: "type: performance: memory"
  color: "016175"
  description: Memory improvement
- name: "type: performance: speed"
  color: "016175"
  description: Wall-time performance improvements

```

steven-murray (approved):

PR #455, opened 2026-05-04
ux: don't always emit warning for Jy->mK

### LINKED ISSUE #432: REmove annoying warning that you can't convert to mK
There is an annoying warning that happens when you try to do `ds.Jy_to_mK()` on a dataset that is already in mK. In my opinion, no warning should be raised since there's nothing wrong with just ignoring conversion when it's already converted!

steven-murray:
Fixes #432


hera_pspec/pspecdata.py
```diff
@@ -3752,9 +3752,10 @@ def Jy_to_mK(self, beam=None):
         for i, dset in enumerate(self.dsets):
             # check dset vis units
             if dset.vis_units.upper() != 'JY':
-                warnings.warn(
-                    f"Cannot convert dset {i} Jy -> mK because vis_units = {dset.vis_units}"
-                )
+                if dset.vis_units.upper() != 'MK':
```
adeliegorce, line 3755: What if the `dset` is in K and not mK?

adeliegorce (approved):
Minor question, otherwise approved, thanks!

hera_pspec/pspecdata.py
steven-murray: Well, then I think it would still be good to raise the warning -- because you expect the output of this function to be in mK!

Issue #146, opened 2018-07-03
Data directory is starting to get large


philbull:
The test data directory (`hera_pspec/data`) that we include with hera_pspec is starting to get quite large. Is there any way we can thin it down a little, maybe by reusing some data files for multiple tests?

philbull:
It looks like the main culprits are the beam files:
```sh
-rw-rw-r-- 1 phil phil 20M Nov  5 15:10 hera_pspec/data/HERA_NF_dipole_power.beamfits
-rw-rw-r-- 1 phil phil 20M Nov  5 15:10 hera_pspec/data/HERA_NF_efield.beamfits
-rw-rw-r-- 1 phil phil 20M Nov  5 15:10 hera_pspec/data/HERA_NF_pstokes_power.beamfits
```
They are all currently used by the tests in `test_pspecbeam.py` however. The `HERA_NF_dipole_power.beamfits` file is used more widely in the tests; `efield` is used only in `test_pspecbeam.py`, while `pstokes_power` is used there and in `test_noise.py`.

PR #457, opened 2026-05-04
add basic pre-commit checks

### LINKED ISSUE #311: Add linting checks in CI
To enforce your desired code style. This really helps with code readability.

You can even do this with a pre-commit hook which makes is really easy for contributors to match your style. See the RASG repos for examples.

steven-murray:
Partially addresses #311, though we will need to add some ruff checks as well to finalise.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+exclude: '(?i)fits?$'
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v6.0.0
+  hooks:
+  - id: trailing-whitespace
+  - id: check-added-large-files
```
adeliegorce, line 7: This one might be a problem as we tend to add "large" files for our tests.. there is an issue raised about that, #146. In the meantime should we do without?

adeliegorce (approved):

adeliegorce:
Thank you! this is a good start on #311, we'll have to think about adding `black` `pylint` and `flake8` eventually

steven-murray:
Thanks @adeliegorce that is a good point. I actually think it's best to keep this check to prevent ourselves from adding more big files. As that issue you linked to mentions, it would be best to actually cut down on the big files we already have. Currently, since those beam files are already in the repo, this check doesn't fail -- it would only fail for new files checked in.

PR #458, opened 2026-05-04
chore: src/ layout + minimal ruff + AGENTS.md

kartikmandar:
- Move the package to a `src/` layout (`hera_pspec/` → `src/hera_pspec/`). Done as a pure-rename commit so `git log --follow` and GitHub blame keep working.
- Add a minimal ruff configuration (`E + W + F` with pragmatic ignores) and apply safe autofixes. Fixes 4 latent `F821` undefined-name bugs that ruff caught.
- Add `AGENTS.md` (canonical) and `CLAUDE.md` (`@AGENTS.md` import).


src/hera_pspec/data/_test_utils.yaml
```diff
@@ -1,6 +1,6 @@
 
 data:
-    root:       ../hera_pspec/
+    root:       ../
```
steven-murray: I wonder if this is the correct path? Shouldn't it be `../src/hera_pspec/`?

steven-murray (approved):
Thanks @kartikmandar this is reall great. I have one small comment you should check, but otherwise I approve.

src/hera_pspec/data/_test_utils.yaml
kartikmandar: Yeah, you are right.

adeliegorce:
Shouldn't this PR be to merge into the #development_v1 branch?

steven-murray:
@adeliegorce I don't think so.

The way I like to do it, at least for repos that are used somewhat broadly, is that any PR that either fixes a bug or does maintenance can just go straight into main. PRs that add new features should probably get pulled into a running release branch (e.g. v1.1) until that release is deemed ready. However, for repos like this where there is not a lot of people using it beyond those actively developing it, even that can be relaxed, IMHO. 

What is important is that any *breaking* changes be merged into a release branch until we decide its ready. This helps not to break people's workflows all the time, providing fewer discrete points at which people need to update their own code. 

This PR is just formatting, so no new features nor breaking changes.

PR #463, opened 2026-05-05
add uv-sync upgrade and lowest-direct test jobs

kartikmandar:
Per @bhazelton's review on #442, adds two CI jobs to `tests.yml`:

- **`tests-min-deps`** (ubuntu, py3.11): `uv sync --resolution lowest-direct --all-extras --dev`. Tests against the lowest declared-compatible version of each direct dependency.
- **`tests-upgraded`** (ubuntu, py3.13): `uv sync --upgrade --all-extras --dev`. Catches breakage from new package releases before users hit it.

Both blocking. Existing matrix job and codecov upload unchanged.

Also pins missing/stale dependency floors so `lowest-direct` actually passes:

- New floors: `scipy`, `pyyaml`, `h5py`, `uvtools`, `typer`, `tqdm`, `ipython`, `nbsphinx`, `sphinx_autorun`, `pytest-cases`.
- Bumped stale floors: `matplotlib 3.0 → 3.9.0`, `astropy 6.0 → 6.1.0`, `pytest 3.5.1 → 8.0`, `pytest-cov 2.5.1 → 5.0`, `coverage 4.5.1 → 7.0`.

`uv.lock` regenerated; no resolved versions changed. Full pytest suite (251/251) verified locally under both new jobs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,8 +30,8 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=2.0",
-    "scipy>=1.14.0",
-    "matplotlib>=3.9.0",
+    "scipy>=1.13.0",
+    "matplotlib>=3.8.0",
     "pyuvdata>=3.2.5",
     "astropy>=6.1.0",
     "pyyaml>=6.0.1",
@@ -66,7 +66,7 @@ docs = [
 ]
 tests = [
     "coverage>=7.0",
-    "pytest>=8.0,<10.0",  # pytest 9 breaks pytest-cases (https://github.com/smarie/python-pytest-cases/issues/374)
+    "pytest>=8.0",
     "pytest-cov>=5.0",
     "pytest-cases>=3.8.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,16 +30,16 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=2.0",
-    "scipy",
-    "matplotlib>=3.0",
+    "scipy>=1.14.0",
+    "matplotlib>=3.9.0",
     "pyuvdata>=3.2.5",
-    "astropy>=6.0",
-    "pyyaml",
-    "h5py",
-    "uvtools",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -30,16 +30,16 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=2.0",
-    "scipy",
-    "matplotlib>=3.0",
+    "scipy>=1.14.0",
+    "matplotlib>=3.9.0",
```
steven-murray: This seems like a very high lower bound for matplotlib -- are you sure we require this late a version? All others seem pretty reasonable, but this one seems too restrictive. Actually maybe scipy is also more restrictive than necessary?

I'm ok with things in the dependency-groups section being forced to be reasonably modern -- these are developer dependencies, and we can force developers to have a modern env. However, we shouldn't force users to have a modern env unless necessary, so the `optional-dependencies` section should be a minimum-spec'd.

steven-murray (commented):
Thanks @kartikmandar this is a really key addition. I have one comment -- and copilot has a few reasonable comments as well.

pyproject.toml
kartikmandar: Yeah, I checked a few different versions for any failures and then set a high enough lower bound. I will check what's the lowest I can go without breaking anything. Also, will see the Copilot suggestions.

kartikmandar: mpl 3.7.0 fails because matplotlib 3.7 and earlier ship a setuptools build (setup.py + setupext.py) that shells out to make to compile bundled C libraries (FreeType and Qhull).

mpl 3.8+ works because matplotlib 3.8 switched to the meson-python PEP 517 backend, which builds via meson instead of make+autoconf.

And scipy 1.13.0 was the first version with numpy 2 support, and uv resolves to that version because of the transitive numpy constraint.

steven-murray: Perfect, thanks!

steven-murray (approved):

PR #464, opened 2026-05-05
style: ruff format

steven-murray:
Applies ruff formatting. I set it to ignore magic trailing commas, because that resulted in some unnecessarily long white spacing.


kartikmandar (approved):
all looks good to me.

PR #479, opened 2026-05-06
refactor: address QoL improvements from #416

### LINKED ISSUE #416: Some QoL improvement suggestions
I strongly suggest we make the following Quality-of-Life improvements to the code, to make it more modern, easy to read, and reason about. I'm making a list here, instead of many many issues, because I think they can generally be implemented in one gross PR.

- [ ] Remove references to `object`, e.g. `class MyClass(object)` as they are completely unnecessary.
- [ ] Add typing to as many places as we can.
- [ ] Replace `OrderedDict` with standard `dict` everywhere (dicts are gauranteed to be ordered from Py3.7 on).
- [ ] Ensure we have a pre-pended "r" for strings containing escaped characters (e.g. math) so we don't get a bunch of unnecessary warnings on import.
- [ ] Use standard setuptools_scm-based versioning like `pyuvdata` and `hera_sim`.
- [ ] Remove all unused imports.
- [ ] Implement at least rudimentary style/correctness checking/linting with `ruff`

steven-murray:
## Summary
- replace remaining `OrderedDict` usage with plain `dict`
- remove unused imports while preserving package re-exports
- add targeted typing to core public interfaces in `noise`, `loss`, `version`, and the CLI

## Testing
- `MPLBACKEND=agg uv run pytest`
- `uv run --with ruff ruff check . --select F401`
- `MPLBACKEND=agg uv run pytest tests/test_cli.py tests/test_container.py tests/test_uvpspec.py`
- `MPLBACKEND=agg uv run pytest tests/test_noise.py tests/test_loss.py tests/test_cli.py`

Closes #416


adeliegorce (approved):
Thank you for this! out of curiosity: did you do it by hand?

PR #480, opened 2026-05-06
fix: warn when pspec_run produces no output

### LINKED ISSUE #244: No warning given when pspec_run does nothing.
I've had an issue where I pass a baseline-pair that doesn't exist (I got it backwards). But `pspec_run` doesn't say anything, just goes about its business and exits (but ends up not creating a file). It would be good to issue a warning in this case that the data selection doesn't contain any data.

steven-murray:
## Summary

- emit a `UserWarning` when `pspec_run` produces no output because the selected data contains no matching baseline-pairs
- cover both the early no-data load path and the later no-output path after baseline-pair validation
- extend the existing `test_pspec_run` cases to assert the warning

## Testing

- `uv run --with ruff ruff check src/hera_pspec/pspecdata.py tests/test_pspecdata.py`
- `MPLBACKEND=agg uv run pytest tests/test_pspecdata.py::test_pspec_run tests/test_pspecdata.py::test_get_argparser tests/test_pspecdata.py::test_get_argparser_backwards_compatibility`

Closes #244

Merge #479 first before reviewing

### DIFF
--- a/src/hera_pspec/pspecdata.py
+++ b/src/hera_pspec/pspecdata.py
@@ -4830,6 +4830,11 @@ def pspec_run(
         The PSpecData object used for OQE of power spectrum, with cached
         weighting matrices.
     """
+    no_output_warning = (
+        "pspec_run produced no output because the selected data contains no "
+        "matching baseline-pairs."
+    )
+
     # type check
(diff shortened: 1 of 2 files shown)


kartikmandar (approved):
Works fine for me in local too.

PR #481, opened 2026-05-06
fix: validate averaging API inputs

### LINKED ISSUE #317: average_spectra() does not check input
This comes from the code review. It is related to issue #281.

As I was trying out the code I called average_spectra() with an object of the wrong type (class). The first argument to average_spectra() should be a UVSpec object but I passed a PSpecData object. This gave an error:

AttributeError: 'PSpecData' object has no attribute 'antnums_to_blpair'

The error doesn’t indicate that the cause of the problem is passing an object of the wrong type. It appears that average_specta() does not check the type of the input object and provide a more informative error. Do all functions check their inputs?

steven-murray:
## Summary

- add informative top-level input validation for `grouping.average_spectra()` to fix the Issue #317 failure mode
- extend the same validation pattern to related high-level averaging/windowing APIs with similarly weak input checks
- add focused regression tests for the new negative-path error messages

## Testing

- `uv run --with ruff ruff check src/hera_pspec/grouping.py src/hera_pspec/uvpspec.py tests/test_grouping.py tests/test_uvpspec.py`
- `MPLBACKEND=agg uv run pytest tests/test_grouping.py::test_grouping_input_validation tests/test_grouping.py::TestGrouping::test_average_spectra tests/test_grouping.py::test_spherical_average tests/test_grouping.py::test_spherical_wf_from_uvp tests/test_uvpspec.py::TestUVPSpec::test_average_spectra tests/test_uvpspec.py::TestUVPSpec::test_get_blpair_groups_from_bl_groups_input_validation tests/test_uvpspec.py::TestUVPSpec::test_get_exact_window_functions tests/test_uvpspec.py::TestUVPSpec::test_fold_spectra`

Closes #317


src/hera_pspec/grouping.py
```diff
@@ -3,6 +3,8 @@
 import os
 import random
 import warnings
+from collections import OrderedDict as odict
```
kartikmandar: Is this being used in the file? Or maybe planning to later on?

steven-murray: Nope, this was a botched merge of main! Thanks for picking it up!

steven-murray:
```suggestion
```

kartikmandar (approved):

Issue #281, opened 2020-08-18
Invalid input data is not detected when plotting delay wedge


hughbg:
I use hera_pspec.plot.delay_wedge() to plot a delay wedge,  and get a Python error. The input data is unsuitable for generating a delay wedge; its unsuitability could be detected before a Python error is generated, and a message printed. This is minor.

The code I run is:
```
psc = hp.PSpecContainer("calibration/viscatBC_g.inco_psc.hdf5")
uvp = psc.get_pspec('dset0_dset1')
blpairs = uvp[0].get_red_blpairs()[0]
ax = hp.plot.delay_wedge(uvp[0], 0, (-5, -5), blpairs=[blpairs,]) 
```
The error produced when plotting is:
```
Traceback (most recent call last):
  File "analyse_sims.py", line 20, in <module>
    ax = hp.plot.delay_wedge(uvp[0], 0, (-5, -5), blpairs=[blpairs,])
  File "/data/home/apw737/gpu_poly/anaconda3_herasim_poly/lib/python3.7/site-packages/hera_pspec/plot.py", line 864, in delay_wedge
    y_edges = np.array([y_axis[0]-ydiff[0]/2.0] + list(y_axis[:-1]+ydiff/2.0) + [y_axis[-1]+ydiff[-1]/2.0])
IndexError: index 0 is out of bounds for axis 0 with size 0
```
ydiff is empty and I think it's because there aren't enough baseline pairs.

Issue #108, opened 2018-05-15
Can't plot with deltasq=True, fold=True, and delay=False


jsdillon:
```
fig,ax = plt.subplots(1,1, figsize=(5,8))
hp.plot.delay_spectrum(uvp, [((24, 25),(37, 38))], 0, 'XX', legend=True, ax=ax, deltasq=True, fold=True, delay=False)
```

This works if I flip any one of those three bits, but as written, I get:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-73-8a58b1c76510> in <module>()
      1 fig,ax = plt.subplots(1,1, figsize=(5,8))
----> 2 hp.plot.delay_spectrum(uvp, [((24, 25),(37, 38))], 0, 'XX', legend=True, ax=ax, deltasq=True, fold=True, delay=False)

/Users/jsdillon/anaconda2/lib/python2.7/site-packages/hera_pspec-0.0.0-py2.7.egg/hera_pspec/plot.pyc in delay_spectrum(uvp, blpairs, spw, pol, average_blpairs, average_times, fold, plot_noise, delay, deltasq, legend, ax)
     92     # Convert to Delta^2 units if requested
     93     if deltasq and not delay:
---> 94         uvp_plt.convert_to_deltasq()
     95 
     96     # Get x-axis units (delays in ns, or k_parallel in Mpc^-1 or h Mpc^-1)

/Users/jsdillon/anaconda2/lib/python2.7/site-packages/hera_pspec-0.0.0-py2.7.egg/hera_pspec/uvpspec.pyc in convert_to_deltasq(self, little_h, inplace)
    363 
    364             # multiply into data
--> 365             uvp.data_array[spw] *= k_mag**3 / (2*np.pi**2)
    366 
    367         # edit units

ValueError: operands could not be broadcast together with shapes (6,100,1) (6,49,1) (6,100,1) 
```

Issue #107, opened 2018-05-15
Clarify plotting documentation and legend


jsdillon:
https://github.com/HERA-Team/hera_pspec/blob/5dd299f4bc0f951e487f2f194ebf291687e62318/hera_pspec/plot.py#L20

First, "List of baseline-pair tuples" sounds like something like `[(1,2),(3,4)]`, when what you really mean is `[((1,2),(3,4))]`. Trying the former gives the rather unhelpful `TypeError: 'int' object has no attribute '__getitem__'`.

Second it's not clear that by spectral window you mean spectral window index, especially since I created my uvp object by specifying channels in a spw_range. In general, it's a bit unintuitive to figure out what the spws are and what they mean by looking to the attributes of a uvp object.

Third the legend (at least in the design review warmup) is pretty useless:
![screen shot 2018-05-15 at 4 33 40 pm](https://user-images.githubusercontent.com/5281139/40088860-cd10a9ba-585d-11e8-911a-0c05a55434e6.png)
It seems like at the very least you might want to show the LST too if you're not going to be smart in trying to make the legend show what's actually changing between the plotted spectra.

More generally, it's odd to me that you must specify which blpairs, spw, and pol to plot (that the default is not all of them) but that you apparently cannot specify which LSTs to plot...only whether to plot them all or average.

bhazelton:
Big plus one from me. I had a really hard time figuring this part out and I never ended up with a useful legend. Even when I set `average_times=True` I got 6 identical legend entries, but only 1 line showed, so I guess they were all on top of each other.

Also, the docs are kind of unhelpful, I never figured out that it wanted `[((1,2),(3,4))]`, I just used this from the docs: `[list(pspec.blpair_array),]` which was limiting. And the docs are slightly wrong in that they show using keywords for spw, pol rather than positional arguments.

PR #482, opened 2026-05-07
fix: resolve plotting issues from #447

steven-murray:
## Summary
- clarify delay-spectrum plotting inputs and docs for blpairs, spectral-window indexing, and time selection
- make `delay_spectrum()` generate smarter titles and legends by putting static metadata in the title and varying metadata in the legend
- add `title_legend=True` so callers can opt out of auto-generated title/legend text even when plotting onto an existing axes
- fix `delay_spectrum(..., deltasq=True, fold=True, delay=False)` so folded Delta^2 cosmology plots work
- make `delay_wedge()` raise an informative error when too few baseline pairs remain after selection

## Linked issues
- Closes #447
- Addresses #107
- Addresses #108
- Addresses #281

## Validation
- `uv run --with ruff ruff check src/hera_pspec/plot.py tests/test_plot.py`
- `MPLBACKEND=agg uv run pytest --no-cov tests/test_plot.py`
- `uv run --group docs sphinx-build docs docs/_build/html`
- targeted code review on the final diff


src/hera_pspec/plot.py
```diff
@@ -207,117 +361,141 @@ def delay_spectrum(
...
+            cax = ax.errorbar(x, _y, fmt="none", ecolor=c, yerr=item["errs"], **kwargs)
+
     # Set log scale
     if logscale:
         ax.set_yscale("log")
 
+    if title:
+        ax.set_title(title)
+
     # Add legend
```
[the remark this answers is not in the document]
steven-murray: @copilot please address this, updating both the comment and the docstring

docs/plot.rst
```diff
@@ -10,16 +10,21 @@ The following example plots the power spectra from a ``UVPSpec`` object, average
...
+                               average_blpairs=True, average_times=True,
+                               delay=False)
 
   # Setting delay=False plots the power spectrum in cosmological units
+  # Use times=uvp.time_avg_array[...] to select specific integrations.
+  # By default, static metadata is written to the title and only the
+  # varying metadata is written to the legend. Pass title_legend=False
+  # to disable the automatic title/legend text entirely.
 
 For a more extensive worked example, see `this example Jupyter notebook <https://github.com/HERA-Team/hera_pspec/blob/master/examples/Plotting_examples.ipynb>`_.
```
adeliegorce, line 29: The example notebook would need to be updated accordingly!

src/hera_pspec/plot.py
```diff
@@ -7,6 +7,143 @@
 from . import conversions, utils, uvpspec
 
 
+def _is_antpair_tuple(value):
```
adeliegorce, line 10: Isn't there a way to use an existing method like `pyuvdata.antpair2ind`?

src/hera_pspec/plot.py
```diff
@@ -7,6 +7,143 @@
...
+    """Return True for a baseline tuple like (ant1, ant2)."""
+    return (
+        isinstance(value, tuple)
+        and len(value) == 2
+        and all(isinstance(ant, (int, np.integer)) for ant in value)
+    )
+
+
+def _is_blpair_tuple(value):
```
adeliegorce, line 19: Same here with `uvp.antnums_to_blpair`?

adeliegorce (changes requested):
Hi! thank you for this! 
I only have a few minor questions + a request to update the plotting example notebook.
Could #107, #108 and #281 be closed with this PR as well now?

steven-murray:
@copilot Let's update the `delay_spectrum` function logic a little. When `average_blpairs` is True, we shouldn't add the blpairs list to the title (because that can be a very long list). This axis is being averaged over, so it doesn't serve a good purpose to list all blpairs in the title.

docs/plot.rst
steven-murray: Done!

src/hera_pspec/plot.py
steven-murray: No, this function is merely checking whether something looks like an antpair tuple. Basically just a validator.

src/hera_pspec/plot.py
steven-murray: No, again this is just a validator for inputs.

src/hera_pspec/plot.py
adeliegorce: Ok my point was that methods like `pyuvdata.antpair2ind` must check if the input is a tuple before converting it to an ind, such that it inherently does what `_is_antpair_tuple` albeit with an extra step. Either way it is fine as is

adeliegorce (approved):

PR #484, opened 2026-05-08
fix: clean up warnings test suite

### LINKED ISSUE #434: Many warnings tests fail
After adding warnings tests in #433, we get a _lot_ of errors, seemingly mainly due to some of the data files that exist in our tests having old formats that cause pyuvdata to raise warnings (and some of these will fail in pyuvdata v3.4). We should try to update these files so that we avoid warnings and keep current.

steven-murray:
## Summary

- refresh packaged beam and visibility fixtures that were emitting warnings under current dependencies
- replace several warning-producing code paths in `hera_pspec` and tighten warning-sensitive tests
- codify intentional warnings with explicit assertions so `pytest -Werror` passes consistently
- stabilize delay-bin stat averaging so large finite stats do not overflow in the warnings job on macOS

## Testing

- `MPLBACKEND=agg uv run pytest -Werror`

## Reviewer notes

- I left separate PR comments on a few production-code warning fixes that may deserve science-team review before merge.

Merge #483 first.

Fixes #434


src/hera_pspec/pstokes.py
jsdillon, line 295: I'd suggest warning when nsamples is 0 but flags is False

src/hera_pspec/pspecdata.py
```diff
@@ -4050,7 +4048,13 @@ def pspec(
...
                     # inverse avg is done b/c integ ~ 1/noise_var
                     # and due to non-linear operation of V_1 * V_2
-                    pol_ints.extend(1.0 / np.mean([1.0 / integ1, 1.0 / integ2], axis=0))
+                    pol_integ = np.zeros_like(integ1, dtype=float)
+                    valid_integ = (integ1 > 0) & (integ2 > 0)
+                    if np.any(valid_integ):
+                        pol_integ[valid_integ] = 2.0 / (
+                            (1.0 / integ1[valid_integ]) + (1.0 / integ2[valid_integ])
+                        )
+                    pol_ints.extend(pol_integ)
```
jsdillon, line 4066: Likewise, I'd maintain the warning here if integ1 or inteq2 is 0 but the corresponding flag is False

src/hera_pspec/pspecdata.py
```diff
@@ -3997,10 +3994,11 @@ def pspec(
...
                         # Wv shape = (nfreqs, nfreqs) ie (64, 64)
                         # qv shape = (nfreqs, ntimes) ie (64, 60)
+                        window_function = np.repeat(
+                            Wv[np.newaxis, :, :], qv.shape[1], axis=0
+                        )
                         pol_window_function.extend(
-                            np.repeat(Wv[np.newaxis, :, :], qv.shape[1], axis=0).astype(
-                                np.float64
-                            )
+                            np.real(window_function).astype(np.float64, copy=False)
```
adeliegorce, line 4001: Saving only the real part of the window functions is fine, they should be read. However it might be useful to write it in the docstring

src/hera_pspec/pstokes.py
```diff
@@ -286,7 +286,13 @@ def _combine_pol_arrays(
 
     # constructing nsamples
     if nsamples_list is not None:
-        combined_nsamples = 4 * (nsamples_list[0] ** -1 + nsamples_list[1] ** -1) ** -1
+        combined_nsamples = np.zeros_like(nsamples_list[0], dtype=float)
+        valid_nsamples = (nsamples_list[0] > 0) & (nsamples_list[1] > 0)
+        if np.any(valid_nsamples):
```
adeliegorce, line 297: If we use `np.where`, is this going to still raise a warning?

src/hera_pspec/uvpspec_utils.py
```diff
@@ -107,20 +107,22 @@ def subtract_uvp(uvp1, uvp2, run_check=True, verbose=False):
...
-                        ) + 1j * np.sqrt(cov1i.imag**2 + cov2i.imag**2)
+                    # add cov in quadrature: real and imag separately
+                    if hasattr(uvp1, "cov_array_real") and hasattr(
+                        uvp2, "cov_array_real"
+                    ):
+                        if uvp1.cov_model == uvp2.cov_model:
+                            cov1r = uvp1.get_cov(key1, component="real")
+                            cov2r = uvp2.get_cov(key2, component="real")
+                            uvp1.cov_array_real[i][blp1_inds, :, :, j] = np.sqrt(
+                                np.abs(cov1r) ** 2 + np.abs(cov2r) ** 2
```
adeliegorce: Why replacing `np.real` by `np.abs` here?

src/hera_pspec/uvpspec_utils.py
```diff
@@ -129,8 +131,8 @@ def subtract_uvp(uvp1, uvp2, run_check=True, verbose=False):
                     window1 = uvp1.get_window_function(key1)
                     window2 = uvp2.get_window_function(key2)
                     uvp1.window_function_array[i][blp1_inds, ..., j] = np.sqrt(
-                        window1.real**2 + window2.real**2
-                    ) + 1j * np.sqrt(window1.imag**2 + window2.imag**2)
+                        np.abs(window1) ** 2 + np.abs(window2) ** 2
```
adeliegorce: Same question here: if `window1` and `window2` are real (as imposed before), there should be no need for `np.abs`?

tests/conftest.py
```diff
@@ -77,6 +77,8 @@ def uvp_example_data() -> UVPSpec:
         taper="bh",
         verbose=False,
     )
+    uvp.cosmo = conversions.Cosmo_Conversions()
```
adeliegorce, line 80: Which warning was this object raising without a `cosmo` attribute?

adeliegorce (approved):
Hi! thanks a lot for these useful changes. I have a couple questions but everything makes sense to me. I should have answered your questions in the comments.

src/hera_pspec/uvpspec_utils.py
adeliegorce: They should be exactly equivalent here since the objects are real and you are taking the square? I don't get why you say there is a change of behaviour?

src/hera_pspec/uvpspec_utils.py
```diff
@@ -107,20 +107,22 @@ def subtract_uvp(uvp1, uvp2, run_check=True, verbose=False):
...
+                        if uvp1.cov_model == uvp2.cov_model:
+                            cov1r = uvp1.get_cov(key1, component="real")
+                            cov2r = uvp2.get_cov(key2, component="real")
+                            uvp1.cov_array_real[i][blp1_inds, :, :, j] = np.sqrt(
+                                np.abs(cov1r) ** 2 + np.abs(cov2r) ** 2
+                            )
+
+                            cov1i = uvp1.get_cov(key1, component="imag")
+                            cov2i = uvp2.get_cov(key2, component="imag")
+                            uvp1.cov_array_imag[i][blp1_inds, :, :, j] = np.sqrt(
```
adeliegorce, line 123: Makes sense to remove the `1j * np.sqrt(cov1i.imag**2 + cov2i.imag**2)` part since `cov_array_imag` is suppose to be an array of floats

src/hera_pspec/pstokes.py
steven-murray: Yeah, it does. Because when you use `where`, e.g.:

```
out = np.where(nsamples==0, 0, 1/nsamples)
```

the `1/nsamples` is evaluated _before_ the `np.where` function evaluates, so it gives the warning.

src/hera_pspec/uvpspec_utils.py
steven-murray: I'm actually totally confused about this now -- why would we add window functions in quadrature when subtracting power spectrum estimates? Shouldn't the window functions be subtracted from each other?

src/hera_pspec/uvpspec_utils.py
steven-murray: Yeah, this was very strange haha. I think it makes a lot more sense now, though I suppose I could also get rid of the unnecessary abs.

tests/conftest.py
steven-murray: The one in uvwindow (line 425)

src/hera_pspec/pspecdata.py
steven-murray: Done!

PR #491, opened 2026-05-20
Improvements to testing infrastructure

### LINKED ISSUE #414: Some proposed improvements to testing
Having read through some of the tests, here are a couple of thoughts about things that could be modernized/improved:

- [ ] Don't use UnitTest anywhere
- [ ] Use `tmpdir` and `tmp_path` instead of creating files and then deleting them manually.
- [ ] Use `pytest.parametrize` instead of in-test for-loops.
- [ ] Split up a bunch of long tests into smaller units.

adeliegorce:
Closes #414 

This PR implements improvements to the testing infrastructure required by #414 such as
- [x] Don't use UnitTest anywhere
- [x] Use tmpdir and tmp_path instead of creating files and then deleting them manually.
- [x] Use pytest.parametrize instead of in-test for-loops.
- [x] Split up a bunch of long tests into smaller units.
- [x] Contextualise `pytest.raises` calls

It is going to be a chunky one so we are starting with a draft PR.


tests/test_noise.py
```diff
@@ -10,69 +10,64 @@
...
-    Test noise.Sensitivity object
-    """
+@pytest.fixture
+def sensitivity_setup():
+    cosmo = conversions.Cosmo_Conversions()
+    beam = pspecbeam.PSpecBeamUV(
+        os.path.join(DATA_PATH, "HERA_NF_pstokes_power.beamfits")
+    )
+    sense = noise.Sensitivity(beam=beam, cosmo=cosmo)
+    return SimpleNamespace(cosmo=cosmo, beam=beam, sense=sense)
```
steven-murray: It would make more sense to make cosmo, beam and sense different fixtures, I think!

tests/test_pstokes.py
```diff
@@ -16,210 +16,197 @@
...
-            nsamples_list=[uvd1.nsample_array, uvd2.nsample_array],
+@pytest.fixture
+def pstokes_setup():
+    uvd1 = pyuvdata.UVData()
+    uvd1.read_miriad(dset1)
+    setattr(uvd1, "vis_units", "Jy")
+    uvd2 = pyuvdata.UVData()
+    uvd2.read_miriad(dset2)
+    setattr(uvd2, "vis_units", "Jy")
+    return SimpleNamespace(uvd1=uvd1, uvd2=uvd2)
```
steven-murray: Same comment here -- I'd make uvd1 and uvd2 different fixtures

tests/test_uvwindow.py
```diff
@@ -17,852 +17,843 @@
...
+        pol=pol,
+        spw_range=spw_range,
+        verbose=verbose,
+        x_orientation=x_orientation,
+        ftbeam_obj=ftbeam_obj,
+        data=data,
+        mapsize=mapsize,
+        freq_array=freq_array,
+        bandwidth=bandwidth,
+    )
```
steven-murray: same comment here!

tests/test_uvwindow.py
```diff
@@ -17,852 +17,843 @@
...
+    kbins = (krange[1:] + krange[:-1]) / 2 * units.h / units.Mpc
+    uvd = UVData()
+    uvd.read(os.path.join(DATA_PATH, dfile), read_data=False)
+    reds, lens, _ = utils.get_reds(uvd, bl_error_tol=1.0, pick_data_ants=False)
+    return SimpleNamespace(
+        ft_file=ft_file,
+        pol=pol,
+        polpair=polpair,
+        spw_range=spw_range,
+        taper=taper,
```
steven-murray: This is a lot of setup -- I'm totally cool with using class-based tests, just not the `unittest` variety (rather use `pytest` class tests). Using a class would help a lot with this very extended set up.

tests/test_uvwindow.py
```diff
@@ -17,852 +17,843 @@
...
+    # tests related to spw_range
+    test1 = uvwindow.FTBeam.from_file(
+        ftfile=ftbeam_setup.ft_file, spw_range=ftbeam_setup.spw_range
+    )
+    assert np.allclose(test1.freq_array, ftbeam_setup.freq_array)
+
+    test2 = uvwindow.FTBeam.from_file(ftfile=ftbeam_setup.ft_file, spw_range=None)
+    assert np.allclose(test2.freq_array, ftbeam_setup.bandwidth)
+
+    pytest.raises(
```
steven-murray: In general, we should prefer using

```
with pytest.raises(ErrorType, match="message to match"):
    <code>
```

instead of the non-context version of `pytest.raises` used throughout the code.

tests/test_uvwindow.py
```diff
@@ -17,852 +17,843 @@
...
+    assert test.verbose
+
+    # test on little_h
+    test = uvwindow.UVWindow(ftbeam_obj=uvwindow_setup.ft_beam_obj_spw, verbose=True)
+    assert test.kunits.is_equivalent(units.h / units.Mpc)
+    test = uvwindow.UVWindow(ftbeam_obj=uvwindow_setup.ft_beam_obj_spw, little_h=False)
+    assert test.kunits.is_equivalent(units.Mpc ** (-1))
+
+
+def test_uvwindow_from_uvpspec(uvwindow_setup):
```
steven-murray: In general, for tests like this that are currently trying to test all the functionality of a particular function, I would instead break the test up into much smaller chunks, each of which ideally tests ONE THING only. 

There are MANY examples of this (especially in `test_pspec` but you haven't touched that yet!)

tests/test_uvwindow.py
```diff
@@ -17,852 +17,843 @@
...
+    )
+
+
+def test_uvwindow_get_kgrid(uvwindow_setup):
+
+    bl_len = uvwindow_setup.lens[12]
+
+    # initialise object
+    test = uvwindow.UVWindow(ftbeam_obj=uvwindow_setup.ft_beam_obj_spw)
+    kgrid, kperp_norm = test._get_kgrid(bl_len)
```
steven-murray: I'm wary of these kinds of tests too -- they raise the coverage, but don't test anything except that the function doesn't blow up with an error. It would be better to actually test that it gives correct results!

tests/test_pstokes.py
adeliegorce: Done!

tests/test_noise.py
adeliegorce: Done!

tests/test_uvwindow.py
adeliegorce: I chose to use a few different fixtures instead of this extended setup, let me know if that's fine for you

tests/test_uvwindow.py
adeliegorce: I made the change for `test_uvwindow` and will keep that in mind in further modifications

src/hera_pspec/uvwindow.py
```diff
@@ -558,10 +558,10 @@ def _kperp4bl_freq(self, freq, bl_len, ngrid):
             Array of k_perp values to match to the FT of the beam.
 
         """
+        assert freq / 1e6 >= 1.0, "Frequency must be given in Hz."
```
steven-murray: assert statements should not be used to validate user input -- instead use an exception, i.e.

```suggestion
        if freq / 1e6 < 1.0:
            raise ValueError("Frequency must be given in Hz.")
```

tests/test_cli.py
```diff
@@ -41,139 +35,140 @@ def single_baseline_files(tmp_path_factory, vanilla_uvp: UVPSpec) -> list[Path]:
...
-            print(result.exc_info)
-            print(result.stdout)
-            assert result.exit_code == 0
-
-        # Verify the result has all the baselines
-        new = PSpecContainer(pth / "combined_single.pspec.h5", "r", keep_open=False)
-        newuvp = new.get_pspec("pspecgroup", "name")
-        assert all(blp in vanilla_uvp.get_blpairs() for blp in newuvp.get_blpairs())
-        assert len(newuvp.get_blpairs()) == len(vanilla_uvp.get_blpairs())
+def test_happy_path(vanilla_uvp: UVPSpec, single_baseline_files: list[Path]):
```
steven-murray: I would actually move these back into the class as they were originally, because soon we'll have #496 which will add more cli functions, and its best to keep them organized.

tests/test_container.py
```diff
@@ -8,313 +7,267 @@
...
-        fname = self.fname
-        group_names, pspec_names = self.group_names, self.pspec_names
-        ps_store = PSpecContainer(self.fname, mode="rw", keep_open=keep_open, swmr=swmr)
-
-        # Make sure that invalid mode arguments are caught
-        pytest.raises(ValueError, PSpecContainer, fname, mode="x")
-
-        # Check that power spectra can be overwritten
+@pytest.fixture
+def container_fname(tmp_path):
```
steven-murray:
```suggestion
def container_fname(tmp_path: Path) -> Path:
```

tests/test_container.py
```diff
@@ -8,313 +7,267 @@
...
+    """Empty PSpecContainer HDF5 file at a temporary path."""
+    fname = tmp_path / "_test_container.hdf5"
+    ps_store = PSpecContainer(fname, mode="rw", swmr=False)
+    del ps_store
```
steven-murray: is this `del` really necessary? The container should be removed from memory ocne it goes out of scope?

tests/test_container.py
```diff
@@ -8,313 +7,267 @@
...
+        assert psc.data.swmr_mode
+        psc._close()
+        psc = PSpecContainer(fname, mode=m, keep_open=True, swmr=False)
+        assert not psc.swmr
+        assert not psc.data.swmr_mode
+        psc._close()
 
 
-def test_combine_psc_spectra():
+def test_combine_psc_spectra(tmp_path):
```
steven-murray: This is maybe too much for this PR, but I would love to see it happen: instead of using files kept around in the data directory, it would be great if we could quickly make valid small UVPSpec objects on the fly to use in testing. 

Probably some of the tests should be using stored files -- those that are testing reading old file formats perhaps -- but in general it's much better to create these on the fly (and it means we can see what's IN THEM by looking at the code that's generating them!).

tests/test_container.py
```diff
@@ -8,313 +7,267 @@
...
     psc.set_pspec("grp1", "uvp_a", uvp1, overwrite=True)
     psc.set_pspec("grp1", "uvp_b", uvp2, overwrite=True)
     container.combine_psc_spectra(psc, dset_split_str=None, ext_split_str="_")
     assert psc.spectra("grp1") == ["uvp"]
 
     # test dset name handling
-    if os.path.exists("ex.h5"):
-        os.remove("ex.h5")
-    psc = PSpecContainer("ex.h5", mode="rw")
+    psc = PSpecContainer(tmp_path / "ex2.h5", mode="rw")
```
steven-murray: I would split this into a different test function since it's doing a different thing.

tests/test_conversions.py
```diff
@@ -1,103 +1,84 @@
...
-        np.testing.assert_almost_equal(tau2kpara, 503153.74952115043)
-
-
-if __name__ == "__main__":
-    unittest.main()
+@pytest.fixture
+def cosmo():
+    return conversions.Cosmo_Conversions(
+        Om_L=0.68440, Om_b=0.04911, Om_c=0.26442, H0=100.0, Om_M=None, Om_k=None
+    )
```
steven-murray, line 12: Does this not conflict with the fixture in the `conftest`? Can we make the name more explicit maybe, or just use the one in `conftest`?

tests/test_grouping.py
```diff
@@ -40,6 +38,50 @@ def case_vanilla_uvp_alternating_times(
...
+        uvd,
+        redundant_blpairs[:3],
+        spw_ranges=[(50, 100)],
+        beam=beam_nf_dipole_wcosmo,
+        cosmo=cosmo,
+    )
+
+
+@pytest.fixture
+def delay_bins_uvp_nocov(vanilla_uvp_with_beam):
```
steven-murray:
```suggestion
def delay_bins_uvp_nocov(vanilla_uvp_with_beam: UVPSpec) -> UVPSpec:
```

tests/test_grouping.py
```diff
@@ -130,385 +172,372 @@ def test_grouping_input_validation(vanilla_uvp: UVPSpec):
...
-        # change units of UVData objects
-        ds.dsets[0].vis_units = "mK"
-        ds.dsets[1].vis_units = "mK"
-
-        baselines = [(24, 25), (37, 38), (38, 39)]
-        # calculate all baseline pairs from group
-        baselines1, baselines2, blpairs = utils.construct_blpairs(
-            baselines, exclude_auto_bls=True, exclude_permutations=True
-        )
+def test_group_baselines():
```
steven-murray: For me, I'd make this a class `TestGroupBaselines` and then make each little block of code that does one test below into its own function in this class.

tests/test_grouping.py
```diff
@@ -130,385 +172,372 @@ def test_grouping_input_validation(vanilla_uvp: UVPSpec):
...
+    # Check that random seed works
+    g1 = grouping.group_baselines(bls5, 3, randomize=True, seed=10)
+    g2 = grouping.group_baselines(bls5, 3, randomize=True, seed=11)
+    g3 = grouping.group_baselines(bls5, 3, randomize=True, seed=10)
+    for i in range(len(g1)):
+        for j in range(len(g1[i])):
+            assert g1[i][j] == g3[i][j]
+
+
+def test_average_spectra(beam_nf_dipole_wcosmo):
```
steven-murray: Same comment as for `test_group_baselines`: make this `TestAverageSpectra`, move out all the stuff between `## Start file prep ##` and `## End file prep ##` as fixtures or setup, then make each individual test its own function that reuses this setup data.

tests/test_grouping.py
```diff
@@ -130,385 +172,372 @@ def test_grouping_input_validation(vanilla_uvp: UVPSpec):
...
+        blpair_weights=None,
+        error_field="noise",
+        error_weights=None,
+        normalize_weights=True,
+        inplace=False,
+        add_to_history="",
+    )
+
+
+def test_sample_baselines():
```
steven-murray: Same again here, though to a lesser extent (I'll stop commenting on this now, I think you see the pattern)

tests/test_grouping.py
```diff
@@ -1024,171 +1024,178 @@ def test_spherical_wf_from_uvp():
...
+        AssertionError, match="Need to compute exact window functions first"
+    ):
+        grouping.spherical_wf_from_uvp(
+            uvp, kbin_edges, little_h="h^-3" in uvp.norm_units
+        )
 
 
-class TestAverageDelayBins:
-    """Tests of the average_in_delay_bins function."""
+def test_get_delay_slices_errors():
```
steven-murray: Yeah, so again I would move this back into the class to organize it -- otherwise it's hard to tell what context these tests are in. 

If you prefer not to have classes, that's fine, then I would make a test *directory* for each source-code *module*, and one test *module* for each *function*, and one test per... test.

But I think having classes as an organizational tool is better.

tests/test_noise.py
```diff
@@ -10,69 +9,67 @@
 from hera_pspec.data import DATA_PATH
 
 
-class Test_Sensitivity(unittest.TestCase):
-    """
-    Test noise.Sensitivity object
-    """
+@pytest.fixture
+def cosmo():
+    return conversions.Cosmo_Conversions()
```
steven-murray: Unnecessary re-implementation of default `cosmo`.

tests/test_noise.py
```diff
@@ -10,69 +9,67 @@
...
-            Tsys, t_int, k=k, Ncoherent=1, Nincoherent=1, form="DelSq"
-        )
-        assert Dsq.shape == (10,)
-        assert Dsq[1] < P_N
+
+@pytest.fixture
+def beam():
+    return pspecbeam.PSpecBeamUV(
+        os.path.join(DATA_PATH, "HERA_NF_pstokes_power.beamfits")
+    )
```
steven-murray: this also feels like it is probably redundant (I can't remember)

tests/test_plot.py
```diff
@@ -34,786 +33,734 @@ def axes_contains(ax, obj_list):
...
+        fold=True,
+        legend=True,
+        label_type="blpair",
+    )
+    elements = [(mpl.lines.Line2D, 1), (mpl.legend.Legend, 1)]
+    assert axes_contains(f3.axes[0], elements)
+    plt.close(f3)
+
+
+def test_delay_spectrum_misc(uvp):
```
steven-murray: Split this up into the actual tests.

tests/test_plot.py
```diff
@@ -34,786 +33,734 @@ def axes_contains(ax, obj_list):
...
+        deltasq=False,
+        colorbar=False,
+        cbax=None,
+        vmin=None,
+        vmax=None,
+        edgecolor="none",
+        flip_xax=False,
+        flip_yax=False,
+        lw=2,
+    )
```
steven-murray: I would create a dict with these parameters in it, then reuse the dict for the following calls to `delay_wedge`, only updating the relevant parameters.

tests/test_pspecbeam.py
```diff
@@ -10,422 +9,349 @@
...
-        Om_p = beam.power_beam_int()
-        Om_pp = beam.power_beam_sq_int()
-        lower_freq = 120.0 * 10**6
-        upper_freq = 128.0 * 10**6
-        num_freqs = 20
-
-        # check pI polarization
+def test_init():
+    beamfile = os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits")
+    bm = pspecbeam.PSpecBeamUV(beamfile)
```
steven-murray: This doesn't test anything. Remove

tests/test_pspecbeam.py
```diff
@@ -10,422 +9,349 @@
...
+        "HERA_NF_efield.beamfits",
+        "isotropic_beam.beamfits",
+    ]
+
+    for beamfile in beamfiles:
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter("always")
+            pspecbeam.PSpecBeamUV(os.path.join(DATA_PATH, beamfile))
+
+        assert not caught, [str(w.message) for w in caught]
```
steven-murray: These should now be automatically tested since we added the new warnings tests. I would remove this.

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
-        beamfile = os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits")
-        self.bm = pspecbeam.PSpecBeamUV(beamfile)
-        self.bm.filename = "HERA_NF_dipole_power.beamfits"
 
-        # Load isotropic beam file
-        beamfile_Q = os.path.join(DATA_PATH, "isotropic_beam.beamfits")
-        self.bm_Q = pspecbeam.PSpecBeamUV(beamfile_Q)
-        self.bm_Q.filename = "isotropic_beam.beamfits"
+@pytest.fixture
+def d(_miriad_raw):
```
steven-murray: I would rename this to something more descriptive

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
-        # load another data file
-        self.uvd = uv.UVData()
-        self.uvd.read_miriad(os.path.join(DATA_PATH, "zen.2458042.17772.xx.HH.uvXA"))
 
-        self.uvd_std = uv.UVData()
-        self.uvd_std.read_miriad(
-            os.path.join(DATA_PATH, "zen.2458042.17772.std.xx.HH.uvXA")
-        )
+@pytest.fixture
+def d_std(_miriad_raw):
```
steven-murray: and this

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
-        Test the Q = dC/dp function.
-        """
-        vect_length = 50
-        x_vect = np.random.normal(size=vect_length) + 1.0j * np.random.normal(
-            size=vect_length
-        )
-        y_vect = np.random.normal(size=vect_length) + 1.0j * np.random.normal(
-            size=vect_length
+@pytest.fixture
+def w():
```
steven-murray: This seems dangerous --- I'm not sure what the point of a fixture that is a list of None's is.

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+    return [None, None]
+
+
+@pytest.fixture
```
steven-murray, line 166: Also I think its deprecated to use `pytest.fixture` without explicitly setting a scope.

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+
+
+def test_str(uvd):
+    """
+    Check that strings can be output.
+    """
+    ds = pspecdata.PSpecData()
+    print(ds)  # print empty psd
+    ds.add(uvd, None)
+    print(ds)  # print populated psd
```
steven-murray, line 473: this doesn't actually test anything

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+    ds = pspecdata.PSpecData()
+    ds.spw_Nfreqs = vect_length
+    # Test if there is a warning if user does not pass the beam
+    key1 = (0, 24, 38)
+    key2 = (1, 24, 38)
+    uvd = copy.deepcopy(uvd)
+    ds_t = pspecdata.PSpecData(dsets=[uvd, uvd])
 
-        x_vect = np.ones(vect_length)
+    for i in range(vect_length):
```
steven-murray: It's really better to not have for-loops if possible in tests (but I haven't checked if its possible here).

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+        ds.scalar(("xx", "yy"))
+
+    # Precomputed results in the following test were done "by hand"
+    # using iPython notebook "Scalar_dev2.ipynb" in the tests/ directory
+    # FIXME: Uncomment when pyuvdata support for this is ready
+    # scalar = ds.scalar()
+    # np.testing.assert_almost_equal(scalar, 3732415176.85 / 10.**9)
+
+    # FIXME: Remove this when pyuvdata support for the above is ready
+    # self.assertRaises(NotImplementedError, ds.scalar)
```
steven-murray:
```suggestion

```

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+
+def test_C_model():
+    # test the key format in ds._C and the shape of stored covariance
+    uvd = UVData()
+    uvd.read(os.path.join(DATA_PATH, "zen.even.xx.LST.1.28828.uvOCRSA"))
+    cosmo = conversions.Cosmo_Conversions()
+    uvb = pspecbeam.PSpecBeamUV(
+        os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits"), cosmo=cosmo
+    )
+    ds = pspecdata.PSpecData(dsets=[uvd, uvd], wgts=[None, None], beam=uvb)
```
steven-murray: all of this should be existing fixtures, I think

tests/test_pspecdata.py
```diff
@@ -102,2471 +105,2539 @@ def diagonal_or_not(mat, places=7):
...
+
+def test_get_analytic_covariance():
+    uvd = UVData()
+    uvd.read(os.path.join(DATA_PATH, "zen.even.xx.LST.1.28828.uvOCRSA"))
+    uvd.nsample_array[:] = 1.0
+    uvd.flag_array[:] = False
+    cosmo = conversions.Cosmo_Conversions()
+    uvb = pspecbeam.PSpecBeamUV(
+        os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits"), cosmo=cosmo
+    )
```
steven-murray: I think this is the same fixture :-)

tests/test_utils.py
```diff
@@ -7,10 +7,17 @@
...
-from hera_pspec import pspecbeam, testing, utils
+from hera_pspec import testing, utils
 from hera_pspec.data import DATA_PATH
 
 
+@pytest.fixture
+def uvd():
+    uvdata = UVData()
+    uvdata.read_miriad(os.path.join(DATA_PATH, "zen.2458042.17772.xx.HH.uvXA"))
+    return uvdata
```
steven-murray: Does this really need to be a different uvd than in the other modules?

tests/test_uvpspec.py
```diff
@@ -51,1280 +51,1257 @@ def case_uvp_exact_wfs(uvp_example_data, uvp_exact_wfs: UVPSpec):
...
+    i = uvp.get_integrations((0, ((1, 2), (1, 2)), ("xx", "xx")))
+    assert i.shape == (10,)
+    assert i.dtype == float
+    np.testing.assert_almost_equal(i[0], 1.0)
+
+    # get nsample
+    n = uvp.get_nsamples((0, ((1, 2), (1, 2)), ("xx", "xx")))
+    assert n.shape == (10,)
+    assert n.dtype == float
+    np.testing.assert_almost_equal(n[0], 1.0)
```
steven-murray: Things like this feel like they should be their own test that is parameterized.

tests/test_uvpspec.py
```diff
@@ -51,1280 +51,1257 @@ def case_uvp_exact_wfs(uvp_example_data, uvp_exact_wfs: UVPSpec):
...
+
+def test_combine_uvpspec_std(tmp_path: Path):
+    # setup uvp build
+    uvd = UVData()
+    uvd_std = UVData()
+    uvd.read_miriad(os.path.join(DATA_PATH, "zen.even.xx.LST.1.28828.uvOCRSA"))
+    uvd_std.read_miriad(os.path.join(DATA_PATH, "zen.even.xx.LST.1.28828.uvOCRSA"))
+    beam = pspecbeam.PSpecBeamUV(
+        os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits")
+    )
```
steven-murray: Another read that should be a fixture.

tests/test_uvwindow.py
```diff
@@ -17,876 +17,1029 @@
 outfile = "test.hdf5"
 
 
-class test_FTBeam(unittest.TestCase):
-    def setUp(self):
+@pytest.fixture()
+def make_ft_beam_obj():
```
steven-murray: This doesn't need to be a fixture -- it can just be a function.

tests/test_uvwindow.py
```diff
@@ -17,876 +17,1029 @@
...
+
+
+@pytest.fixture()
+def ft_bandwidth():
+    return uvwindow.FTBeam.get_bandwidth(os.path.join(DATA_PATH, ftfile))
+
+
+@pytest.fixture()
+def cosmo():
+    return conversions.Cosmo_Conversions()
```
steven-murray: redundant again

steven-murray (commented):
Thank you for this gargantuan effort @adeliegorce!

There are a lot of good things in this PR. In saying that, I've made a bunch of comments here. These are not meant to be exhaustive: almost all of them are representative of a broader pattern that you could ask your chosen AI to go and fix. 

I think the big thing to do at first is choose a consistent structure. I prefer a structure where a source module gets a test module, a source function gets a test class, and then there are very modular tests of that function inside the class. 

THe other big thing is to try and modularise the tests a lot more: currently there are some absolute humungous tests which should probably be broken down into smaller specific tests.

The other hting I noticed generally is that there is a lot of code repetition for test setup (reading the same beam file many times throughout the test base for example). We could probably slim down the total run time by cleaning this up. I think a targeted request to Claude to identify as many repeated bits of code as possible should help.

tests/test_pspecbeam.py
adeliegorce: Doesn't this check that the initialization of the object from a file works ok?

tests/test_utils.py
adeliegorce: No, the local uvd fixture cannot be replaced by any existing conftest fixture. The local fixture loads zen.2458042.17772.xx.HH.uvXA, while conftest has:
- uvd_zen_2458116 — loads zen.2458116.31939.HH.uvh5 (different file, different format)
- uvd_zen_even_xx — loads zen.even.xx.LST.1.28828.uvOCRSA (different file)

The tests using uvd (in test_calc_blpair_reds) assert specific baseline counts (e.g. len(bls1) == 12, blps == [((24, 25), (37, 38))], specific antenna indices like 24, 25, 37, 38, 39) that are tied to the exact array geometry of that file. Swapping in a different dataset would break those assertions.

steven-murray: OK good to know. Even so, I would prefer to name these uvd's more descriptively so we know what on earth they are and why we're using that one not something else!

tests/test_pspecbeam.py
steven-murray: Well, it tests that nothing explodes, yes. However there are other tests that also load this beamfile and actually use it, so we should just use those and remove this, I think.

tests/test_utils.py
adeliegorce: This is an excellent question to which I have no answer :) something to keep in mind..

src/hera_pspec/uvwindow.py
```diff
@@ -977,7 +978,9 @@ def get_cylindrical_wf(
 
         # normalisation of window functions
         sum_per_bin = np.sum(cyl_wf, axis=(1, 2))[:, None, None]
-        cyl_wf = np.divide(cyl_wf, sum_per_bin, where=sum_per_bin != 0)
+        cyl_wf = np.divide(
+            cyl_wf, sum_per_bin, out=np.zeros_like(cyl_wf), where=sum_per_bin != 0
+        )
```
steven-murray: It seems wrong to have the `out=` parameter as well as the return value here

src/hera_pspec/uvwindow.py
```diff
@@ -1043,7 +1046,9 @@ def cylindrical_to_spherical(self, cyl_wf, kbins, ktot, bl_lens, bl_weights=None
             )
             cyl_wf = cyl_wf[None]
         # check shapes are consistent
-        assert bl_lens.size == cyl_wf.shape[0]
+        assert bl_lens.size == cyl_wf.shape[0], (
+            "bl_lens size must match cyl_wf.shape[0]"
+        )
```
steven-murray: IN actual python code you should only use `assert` statements for debugging purposes, i.e. for conditions that should never happen. For checking bad user input you should use a normal exception, i.e.:

```python
if bl_lens.size != cyl_wf.shape[0]:
    raise ValueError("bl_lens size must match cyl_wf.shape[0]")
```

tests/test_cli.py
```diff
@@ -41,143 +39,75 @@ def single_baseline_files(tmp_path_factory, vanilla_uvp: UVPSpec) -> list[Path]:
...
+        for extra in extras:
+            with open(pth / f"{outpath_suffix}.{extra}.pkl", "rb") as fl:
+                data = pickle.load(fl)
+                assert all(blp in data for blp in vanilla_uvp.get_blpairs())
+                assert len(data) == len(vanilla_uvp.get_blpairs())
+
+
+def test_dummy_command(cli_runner: CliRunner) -> None:
+    """Check that the hello command runs and greets the user."""
+    result = cli_runner.invoke(cli.app, args=["hello"])
```
steven-murray: This can be removed (and the hello command as well, from the CLI)

tests/test_conversions.py
```diff
@@ -1,103 +1,110 @@
...
+        self,
+        cosmo_h25: conversions.Cosmo_Conversions,
+        method: str,
+        args: tuple,
+        kwargs: dict,
+        expected: float,
+    ) -> None:
+        """Check that little_h=True results are H0-independent, and little_h=False scales by H0 ratio."""
+        result = getattr(cosmo_h25, method)(*args, **kwargs)
+        np.testing.assert_almost_equal(result, expected)
```
steven-murray, line 97: I think the name of this test is wrong -- many of the parameterizations do not have anything to do with littleh

tests/test_grouping.py
```diff
@@ -21,26 +21,74 @@
 )
 from hera_pspec.data import DATA_PATH
 
+DATA_PATH = Path(DATA_PATH)
 
-def case_vanilla_uvp(vanilla_uvp: UVPSpec):
+
+def case_vanilla_uvp(vanilla_uvp: UVPSpec) -> UVPSpec:
```
steven-murray, line 27: Is there a reason we couldn't have just used the existing `vanilla_uvp` fixture intead of defining a new one?

tests/test_grouping.py
```diff
@@ -21,26 +21,74 @@
...
     return vanilla_uvp
 
 
-def case_vanilla_uvp_with_beam(beam_nf_dipole, vanilla_uvp_with_beam: UVPSpec):
+def case_vanilla_uvp_with_beam(vanilla_uvp_with_beam: UVPSpec) -> UVPSpec:
```
steven-murray, line 31: Same question for this one

tests/test_grouping.py
```diff
@@ -21,26 +21,74 @@
...
     return vanilla_uvp_with_beam
 
 
 def case_vanilla_uvp_alternating_times(
-    beam_nf_dipole, vanilla_uvp_alternating_times: UVPSpec
-):
+    vanilla_uvp_alternating_times: UVPSpec,
```
steven-murray, line 36: and this one

tests/test_grouping.py
```diff
@@ -21,26 +21,74 @@
...
-def test_grouping_input_validation(vanilla_uvp: UVPSpec):
+@pytest.fixture(scope="session")
+def redundant_blpairs(uvd_zen_even_xx: UVData) -> list:
+    """Redundant baseline groups from zen.even.xx.LST.1.28828.uvOCRSA."""
+    ap, a = uvd_zen_even_xx.get_enu_data_ants()
+    return redcal.get_pos_reds(dict(zip(a, ap)), bl_error_tol=1.0)
+
+
+@pytest.fixture(scope="session")
+def uvp_from_miriad(
```
steven-murray: maybe a better name here would be something like `uvp_three_red_bls`?

tests/test_grouping.py
```diff
@@ -130,145 +178,162 @@ def test_grouping_input_validation(vanilla_uvp: UVPSpec):
...
+        uvp.set_stats("noise", key, Pn[blp])
+    # Add the simple error bar (all are set to be one) to stat_array
+    errs = np.ones((uvp.Ntpairs, uvp.Ndlys))
+    for key in keys:
+        uvp.set_stats("simple", key, errs)
+    return uvp, blpairs
+
+
+@pytest.fixture(scope="session")
+def uvp_with_exact_wf(
```
steven-murray: I just want to make sure that there's a good reason not to use the existing uvp with exact window functions from further up in this module?

tests/test_pspecbeam.py
```diff
@@ -9,53 +10,58 @@
...
 
-    def tearDown(self):
-        pass
 
-    def runTest(self):
-        pass
+@pytest.fixture(scope="module")
+def pstokes_beam() -> pspecbeam.PSpecBeamUV:
+    """PSpecBeamUV built from the HERA_NF_pstokes_power beamfits file, for testing pStokes-polarization features."""
+    return pspecbeam.PSpecBeamUV(DATA_PATH / "HERA_NF_pstokes_power.beamfits")
```
steven-murray: I think this was already defined in `conftest`?

tests/test_pspecbeam.py
```diff
@@ -9,53 +10,58 @@
...
-    def test_init(self):
-        beamfile = os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits")
-        bm = pspecbeam.PSpecBeamUV(beamfile)
 
-    def test_beamfiles_load_without_warnings(self):
-        beamfiles = [
+@pytest.fixture
+def efield_beam() -> pspecbeam.PSpecBeamUV:
+    """An efield-type beam with beam_type force-set to 'power' so later validation steps in beam_normalized_response can be reached."""
+    beam = pspecbeam.PSpecBeamUV(DATA_PATH / "HERA_NF_efield.beamfits")
```
steven-murray, line 23: Could this just use an already-defined beam from conftest?

tests/test_pspecbeam.py
```diff
@@ -9,53 +10,58 @@
...
-            lower_freq, upper_freq, num_freqs, pol="pI", num_steps=2000
-        )
+        ],
+    )
+    def test_loads_without_warnings(self, beamfile: str) -> None:
+        """Check that loading each shipped beamfits file does not emit warnings."""
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter("always")
+            pspecbeam.PSpecBeamUV(DATA_PATH / beamfile)
+        assert not caught, [str(w.message) for w in caught]
```
steven-murray: since we have warnings tests now, this should not be necessary, especially since most of these are already fixtures

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
 
-        # Check that user-defined cosmology can be specified
+    def test_custom_cosmology(self) -> None:
+        """Check that a user-defined cosmology can be passed to the constructor."""
         beam = pspecbeam.PSpecBeamUV(
-            pstokes_beamfile, cosmo=conversions.Cosmo_Conversions()
+            DATA_PATH / "HERA_NF_pstokes_power.beamfits",
+            cosmo=conversions.Cosmo_Conversions(),
         )
+        assert isinstance(beam.cosmo, conversions.Cosmo_Conversions)
```
steven-murray, line 77: You can just pass in the existing fixture from conftest, and just check the assert statement

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
-            pol=pol,
-        )
+    @pytest.mark.parametrize("pol", ["pZ", "XX"])
+    def test_invalid_pol_raises(
+        self, pstokes_beam: pspecbeam.PSpecBeamUV, pol: str
+    ) -> None:
+        """Check that an unrecognized or wrong-basis polarization raises."""
+        # No match= because pyuvdata's polstr2num intercepts the invalid pol string
+        # before pspecbeam's own validation, and the exception type/message varies
+        # across pyuvdata versions (KeyError in older, ValueError in newer).
```
steven-murray: Here, just enforce the newer pyuvdata version in the pyproject.toml, and check that the ValueError is raised with appropriate `match=`

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
-            lower_freq,
-            upper_freq,
-            num_freqs,
-            pol="pI",
-        )
+    def test_dipole_beam(self, beam_nf_dipole: pspecbeam.PSpecBeamUV) -> None:
+        """Check that dipole (linear-pol) beams compute a scalar for XX but raise for pI."""
+        beam_nf_dipole.compute_pspec_scalar(LOWER_FREQ, UPPER_FREQ, NUM_FREQS, pol="XX")
+        with pytest.raises(
+            (KeyError, ValueError)
```
steven-murray: See above comment

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
-            num_freqs,
-            pol="pI",
-        )
+    def test_efield_beam(self) -> None:
+        """Check that efield beams compute a scalar for XX but raise for pI."""
+        beam = pspecbeam.PSpecBeamUV(DATA_PATH / "HERA_NF_efield.beamfits")
+        beam.compute_pspec_scalar(LOWER_FREQ, UPPER_FREQ, NUM_FREQS, pol="XX")
+        with pytest.raises(
+            (KeyError, ValueError)
+        ):  # see note above re pyuvdata versions
```
steven-murray: see above comment

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
+        freq = np.linspace(130.0 * 1e6, 140.0 * 1e6, 10)
+        with pytest.raises(ValueError, match="Do not have the right polarization"):
+            pspecbeam.PSpecBeamUV.beam_normalized_response(
+                beam_nf_dipole, pol="ll", freq=freq
+            )
+
+    def test_beam_normalized_response_raises_on_non_power_beam(self) -> None:
+        """Check that a non-power beam_type raises a ValueError."""
+        freq = np.linspace(130.0 * 1e6, 140.0 * 1e6, 10)
+        beam = pspecbeam.PSpecBeamUV(DATA_PATH / "HERA_NF_efield.beamfits")
```
steven-murray, line 218: this is a fixture

tests/test_pspecbeam.py
```diff
@@ -65,367 +71,455 @@ def test_UVbeam(self):
...
+        with pytest.raises(ValueError, match="only healpix format supported"):
+            pspecbeam.PSpecBeamUV.beam_normalized_response(
+                efield_beam, pol="xx", freq=freq
+            )
+
+
+@pytest.fixture(scope="module")
+def gauss_beam() -> pspecbeam.PSpecBeamGauss:
+    """A Gaussian primary beam (FWHM=0.8), used both as its own test subject and as a reference for PSpecBeamFromArray."""
+    return pspecbeam.PSpecBeamGauss(0.8, np.linspace(115e6, 130e6, 50, endpoint=False))
```
steven-murray, line 260: Not something to do in this PR, but just a note that the PSpecBeamGauss class should probably be updated to use the new pyuvdata AnalyticBeam class.

tests/test_pspecdata.py
```diff
@@ -102,276 +109,378 @@ def diagonal_or_not(mat, places=7):
...
-        d_float = 12.0
-        d_dict = {"(0,1)": np.arange(5), "(0,2)": np.arange(5)}
 
-        self.assertRaises(TypeError, pspecdata.PSpecData, d_arr, d_arr)
-        self.assertRaises(TypeError, pspecdata.PSpecData, d_lst, d_lst)
-        self.assertRaises(TypeError, pspecdata.PSpecData, d_float, d_float)
-        self.assertRaises(TypeError, pspecdata.PSpecData, d_dict, d_dict)
+@pytest.fixture
+def w():
+    return [None, None]
```
steven-murray: this seems like a pointless fixture

tests/test_pspecdata.py
```diff
@@ -102,276 +109,378 @@ def diagonal_or_not(mat, places=7):
...
+                _cache[key] = result
+        return _cache[key]
 
-        # Load datafiles
-        self.d = []
-        for dfile in dfiles:
-            _d = uv.UVData()
-            _d.read_miriad(os.path.join(DATA_PATH, dfile))
-            self.d.append(_d)
+    return _load
```
steven-murray: This function seems like over-engineering to me. Why not just go through the module, find all files that need to be read as miriad files, and make a one-line fixture for each that returns a UVData object with a nice name? The you can use them in any function easily (either in lists or not).

tests/test_pspecdata.py
```diff
@@ -102,276 +109,378 @@ def diagonal_or_not(mat, places=7):
...
-        # Test whether unequal no. of weights is picked up
-        self.assertRaises(
-            AssertionError,
-            pspecdata.PSpecData,
-            [uv.UVData(), uv.UVData(), uv.UVData()],
-            [uv.UVData(), uv.UVData()],
-        )
+@pytest.fixture
+def uvd_std(_miriad_raw):
+    return copy.deepcopy(_miriad_raw("zen.2458042.17772.std.xx.HH.uvXA"))
```
steven-murray: For things like this, I'd also lie to know if it's really necessary to use a different file than the ones in the dfiles lists above? What's different about this file that we really need?

tests/test_pspecdata.py
```diff
@@ -102,276 +109,378 @@ def diagonal_or_not(mat, places=7):
...
+def pspec_blpairs() -> tuple:
+    """Standard (bls1, bls2, blpairs) derived from pspec_bls."""
+    bls1, bls2, blpairs = utils.construct_blpairs(
+        pspec_bls, exclude_auto_bls=True, exclude_permutations=True
+    )
+    return bls1, bls2, blpairs
+
+
+@pytest.fixture
+def pspec_ds(beam_nf_dipole, uvd):
```
steven-murray:
```suggestion
def pspec_ds(beam_nf_dipole, uvd) -> pspecdata.PSpecData:
```

tests/test_pspecdata.py
```diff
@@ -493,92 +611,114 @@ def test_get_Q(self):
...
+    with pytest.warns(UserWarning, match="The beam response could not be calculated"):
+        ds_t.get_integral_beam(pol)
+
+    try:
+        integral_matrix = ds.get_integral_beam(pol)
+        # Test that if the number of delay bins hasn't been set
+        # the code defaults to putting that equal to Nfreqs
+        assert ds.spw_Ndlys == ds.spw_Nfreqs
+    except IndexError:
+        integral_matrix = np.ones((ds.spw_Ndlys, ds.spw_Ndlys))
```
steven-murray: I can't understand why we would need a try/except inside a test function, when we get to specify the input data?

tests/test_pspecdata.py
```diff
@@ -638,682 +779,744 @@ def test_get_unnormed_E(self):
...
-            NotImplementedError,
-            self.ds.get_MW,
-            random_G,
-            random_H,
-            mode="H^-1",
-            exact_norm=True,
-        )
+        for i in range(ds.spw_Ndlys):
+            for j in range(ds.spw_Ndlys):
+                assert frac_non_herm[i, j] <= tol
```
steven-murray: I think this could just be

```python
assert np.all(frac_non_herm<tol)
```

tests/test_uvwindow.py
```diff
@@ -1,892 +1,1043 @@
...
-        self.x_orientation = "east"
-
-        self.ftbeam_obj = uvwindow.FTBeam.from_file(
-            ftfile=self.ft_file, spw_range=self.spw_range
+@pytest.fixture()
+def make_ft_beam_obj() -> Callable[[tuple[int, int] | None], uvwindow.FTBeam]:
+    def _factory(spw_range: tuple[int, int] | None = None) -> uvwindow.FTBeam:
+        return uvwindow.FTBeam.from_file(ftfile=DATA_PATH / ftfile, spw_range=spw_range)
+
+    return _factory
```
steven-murray, line 29:
```suggestion
def make_ft_beam_obj(spw_range: tuple[int,int]) -> Callable[[tuple[int, int] | None], uvwindow.FTBeam]:
    return uvwindow.FTBeam.from_file(ftfile=DATA_PATH / ftfile, spw_range=spw_range)
```

tests/test_uvwindow.py
```diff
@@ -1,892 +1,1043 @@
...
+        ftbeam_obj=ft_beam_spw,
+        taper="blackman-harris",
+        cosmo=cosmo,
+        little_h=True,
+        verbose=False,
+    )
+
+
+@pytest.fixture(scope="module")
+def lens(uvd_zen_2458116: UVData) -> np.ndarray:
```
steven-murray:
```suggestion
def red_bl_lens(uvd_zen_2458116: UVData) -> np.ndarray:
```

steven-murray (commented):
Thanks so much @adeliegorce this is really great! I do have some more comments, though I think these are much less intense than previous rounds, so I think we're close :-)

I should say that, given the volume of the PR, I almost certainly missed some stuff. I would again treat my comments as indicative and use them to find patterns of similar stuff in the code, rather than ONLY fixing the specific things I noted.

src/hera_pspec/uvwindow.py
adeliegorce: This was to prevent some warnings from popping up. `out` only specifies the buffer that `numpy` writes results into, and here we pre-fill it with zeros. Where `where=False`, `numpy` leaves the buffer 0 and doesn't compute anything, so no warning fires. For positions where where=True, it divides normally and writes the result.

src/hera_pspec/uvwindow.py
adeliegorce: Ok, I made the change for the `uvwindow` module but not all the modules - should we create an issue for this?

tests/test_cli.py
adeliegorce: I removed the `test_dummy_command` but not `hello`.
When `hello` is removed, `fast_merge_baselines` becomes the only command and `Typer` stops treating it as a subcommand, so the related tests fail.
Seems easier to simply keep it.

tests/test_conversions.py
adeliegorce: Ok, I split this into two separate tests for clarity

tests/test_grouping.py
adeliegorce: You're right, `case_vanilla_uvp` is not defining anything new but it exists because `@parametrize_with_cases` needs a `case_*` function to collect, and it can't directly take in a plain `pytest` fixture (as far as I know)

tests/test_grouping.py
adeliegorce: Same answer :D

src/hera_pspec/uvwindow.py
steven-murray: I think maybe two things are being conflated here: `where=` does the job of figuring out which elements to compute, while `out=` provides an output buffer. However, the results being written into the buffer are never being used, because the thing you're passing into `out` is a temporary array. To avoid the warnings messages it should be enough to provide the `where=` argument. If you'd like to provide the `out=` argument as well, I would just pass in `cyl_wf` directly, rather than `np.zeros_like(cyl_wf)`.

tests/test_grouping.py
adeliegorce: Done

tests/test_cli.py
steven-murray: Yeah, good idea!

tests/test_grouping.py
steven-murray: Ah! I see.

tests/test_grouping.py
adeliegorce: Hm if you're referring to `delay_bins_exact_wf_uvp` vs. `uvp_with_exact_wf`:
the former starts from `vanilla_uvp_with_beam fixture` whilst the latter is computed from real visibilities (`uvd_zen_2458116`) via `PSpecData.pspec()`.

There is also a difference in the beam as `delay_bins_exact_wf_uvp` uses a Gaussian beam, while `uvp_with_exact_wf` uses the real HERA dipole beam file (`FT_beam_HERA_dipole_test`).

In the end `delay_bins_exact_wf_uvp` is more of a lightweight fixture for testing the delay-binning machinery

adeliegorce: Ok I replaced the former by `uvp_exact_wfs` which is defined in `conftest` but it needed some beam attributes so I added them

src/hera_pspec/uvwindow.py
adeliegorce: Ok! turns out I thought `where` was enough also but apparently not, the warnings are still triggered. I think having `out=np.zeros_like(cyl_wf)` or `out=cyl_wf` is perfectly equivalent since `cyl_wf=0` where `sum_per_bin==0`, so I'm fine with your plan!

tests/test_pspecbeam.py
adeliegorce: Nicely spotted! I removed it.

tests/test_pspecbeam.py
adeliegorce: I don't think so because of `beam.primary_beam.beam_type = "power"`

tests/test_pspecbeam.py
adeliegorce: Done! I hope it won't impact coverage.

tests/test_pspecbeam.py
adeliegorce: Not sure because this test is specifically checking that you can pass a `cosmo` argument to the constructor and have it stored. The conftest fixtures like `beam_nf_pstokes` are already constructed, so you can't use them to test the constructor's `cosmo` parameter.

So unless we simply remove this test I think it should stay as is

tests/test_pspecbeam.py
adeliegorce: Ok I think in the new version of the tests actually we don't even need this split anymore. I've removed it

tests/test_uvwindow.py
adeliegorce: I know this looks unnecessarily convoluted but from what I've seen a fixture function is called by `pytest` without arguments. You can't invoke `make_ft_beam_obj` like a normal function and pass it `spw_range`. The `_factory` infrastructure lets you work around this

tests/test_pspecdata.py
adeliegorce: This is a good point and looking at this made me realise some tests in this file were running on empty `PSpecData` objects. I fixed this, namely in `test_get_Q`

tests/test_pspecbeam.py
adeliegorce: Not sure I understand why?

tests/test_pspecbeam.py
adeliegorce: Ok I created issue #517 for this

tests/test_pspecdata.py
adeliegorce: 100% agree, removed it

tests/test_pspecdata.py
adeliegorce: Agreed, done.

tests/test_pspecdata.py
adeliegorce: Agreed, but beyond the scope of this PR ;) I have added a comment to #146 to remember to do this

steven-murray (approved):
🎉  Great work @adeliegorce thanks for all the effort!

PR #496, opened 2026-06-02
Migrate legacy pipeline scripts to native pspec typer commands (#467)

kartikmandar:
Replaces the `pspec_run`/`bootstrap_run`/`auto_noise_run`/`generate_pstokes_run` scripts with native `pspec` typer subcommands (`run`, `bootstrap`, `auto-noise`, `generate-pstokes`); deletes the now-unused argparse builders, scripts, and their `script-files` entries. Breaking: idiomatic arg syntax (e.g. `--spw-range 200 300` repeatable) and `hera_cal._cli_tools` profiling dropped for now. Fixes 4 latent bugs (bool flags for `symmetric_taper`/bootstrap stds, `auto-noise` unbound `spectra`, `generate-pstokes` `--pstokes` list default). 266 tests pass (incl. `-Werror`), ruff clean, `cli.py` 100% covered.


.github/workflows/tests.yml
```diff
@@ -2,9 +2,9 @@ name: Run Tests
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, release/v1 ]
```
steven-murray: I suggest we use the glob ability here, so `release/v*`

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+    dset_pair: list[str] | None = typer.Option(
+        None,
+        "--dset-pair",
+        click_type=click.Tuple([int, int]),
+        help="Dataset pairing for OQE, e.g. --dset-pair 0 1 (repeatable).",
+    ),
+    dset_label: list[str] | None = typer.Option(
+        None, "--dset-label", help="String label for each input dataset (repeatable)."
+    ),
+    spw_range: list[str] | None = typer.Option(
```
steven-murray: can we just make the type `tuple[int,int]` here instead?

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+        None,
+        "--spw-range",
+        click_type=click.Tuple([int, int]),
+        help="Spectral-window channel selection, e.g. --spw-range 200 300 (repeatable).",
+    ),
+    n_dlys: list[int] | None = typer.Option(
+        None, "--n-dlys", help="Number of delays per spectral window (repeatable)."
+    ),
+    pol_pair: list[str] | None = typer.Option(
```
steven-murray: again, can we just use `tuple[int,int]` as type?

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+        click_type=click.Tuple([str, str]),
+        help="Polarization-string pair, e.g. --pol-pair xx xx (repeatable).",
+    ),
+    blpair: list[str] | None = typer.Option(
+        None,
+        "--blpair",
+        click_type=click.Tuple([int, int, int, int]),
+        help="Baseline-pair antennas, e.g. --blpair 1 2 3 4 -> ((1,2),(3,4)) (repeatable).",
+    ),
+    input_data_weight: str = typer.Option(
```
steven-murray: this should probably be a `Literal["option1", option2", etc]`

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+        "identity", help="Data weighting for OQE. See PSpecData.pspec."
+    ),
+    norm: str = typer.Option("I", help="M-matrix normalization type for OQE."),
```
steven-murray: this should probably also be a Literal with specific options

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+    exclude_cross_bls: bool = typer.Option(
+        False,
+        help="If blpairs not given, exclude baselines paired with a different baseline.",
+    ),
+    exclude_permutations: bool = typer.Option(
+        False, help="If blpairs not given, exclude baseline-pair permutations."
+    ),
+    nblps_per_group: int | None = typer.Option(
+        None,
+        "--nblps-per-group",
```
steven-murray: I don't think this alias is required as it is the same as the variable name

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+    ),
+    store_cov: bool = typer.Option(
+        False, help="Compute and store bandpower covariance."
+    ),
+    store_cov_diag: bool = typer.Option(
+        False, help="Compute and store QE-formalism error bars."
+    ),
+    return_q: bool = typer.Option(False, help="Return unnormalized bandpowers."),
+    overwrite: bool = typer.Option(False, help="Overwrite output if it exists."),
+    cov_model: str = typer.Option(
```
steven-murray: should be literal

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+    verbose: bool = typer.Option(False, help="Report feedback to stdout."),
+    file_type: str = typer.Option(
+        "uvh5", help="Filetype of input UVData. Default 'uvh5'."
+    ),
+    filter_extension: list[str] | None = typer.Option(
+        None,
+        "--filter-extension",
+        click_type=click.Tuple([int, int]),
+        help="Per-spw filter extension, e.g. --filter-extension 20 20 (repeatable).",
+    ),
```
steven-murray: I don't think you need the alias again, and can make the type `tuple[int, int]`

src/hera_pspec/cli.py
```diff
@@ -171,3 +177,347 @@ def fast_merge_baselines(
...
+    symmetric_taper: bool = typer.Option(
+        True,
+        "--symmetric-taper/--no-symmetric-taper",
```
steven-murray: I think this is  the default pair of names, so no need to specify

steven-murray (commented):
Thanks again @kartikmandar this is looking a lot nicer. I have made several comments on particular options so far -- I'm going to stop here and let you update those and all similar things throughout the code (I've just commented on a subset of them).

Some extra thoughts:

- It might be better to make a `cli/` subpackage and have each CLI command in its own module, because they're all rather big. 
- However since the bulk of these functions are literally just calling an existing function directly, I wonder if there's a nice way to just re-use the existing parameter types and docstrings rather than re-defining everything. In the past I've used both `typer` and `cyclopts` and in my opinion the latter does a better job of allowing this. Might be worth looking at.
- If possible to just re-use existing functions directly, then it becomes more reasonable to leave all of this in one cli module.

.github/workflows/tests.yml
kartikmandar: fixed that.

src/hera_pspec/cli.py
kartikmandar: typer can't express list[tupel[int,int]] natively so had to use the click.Tuple workaround.

src/hera_pspec/cli.py
kartikmandar: similar issue.

src/hera_pspec/cli.py
kartikmandar: yeah will do this change.

src/hera_pspec/cli.py
kartikmandar: will change this too

src/hera_pspec/cli.py
kartikmandar: got it.

src/hera_pspec/cli.py
kartikmandar: will change this.

src/hera_pspec/cli.py
kartikmandar: this has the same issue.

src/hera_pspec/cli.py
kartikmandar: sure

src/hera_pspec/cli.py
```diff
@@ -5,96 +5,75 @@
...
 
 from . import container, grouping, pspecdata, pstokes, utils  # noqa: E402
 from .uvpspec import recursive_combine_uvpspec  # noqa: E402
 
 
-@app.command()
+@app.command
 def hello() -> None:
-    # This is a test command which we need for the CLI interface to be broken into
-    # subcommands (at least two commands need to be defined for it to be used as subc's)
```
steven-murray, line 23: This one can now be removed!

src/hera_pspec/cli.py
```diff
@@ -179,159 +158,147 @@ def fast_merge_baselines(
...
+    input_data_weight: Literal["identity", "iC", "dayenu"] = "identity",
+    norm: Literal["I", "H^-1", "V^-1/2"] = "I",
+    taper: str = "none",
+    beam: Path | None = None,
+    cosmo: list[float] | None = None,
+    rephase_to_dset: int | None = None,
+    trim_dset_lsts: bool = False,
+    broadcast_dset_flags: bool = False,
+    time_thresh: float = 0.2,
+    jy2mk: Annotated[bool, Parameter(name="--Jy2mK", negative="--no-Jy2mK")] = False,
```
steven-murray: I think we should leave this as default without providing the `name/negative` renaming.

src/hera_pspec/cli.py
```diff
@@ -179,159 +158,147 @@ def fast_merge_baselines(
...
+        Baseline-pair antennas, e.g. --blpair 1 2 3 4 -> ((1,2),(3,4)) (repeatable).
+    input_data_weight
+        Data weighting for OQE. See PSpecData.pspec.
+    norm
+        M-matrix normalization type for OQE.
+    taper
+        Taper function for the OQE delay transform.
+    beam
+        Filepath to a UVBeam healpix beam map.
+    cosmo
```
steven-murray: Since the parameters for this function are virtually identical to the library function, I wonder if we can't just use `pspec_run.__doc__`?

AGENTS.md
```diff
@@ -61,7 +61,7 @@ The library is organized around three persistent object types and a computation
 
 There are two parallel entry-point styles:
 
-- The new `pspec` Typer app (`src/hera_pspec/cli.py`) — currently exposes `hello`, `fast-merge-baselines`, `run`, `bootstrap`, `auto-noise`, and `generate-pstokes`. Add new subcommands here.
+- The `pspec` cyclopts app (`src/hera_pspec/cli.py`, `app = cyclopts.App()`) — currently exposes `hello`, `fast-merge-baselines`, `run`, `bootstrap`, `auto-noise`, and `generate-pstokes`. Add new subcommands here.
```
steven-murray: remove `hello` from this list

AGENTS.md
```diff
@@ -19,7 +19,7 @@ The project uses **uv** for environment and lockfile management (matching CI). P
   - To bypass coverage while iterating: `uv run pytest --no-cov tests/test_pspecdata.py::test_name`.
 - Warnings job (matches `warnings-tests.yml`, fails on any warning): `MPLBACKEND=agg uv run pytest -Werror`. New code must not emit warnings under this run.
 - Build docs locally: `uv run --group docs sphinx-build docs docs/_build/html` (RTD config at `.readthedocs.yaml`).
-- CLI entrypoint installed as `pspec` (defined at `src/hera_pspec/cli.py`, `app = typer.Typer()`); the legacy `scripts/*.py` shims are also installed via `tool.setuptools.script-files`.
+- CLI entrypoint installed as `pspec` (defined at `src/hera_pspec/cli.py`, `app = cyclopts.App()`); the legacy `scripts/*.py` shims are also installed via `tool.setuptools.script-files`.
```
steven-murray: are these shims still in this PR?

steven-murray (commented):
THanks @kartikmandar this is looking very good. Only a few small comments, and I think that if we are basically providing a very thin wrapper around an existing function, we should just use its `__doc__`!

AGENTS.md
kartikmandar: fixed.

src/hera_pspec/cli.py
kartikmandar: done.

src/hera_pspec/cli.py
kartikmandar: done.

src/hera_pspec/cli.py
kartikmandar: yeah I changed that.

AGENTS.md
kartikmandar: Only `scripts/pspec_red.py` is still installed via `script-files` now; updated the
line to say so. (`scripts/psc_merge_spectra.py` is still in the repo but isn't
shipped.) The four `*_run.py` scripts were removed earlier in this PR.

steven-murray (approved):

PR #501, opened 2026-06-15
docs: render the example notebooks in the online docs

kartikmandar:
Closes #466.

Moves `examples/` → `docs/examples/` and renders the tutorial notebooks on
Read the Docs via `nbsphinx` (theme switched to `sphinx_rtd_theme`; `pandoc`
added to the RTD apt packages). Adds `docs/tutorials.rst` with a TOC of all 11
notebooks and fixes the stale `tree/master/examples` links.

Notebooks render from committed outputs (`nbsphinx_execute = "never"`), not
executed on build. Render-only scope — no CI docs job added.

Verified: `pytest` 265 passed; `ruff check`/`format` clean; docs build succeeds.


.readthedocs.yaml
```diff
@@ -11,6 +11,10 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  # nbsphinx converts the example notebooks' Markdown cells with pandoc, which is
+  # required even though notebooks are rendered from saved outputs (not executed).
+  apt_packages:
+    - pandoc
```
steven-murray: I've used nbsphinx a lot and never had to add pandoc explicitly as an apt dependency in the RTD config... are you sure this is necessary?

steven-murray (commented):
Thank you @kartikmandar this helps a lot. I have just the one comment to check up on.

I will say that currently the documentation is a little confusing and unorganized (totally unrelated to this PR which does a great job of organising the notebook docs). We should have a separate PR in which the API documentation is better summarized.

Also, it would be good to go through these tutorial notebooks and decide which are still useful (some are quite old) and also to run them all again to ensure they actually are up to date with the code itself.

.readthedocs.yaml
kartikmandar: Ahh you are right about that, I added pandoc because for a local sphinx-build pandoc was still needed. But I think that can be left for the developer/maintainer to install themselves.

steven-murray (approved):

steven-murray (approved):

PR #508, opened 2026-06-24
ci: cap pytest below 9.1 to unbreak pytest-cases

kartikmandar:
Closes #503.

pytest 9.1 changed pytest's internal `IdMaker` signature, and
`pytest-cases` 3.10.1 still calls the old signature at import time. This breaks
the `lowest-direct` and `upgraded` dependency-resolution jobs.

Temporarily cap pytest at `<9.1` until `pytest-cases` supports pytest 9.1+.

Verified:
- `uv lock --upgrade-package pytest --dry-run` keeps pytest at 9.0.3.
- `uv run pytest --co -q tests/test_grouping.py tests/test_uvpspec.py` collects
  111 tests successfully.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,10 @@ docs = [
 ]
 tests = [
     "coverage>=7.0",
-    "pytest>=8.0",
+    # Upper bound: pytest 9.1 removed a parameter from the internal IdMaker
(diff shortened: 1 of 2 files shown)


adeliegorce (approved):
Thanks! The PR is fine but let's not close issue #503 to remember this was a temporary fix

Issue #451, opened 2026-05-04
Include window function computation into pipeline


adeliegorce:
The window function computation (with the `UVWindow` module, beyond the delay approximation) should be included in the pipeline (so that they're computed automatically as one of the tasks in the makeflow, instead of run separately by hand).

PR #514, opened 2026-07-07
Add FT-beam and window function CLI; fix k_parallel scaling of exact WFs

vincentmackay:
Addresses #451 (will be fully closed when I also address hera_pipelines accordingly). Moves the FT-beam and window function production scripts from hera_pipelines into hera_pspec as `pspec compute-ft-beam` / `pspec compute-window-functions`.

Note that working on this exposed a bug that affected the IDR3.1.1a theory release window functions: `UVWindow.from_uvpspec` assumed the FT beam and the data share a channel width, but I was accidentally computing the FT beam on a 97.66 kHz grid vs the data's 122.07 kHz, scaling the WFs' k_parallel axis by 1.25x. To prevent this error in the future, I forced the FT beam to be computed on the data's exact channels (`FTBeam.select_freqs`; interpolates with a warning on mismatch, raises an exception if not covered).

- commit 1: the bug fix + regression tests
- commit 2: vectorized cylindrical binning (~20x faster; new test pins it to the old loop implementation at 1e-12)
- commit 3: `FTBeam.from_beam` (port of @adeliegorce's routine, az_za + healpix), `write_hdf5`, the two CLI commands

Parts of this were done with the help of Claude (Fable 5), e.g., for finding the vectorization, and documenting/cleaning up.

### DIFF
--- a/src/hera_pspec/cli.py
+++ b/src/hera_pspec/cli.py
@@ -475,9 +475,7 @@ def compute_ft_beam(
     mapsize: float = 1.0,
     npix: int = 301,
     label: str = "HERA",
-    search_dirs: Annotated[
-        list[Path] | None, Parameter(consume_multiple=True)
-    ] = None,
+    search_dirs: Annotated[list[Path] | None, Parameter(consume_multiple=True)] = None,
     force_recompute: bool = False,
     verbose: bool = False,
 ) -> None:
@@ -535,8 +533,7 @@ def compute_ft_beam(
     uvp = _load_uvp(pspec_file, group, name)
     freq_array = np.unique(uvp.freq_array)
     logger.info(
-        "Using the %d frequency channels of %s (%.2f-%.2f MHz, "
-        "channel width %.2f kHz)",
+        "Using the %d frequency channels of %s (%.2f-%.2f MHz, channel width %.2f kHz)",
         freq_array.size,
         pspec_file,
         freq_array.min() / 1e6,
@@ -554,11 +551,7 @@ def compute_ft_beam(
             return
 
     ftbeam = FTBeam.from_beam(
-        beamfile=beam_file,
-        pol=pol,
-        freq_array=freq_array,
-        mapsize=mapsize,
-        npix=npix,
+        beamfile=beam_file, pol=pol, freq_array=freq_array, mapsize=mapsize, npix=npix
     )
     out_path = Path(out_dir) / fname
     ftbeam.write_hdf5(
@@ -566,7 +559,7 @@ def compute_ft_beam(
         overwrite=True,
         extra_attrs={
             "beam_file": str(beam_file),
-            "created_utc": datetime.datetime.now(datetime.timezone.utc).isoformat(),
+            "created_utc": datetime.datetime.now(datetime.UTC).isoformat(),
             "producer": "pspec compute-ft-beam",
         },
     )
@@ -629,9 +622,7 @@ def _compute_one_wf(
         f.attrs["spw_freq_max_hz"] = float(spw_freqs.max())
         f.attrs["pspec_file"] = str(pspec_file)
         f.attrs["ft_beam_file"] = str(ft_beam_file)
-        f.attrs["created_utc"] = datetime.datetime.now(
-            datetime.timezone.utc
-        ).isoformat()
+        f.attrs["created_utc"] = datetime.datetime.now(datetime.UTC).isoformat()
         f.attrs["producer"] = "pspec compute-window-functions"
     logger.info("Wrote %s", out_path)
     return str(out_path.absolute())
--- a/src/hera_pspec/cli.py
+++ b/src/hera_pspec/cli.py
@@ -1,18 +1,24 @@
 """A CLI interface for hera_pspec."""
 
 import copy
+import datetime
 import glob
+import logging
 import pickle
 import sys
+from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 from typing import Annotated, Literal
 
 import h5py
+import numpy as np
 from cyclopts import App, Parameter
 from rich.console import Console
+from rich.logging import RichHandler
 from tqdm import tqdm
 
 cns = Console()
+logger = logging.getLogger("hera_pspec")
 
 app = App(name="pspec", version_flags=[], help_flags=["--help"])
 # cyclopts pattern: register subcommands after app is constructed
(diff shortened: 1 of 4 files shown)


src/hera_pspec/cli.py
```diff
@@ -391,3 +397,345 @@ def generate_pstokes(
...
+        PSpecContainer file whose frequency channels define the FT-beam
+        grid.
+    out_dir
+        Directory where freshly computed FT-beam HDF5 files are written.
+    group
+        Group within pspec_file. May be omitted if the file has one group.
+    name
+        Spectrum name within the group. May be omitted if unique.
+    mapsize
+        Half-width of the flat-sky map the beam is projected onto (rad).
```
steven-murray, line 508: I haven't looked further down yet, but I am assuming this is not actually in radians but in direction cosines? I.e. `l=sin(theta)`?

src/hera_pspec/cli.py
```diff
@@ -391,3 +397,345 @@ def generate_pstokes(
...
+        f.attrs["spw_index"] = int(spw)
+        f.attrs["dataset_label"] = dataset_label
+        f.attrs["taper"] = str(sub.taper)
+        f.attrs["spw_freq_min_hz"] = float(spw_freqs.min())
+        f.attrs["spw_freq_max_hz"] = float(spw_freqs.max())
+        f.attrs["pspec_file"] = str(pspec_file)
+        f.attrs["ft_beam_file"] = str(ft_beam_file)
+        f.attrs["created_utc"] = datetime.datetime.now(datetime.UTC).isoformat()
+        f.attrs["producer"] = "pspec compute-window-functions"
+    logger.info("Wrote %s", out_path)
```
steven-murray, line 625: This is fine for now, but I'm flagging that in general it would be better to define external read/write methods for a "window function file" format, and then re-use these across the code, rather than having the format hard-coded inside this function.

src/hera_pspec/uvwindow.py
```diff
@@ -9,11 +10,64 @@
...
 from . import uvpspec_utils as uvputils
 
+logger = logging.getLogger(__name__)
+
+# Cartesian half-width used for the flat-sky beam projection in
+# FTBeam.from_beam. The Cartesian grid spans rx in [-_MAX_R, +_MAX_R];
+# zenith angle is za = sqrt(x^2 + y^2) in radians, so the corner of the
+# grid sits at za = sqrt(2)*_MAX_R = 1 rad (~57 deg), a sensible outer
+# bound for the beam-relevant sky.
+_MAX_R = 0.7
```
steven-murray, line 25: This is a different default than in the CLI -- is this intentional? I wonder what @adeliegorce recommends this parameter to be set to?

tests/test_uvwindow.py
```diff
@@ -116,7 +116,169 @@ def test_init(self):
 
     def test_from_beam(self):
 
-        pytest.raises(NotImplementedError, uvwindow.FTBeam.from_beam, beamfile="test")
+        from pyuvdata import UVBeam
```
steven-murray, line 119: I would put this import at the top of the module

tests/test_uvwindow.py
```diff
@@ -116,7 +116,169 @@ def test_init(self):
...
+
+        beamfile = os.path.join(DATA_PATH, "HERA_NF_dipole_power.beamfits")
```
steven-murray, line 121: Prefer `pathlib.Path` operations (i.e. `beamfile = Path(DATA_PATH) / "HERA_NF_dipole_power.beamfits"`)

tests/test_uvwindow.py
```diff
@@ -116,7 +116,169 @@ def test_init(self):
...
+        ngrid = test.ft_beam.shape[-1]
+        for i in range(freq_array.size):
+            assert np.argmax(test.ft_beam[i]) == (ngrid**2) // 2
+
+        # pol fed as int
+        test2 = uvwindow.FTBeam.from_beam(
+            beamfile=beamfile, pol=-5, freq_array=freq_array, mapsize=1.0, npix=29
+        )
+        assert test2.pol == "xx"
+        assert np.allclose(test.ft_beam, test2.ft_beam)
```
steven-murray, line 147: Test functions should preferably test ONE thing. Split this second (and third) test into different, explicitly named functions.

tests/test_uvwindow.py
```diff
@@ -116,7 +116,169 @@ def test_init(self):
...
+        assert np.allclose(sub.freq_array, self.bandwidth[3:10])
+        assert np.allclose(sub.ft_beam, test.ft_beam[3:10])
+        # original object untouched by inplace=False
+        assert test.freq_array.size == self.bandwidth.size
+
+        # inplace=True modifies the object
+        test2 = uvwindow.FTBeam.from_file(ftfile=self.ft_file, spw_range=None)
+        out = test2.select_freqs(self.bandwidth[3:10], inplace=True)
+        assert out is None
+        assert np.allclose(test2.ft_beam, test.ft_beam[3:10])
```
steven-murray, line 208: same comment here.

tests/test_uvwindow.py
```diff
@@ -116,7 +116,169 @@ def test_init(self):
...
+        with pytest.warns(UserWarning, match="interpolating"):
+            part = uvwindow.FTBeam.from_file(ftfile=self.ft_file, freq_array=midpoints)
+        with pytest.warns(UserWarning, match="interpolating"):
+            ref = full.select_freqs(midpoints, inplace=False)
+        assert np.allclose(part.freq_array, ref.freq_array)
+        assert np.allclose(part.ft_beam, ref.ft_beam)
+
+    def test_write_hdf5(self):
+
+        import tempfile
```
steven-murray, line 265: there is a pytest fixture that provides temporary files (`tmp_path` IIRC)

tests/test_uvwindow.py
```diff
@@ -520,6 +685,138 @@ def test_get_wf_for_tau(self):
         wf_array1 = np.zeros((kperp_bins.size, test.Nfreqs))
         kpara, cyl_wf = test._get_wf_for_tau(tau, wf_array1, kperp_bins, kpara_bins)
 
+    def test_cylindrical_wf_matches_loop_reference(self):
```
steven-murray, line 688: I like that this test has been run, but it seems extremely convoluted for a single test. Not quite sure what to do about it, but keeping a full copy of the old code around just in the tests seems like a code smell to me. I would prefer to test against simple known examples rather than full alternate code implementations.

steven-murray (commented):
Thanks you @vincentmackay this is looking fantastic! I just have a few small comments.

src/hera_pspec/uvwindow.py
adeliegorce: Yes `_MAX_R` should be set to 0.7. Then to pad the beam simulation and "artificially" cover smaller kperp, I have set `mapsize=1` in previous computations; if `mapsize=_MAX_R` you just interpolate the beam simulation as is, with no padding.

adeliegorce: Actually we could probably just set it to `1./np.sqrt(2.)`

src/hera_pspec/cli.py
adeliegorce: This is a poorly named and defined parameter and it is totally on me! The cartesian map the beam is interpolated over has width 1 rad * sqrt(2)* mapsize (provided max_r=0.7).  If mapsize=1/sqrt(2)~0.7 then the map is 1 rad wide.

src/hera_pspec/uvwindow.py
```diff
@@ -9,11 +10,64 @@
 import uvtools.dspec as dspec
 from astropy import units
 from pyuvdata import utils as uvutils
-from scipy.interpolate import RegularGridInterpolator
+from scipy.interpolate import RegularGridInterpolator, interp1d
```
adeliegorce, line 13: This is due to my code but `interp1d` is legacy code and we probably should not use it

src/hera_pspec/uvwindow.py
```diff
@@ -87,33 +141,158 @@ def __init__(
...
+            Default: 301.
         verbose : bool, optional
             If True, print progress, warnings and debugging info to stdout.
         x_orientation: str, optional
             Orientation in cardinal direction east or north of X dipole.
             Default keeps polarization in X and Y basis.
             Used to convert polstr to polnum and conversely.
         """
-        raise NotImplementedError("Coming soon...")
+        from pyuvdata import UVBeam
```
adeliegorce, line 195: This import could probably be moved to the top

src/hera_pspec/uvwindow.py
```diff
@@ -132,37 +311,64 @@ def from_file(cls, ftfile, spw_range=None, **kwargs):
             In (start_chan, end_chan). Must be between 0 and 1024 (HERA
             bandwidth).
             If None, whole instrument bandwidth is considered.
+        freq_array : 1D array or list of floats, optional
```
adeliegorce, line 314: Isn't the `spw_range` parameter redundant now we have `freq_array`?

tests/test_uvwindow.py
adeliegorce, line 1: Unfortunately this will be extremely annoying to merge with #491 . It is probably worth putting any test writing on hold until this PR has been merged. Then Claude can easily merge your new tests into the new test_uvwindow.

adeliegorce (commented):
Thanks @vincentmackay! good job spotting the frequency channelization mismatch. My main comment is that we need to coordinate merging #491 and this PR as #491 introduces major changes to the testing infrastructure and will create conflicts with the tests you've added

src/hera_pspec/cli.py
vincentmackay: Ah right it's not rad. I reworded to match Adélie's definition. I left `_MAX_R` to 0.7 just to match the currently-produced WFs, but could change to 1/sqrt(2) if we feel like that's best.

src/hera_pspec/cli.py
vincentmackay: Yeah, we should have a discussion about best formatting for WFs (do we save them in hera_pspec objects, or as separate objects since they're so large...). Left for a follow-up PR.

src/hera_pspec/uvwindow.py
vincentmackay: See comment above (basically I left 0.7 because that's what the latest data release uses, but we could change it for the next one).

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: I went with two tests that cover the "granularity" extremes (one band-wide bin, should reduce to freq mean; vs one bin per k_para value, must reduce to identity). I'll be honest, I got that by brainstorming with Claude about how to get over that code smell, think it made sense. Note: the old loop-copy comparison agreed at 1e-12 before I deleted it.

src/hera_pspec/uvwindow.py
vincentmackay: Got it! Replaced with an explicit linear interpolation, no scipy legacy API.

src/hera_pspec/uvwindow.py
vincentmackay: Done.

src/hera_pspec/uvwindow.py
vincentmackay: You're right, so now `spw_range` and `update_spw` now show `DeprecationWarning` (and point at `freq_array / `select_freq`). We could fully remove them in a future release.

tests/test_uvwindow.py
vincentmackay: Yeah, like you suggested I ended up hiring Claude to help rebase now that #491 is in. One heads-up, I moved the input power spectrum channels for the test that checks the no-cosmology warning, from 5-25 to 175-195. The old range wasn't covered by the FT-beam file. Hopefully everything is good!

tests/test_uvwindow.py
```diff
@@ -205,17 +214,125 @@ def test_float_pol_raises_typeerror(self, ft_beam_spw: uvwindow.FTBeam) -> None:
             uvwindow.FTBeam(pol=3.4, data=data, freq_array=freq_array, mapsize=mapsize)
 
 
+@pytest.fixture(scope="module")
+def nf_dipole_beamfits() -> Path:
```
adeliegorce: I don't think this needs to be a fixture, it could be a global parameter or the path could be called explicitly in each test

tests/test_uvwindow.py
```diff
@@ -249,7 +368,10 @@ def test_no_spw_range_uses_full_bandwidth(self, ft_bandwidth: np.ndarray) -> Non
 
     @pytest.mark.parametrize("bad_spw", [(13,), (20, 10), (1001, 1022)])
     def test_invalid_spw_range(self, bad_spw: tuple[int, ...]) -> None:
```
adeliegorce, line 411: Did you get a chance to add the equivalent test for the new `freq_array` parameter?

tests/test_uvwindow.py
```diff
@@ -1159,3 +1287,196 @@ def test_k_wrong_units_error(
...
+
+
+class TestFTBeamWriteHdf5:
+    def test_roundtrip(self, ft_beam_spw: uvwindow.FTBeam, tmp_path: Path) -> None:
+        # the filename carries no pol suffix: the round trip relies on the
+        # pol attribute written by write_hdf5
+        fname = tmp_path / "ft_beam_roundtrip.hdf5"
+        ft_beam_spw.write_hdf5(fname, extra_attrs={"beam_file": "sim.fits"})
+        back = uvwindow.FTBeam.from_file(ftfile=fname)
+        assert back.pol == ft_beam_spw.pol
```
adeliegorce: Is it possible to simply check `ft_beam_spw == back` or are some attributes different? (should there be?)

adeliegorce (commented):
Hi @vincentmackay! 
Thank you for this new version and for addressing my comments. They're all resolved now. The tests are well thought-of (not only testing features but also that the output makes sense), that is great, however, as @steven-murray pointed out, we should avoid letting the coverage drop so would it be possible to add tests covering the new CLI methods?
I also added a couple minor comments.

tests/test_uvwindow.py
vincentmackay: Done.

tests/test_uvwindow.py
vincentmackay: Added!

tests/test_uvwindow.py
vincentmackay: You're right, now there wasn't an `__eq__` defined for that class so I added it, and now the test is just `assert back == ft_beam_spw`.

vincentmackay:
I added CLI tests to improve patch coverage, and addressed the other items brought up by Adélie. Let me know if you catch anything!

adeliegorce (approved):
All good on my side! thanks a lot @vincentmackay :)

PR #524, opened 2026-08-03
refactor: use pyuvdata GaussianBeam inside PSpecBeamGauss

kartikmandar:
Closes #517.

`PSpecBeamGauss` now wraps a `pyuvdata.analytic_beam.GaussianBeam`, exposed as
`self.analytic_beam` as a hook for future window-function work. The constructor takes
mutually exclusive `fwhm=` (power-beam FWHM in radians, the existing API) and `beam=`
(a ready-made achromatic `GaussianBeam`) arguments, with the E-field vs power sigma
convention (a factor of sqrt(2)) converted internally. The closed-form beam integrals
are untouched, so science output is identical — the pre-existing tests pass unedited.
Chromatic beams (`spectral_index != 0`) are rejected for now and left as a possible
follow-up.


docs/pspecbeam.rst
```diff
@@ -73,7 +73,19 @@ For example, to specify a Gaussian beam with a constant FWHM of 0.1 radians, def
   beam_freqs = np.linspace(100e6, 200e6, 200) # in Hz
 
   # Create a new Gaussian beam object with full-width at half-max. of 0.1 radians
-  beam_gauss = hp.PSpecBeamGauss(fwhm=0.1, beam_freqs=beam_freqs)
+  beam_gauss = hp.PSpecBeamGauss(0.1, beam_freqs=beam_freqs)
```
steven-murray: Can we add the argument name back? I prefer the clarity.

docs/pspecbeam.rst
```diff
@@ -73,7 +73,19 @@ For example, to specify a Gaussian beam with a constant FWHM of 0.1 radians, def
...
+
+The beam is backed by a ``pyuvdata.analytic_beam.GaussianBeam``, exposed as the ``analytic_beam`` attribute. You can also construct a ``PSpecBeamGauss`` directly from an achromatic, sigma-specified ``GaussianBeam``:
```
steven-murray:
```suggestion
The beam is backed by a ``pyuvdata.analytic_beam.GaussianBeam``, exposed as the ``analytic_beam`` attribute. You can also construct a ``PSpecBeamGauss`` directly from an achromatic ``GaussianBeam``:
```

src/hera_pspec/pspecbeam.py
```diff
@@ -349,15 +350,24 @@ def get_Omegas(self, polpairs):
 
 
 class PSpecBeamGauss(PSpecBeamBase):
-    def __init__(self, fwhm, beam_freqs, cosmo=None):
+    def __init__(self, beam, beam_freqs, cosmo=None):
```
steven-murray: I would have both arguments (i.e. `fwhm` and `beam`, and only allow one to be specified at a time (and both default to None)

steven-murray (commented):
Looks good! Only one requested change.

src/hera_pspec/pspecbeam.py
kartikmandar: Done in 25dfa09. The constructor now takes mutually exclusive `fwhm=` and `beam=` keyword arguments (both defaulting to `None`), raising a `ValueError` if both or neither are given. Also restored the `fwhm=` keyword in the docs and notebook examples and applied your wording suggestion on the rst.

steven-murray (approved):

PR #528, opened 2026-08-21
Fix spherical_average for UVPSpec without window functions

vincentmackay:
`spherical_average` raises `UnboundLocalError` when the input UVPSpec has no `window_function_array` (computed with `store_window=False`): `Nk_theory` and the theory k-bin edges are only defined under `store_window`, but the `Aw` matrix was allocated and populated unconditionally. This guards both under `store_window`, and adds a regression test. Surfaced running IDR3 power spectra, which don't store window functions.


vincentmackay:
Hi @adeliegorce -- you're right, I had accidentally based that PR on `main` instead of `release/v1` so it showed all of @kartikmandar's PR in the diff. My bad! I changed it now and the diff is now just the one fix I brought. Sorry for the confusion.

adeliegorce (approved):
All good, cheers!

